PR #1056, opened 2014-12-18
Update _betterbook.yaml

yuzver:
Some lines of translation.

### DIFF
--- a/src/site/_ru/_betterbook.yaml
+++ b/src/site/_ru/_betterbook.yaml
@@ -1,5 +1,5 @@
 toc:
-  - title: "Web <strong>Fundamentals</strong>"
+  - title: "Основополагающие <strong>веб-стандарты</strong>"
     path: /fundamentals/
     home: true
   - title: Начало работы
@@ -57,9 +57,9 @@ toc:
         path: /fundamentals/discovery-and-distribution/
         id: discovery-and-distribution
         section:
-          - title: "Optimization for Crawlers"
+          - title: "Оптимизация под поисковых ботов"
             path: /fundamentals/discovery-and-distribution/optimization-for-crawlers/
-          - title: "Enriching Search Results"
+          - title: "Улучшение результатов поиска"
             path: /fundamentals/discovery-and-distribution/enriching-search-results/
       - title: Доступ к устройству
         path: /fundamentals/device-access/
@@ -97,6 +97,6 @@ toc:
         path: /fundamentals/resources/styleguide/
       - title: Пример статьи
         path: /fundamentals/resources/example-article/
-  - title: "Web <strong>Starter Kit</strong>"
+  - title: "Инструмент <strong>веб-разработки</strong>"
     path: /starter-kit/
     in_header: false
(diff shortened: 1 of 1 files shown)


src/site/_ru/_betterbook.yaml
```diff
@@ -1,5 +1,5 @@
 toc:
-  - title: "Web <strong>Fundamentals</strong>"
+  - title: "Основополагающие <strong>веб-стандарты</strong>"
```
PaulKinlan: I believe the emphasis is wrong here.  The brand guidelines we have are for bolding "Fundamentals"

src/site/_ru/_betterbook.yaml
```diff
@@ -57,9 +57,9 @@ toc:
         path: /fundamentals/discovery-and-distribution/
         id: discovery-and-distribution
         section:
-          - title: "Optimization for Crawlers"
+          - title: "Оптимизация под поисковых ботов"
             path: /fundamentals/discovery-and-distribution/optimization-for-crawlers/
-          - title: "Enriching Search Results"
+          - title: "Улучшение результатов поиска"
```
PaulKinlan: Looking at the translation it says "improving", I think this is not correct, we are not improving the placing of search results, we are making your site look better.

PaulKinlan:
I've added some comments, otherwise it looks good.

src/site/_ru/_betterbook.yaml
```diff
@@ -57,9 +57,9 @@ toc:
         path: /fundamentals/discovery-and-distribution/
         id: discovery-and-distribution
         section:
-          - title: "Optimization for Crawlers"
+          - title: "Оптимизация под поисковых ботов"
```
denpatin: I presume it would sound slightly better from the point of the stylistics of Russian if you write "для" instead of "под".

src/site/_ru/_betterbook.yaml
denpatin: @PaulKinlan Do I get it right that by "enriching" you understand some good practices to follow when marking up the page so there appear more chances of the considered website to be found and placed as above as possible in the search result list?

yuzver:
Thank you for the comments. Changes have been made.

src/site/_ru/_betterbook.yaml
```diff
@@ -1,5 +1,5 @@
 toc:
-  - title: "Web <strong>Fundamentals</strong>"
+  - title: "<strong>Фундаментальные</strong> веб-стандарты"
```
denpatin: As for me, the word "фундаментальный" sounds a bit unnatural here, even weird. Also, the word "стандарт" is also too strong in this context. Frankly, I would simply say in a title-like manner: "Основы веба" or even "Веб: основы". Or just "Основы работы в вебе". Of course, the choice is up to you :)

petele:
We need the CLA signed before we can integrate this.

yuzver:
Hello, already signed.

petele:
Our system doesn't seem to think so :( Can you please double check you used the right email and user name?

yuzver:
Pete, thrice checked - github account name & email. both is correct. seriously.

try again to accept the agreement and: "It looks like you've already signed this CLA. If you'd like to edit your contact information, you may do so below."

Issue #997, opened 2014-11-12
Move Configuration files to the top level


ianbarber:
At the moment the configuration files are scattered throughout the project. We should consolidate and simplify. This includes the jekyll configs, gruntfiles, gemfiles, and NPM json. The GAE config could probably stay where it is, but it might be nice to pull it up and have it copied over as part of the build. 

cc @wibblymat FYI.

wibblymat:
Closed by #1060

PR #1060, opened 2014-12-20
Rearranged things.

wibblymat:
- Translations separated into a path of their own
- Moved dot files up to the root
- Moved appengine path up to the root


PaulKinlan:
@wibblymat looks like a re-basing is in order.

wibblymat:
Rebasing always makes me sad. Why do all of the other commits now show up in this PR? The diff is now totally impossible to understand :(

wibblymat:
Ok, managed to sort it out. I have split the PR into separate commits so that it is possible to review the code changes (by looking at just the third commit) without the file moves drowning everything out.

wibblymat:
Fixes #997, in case anyone is wondering why I bothered.

PaulKinlan:
Thanks bud.

Issue #998, opened 2014-11-12
Move local build to GAE


ianbarber:
We currently have 3 local build states, which include one targeting GAE for web-central.appspot.com, one targeting developers.google.com and one straight Jekyll build. We should simplify to a single output structure which is handled by GAE on local and staging, and is easily mappable for d.g.c. This will require a new local dev requirement, but in the process we should be able to simplify some of the other build requirements. 

/cc @wibblymat

wibblymat:
Closed by #1076

Issue #843, opened 2014-09-19
Simplify the build process


PaulKinlan:
*  Remove Grunt
*  Remove Makefile 
*  Use the gulp process from Web Starter Kit
*  Fix font building
*  Make it easier to rebuild CSS
*  Make it so that we can test all locales

petele:
I'm going to close this for now as we're working on a complete site refresh using material design lite which should fix this issue.

PR #1076, opened 2015-01-06
Local development with GAE

### LINKED ISSUE #1048: 'make server' requires links to have .html on the end to work locally
The local server needs .html to be added to the ends of links to work: was looking at /stickyness/ -> some other URL under /stickyness/.... and it just 404's on me :(

wibblymat:
Closes #1048 and #998, a step towards #843.


PaulKinlan:
This looks good overall.  Just to confirm, we are getting rid of the  "open" because we now have a full template that is rendered on app engine, but the build for deployment still correctly misses out the <html><head> magic.

wibblymat:
Correct. The build is now always the devsite-style, and the App Engine script adds the extra tags.

PR #1079, opened 2015-01-07
Update 'View Source' links to match new code layout.

### LINKED ISSUE #1078: View Source links broken
The "View Source" links in the header 404. For example, the homepage's source currently links to:

https://github.com/Google/WebFundamentals/tree/master/src/site/_en/fundamentals/index.html

It should instead link to:

https://github.com/Google/WebFundamentals/tree/master/src/_langs/en/fundamentals/index.html

Looks like the fix would simply involve a small change to https://github.com/google/WebFundamentals/blob/master/src/_includes/beta.liquid.

wibblymat:
Fixes #1078.


addyosmani:
:+1:

PR #1115, opened 2015-01-29
Upate Chinese translation

viztor:


PaulKinlan:
Sorry for the delay, we will look through this as soon as possible.

PaulKinlan:
Sorry for the delay, looking at this there are some files that are added but not translated, some have partial translations too.  We can only pull in pages where there are full translations.

petele:
As Paul mentioned, we only want to integrate fully translated pages.

PR #1118, opened 2015-01-31
Suggest Clumsy as a testing option on Windows

Aeyoun:
The open-source utility is very easy to use and has a lower barrier to entry than the other tools. It is very close to Network Link Conditioner on OS X.

Disclaimer: I have no relation to Clumsy or their website other than having used it as a user for eight months.

### DIFF
--- a/src/_langs/en/fundamentals/tools/test/mobilenetwork.markdown
+++ b/src/_langs/en/fundamentals/tools/test/mobilenetwork.markdown
@@ -40,7 +40,8 @@ you'll be able to test for different network types.
 
 <img src="imgs/network-panel.png" alt="OS X Network Panel" />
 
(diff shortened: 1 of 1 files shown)


PaulKinlan:
cc @addyosmani and @gauntface for feedback.  I believe we have been trying to limit ourselves to one choice rather than a selection to help avoid choice paralysis.  I'll let Matt decide what to do.

addyosmani:
I personally don't have experience with Clumsy. @gauntface? Where possible I would like us to manually evaluate options to ensure they provide sane recommendations before we commit to including them.

petele:
Seeing that most folks don't have experience, I'm going to close this for now, if it becomes more popular or offers a special set of features, happy to revisit.

PR #1120, opened 2015-02-03
Cleaning up the name and autocomplete attributes section

madmath:
I listed a few more autocomplete attribute values (with their descriptions in case they are ambiguous) and put a link to the HTML Autofill spec.


madmath:
^^ Done with the CLA.

PaulKinlan:
Awesome. Will pull when I land

Issue #1121, opened 2015-02-03
Typo in /fundamentals/performance/critical-rendering-path/measure-crp.html


srus:
__documentContentLoaded__ should be __DOMContentLoaded__.

PaulKinlan:
@srus did we ever pull this in ?

srus:
Yes, you [merged it](https://github.com/google/WebFundamentals/pull/1122) with master. I accidentally closed this issue, but realized that is your decision to do that, so I reopened it.

PaulKinlan:
Ah. Right :) I am ok with you closing it.

srus:
:+1:

PR #1122, opened 2015-02-03
Fix typo (#1121): "DOMContentLoaded" instead of "documentContentLoaded"

srus:


srus:
Aside of that, there is a minor issue with the name of that event. In many places you can read __domContentLoaded__ instead of **DOMContentLoaded**.

PaulKinlan:
Ah, this looks like it was a translation issue.

PaulKinlan:
Thank you

PR #1128, opened 2015-02-06
setting-up changes for Sass install instructions

yufengg:
I may be mistaken here, but going through a clean install of NPM, Gulp, and then Sass led me to not have Sass available (this was on Mac OSX Mavericks), Feb 2 2015. 

The existing doc seems to imply that Sass will magically appear somehow when you set up npm and gulp, but no such thing happened and I got a lot of CSS-less websites when I ran `gulp serve`. 

I tried to get it via npm (found a few sass related things) but they didn't cut it.

Eventually I got it the official way via Ruby: `gem install sass`

Feel free to edit the wording of the edit.


yufengg:
I'm a googler, ldap is same as my github username, no need for CLA.

PR #1142, opened 2015-02-13
wrong folder for Sass file

xvweirong:
Sass file should be placed in folder app/styles/components/_components/ instead of app/styles/components/ according to context code


xvweirong:
Greetings,
I just signed CLA.
Kind Regards,David Xu

Date: Fri, 13 Feb 2015 03:31:44 -0800

petele:
This is a subjective thing and making the change is somewhat expensive for little ROI.  I'm closing for now.

PR #1145, opened 2015-02-13
CI on Travis + auto deploy on green

### LINKED ISSUE #1116: Move staging build server to compute
We desperately need to move the staging build server to Compute Engine and away from my RackSpace server, we have not had a working staging environment since we introduced all the translations.  @crhym3 had started to set up the environment but we would love to get that scheduled and started.

cc @marcacohen

crhym3:
@wibblymat can you check whether `grunt develop` still works before merging? I'm having troubles with building locally - some weird errors are thrown by liquid templates. I think it's my setup because Travis builds just fine.

Closes #1116


wibblymat:
All looks good to me :+1:

PR #1158, opened 2015-03-06
Fixing some really bad german writing styles

fokusferit:
-> changed the use of different translations of "breakpoints" -> just call it "Übergangspunkt" and not "Haltepunkt" or use only "haltepunkt"..
-> some german grammar fix
-> Would use "bildschirm" also for smartphone "Displays" instead of "displays". More consistent style.


PaulKinlan:
Thanks.  much appreciated. I'll pull this in, but it might be a week or two before live.

PR #1180, opened 2015-03-25
Update docker image URL

crhym3:
Also, this cuts ~600Mb off the image.
The workflow and tools/docker.sh should be good to go now.

/cc @wibblymat


crhym3:
@google/web-starter-kit this can be applied to any project, e.g. WSK:
https://github.com/google/WebFundamentals/blob/docker-gcr-io/README.md#alternative-dev-workflow-based-on-docker

crhym3:
@surma this is the demo I've just shown :)

PR #1220, opened 2015-04-22
Strip html tag from css class

chenxsan:
I happened to inspect the HTML source code in firefox and found that Google web fundamentals sites are using BEM for css naming. One thing I noticed was a weird css class `main-nav__item main-nav__item--<strong>fundamentals</strong> main-nav__item--home main-nav__item--current`, you can check the screenshot below:

screen shot 2015-04-22 at 06 05 41

I checked the [navigation.liquid](https://github.com/google/WebFundamentals/blob/master/src/_includes/navigation.liquid) files, it has [codes as below](https://github.com/google/WebFundamentals/blob/master/src/_includes/navigation.liquid#L47):

```
<li class="main-nav__item main-nav__item--{% if section.en_title %}{{ section.en_title | downcase | replace:' ','-' }}{% else %}{{ section.title | downcase | replace:' ','-' }}{% endif %} 
```

So I think we need to strip the html tag, and here's a screenshot after what I've done:

screen shot 2015-04-22 at 09 44 21

### DIFF
--- a/src/_includes/navigation.liquid
+++ b/src/_includes/navigation.liquid
@@ -44,7 +44,7 @@
     {% assign stripTitle = section.title | strip_html %}
     {% assign currentPath = section.path | append:"index.html" %}
     {% if section.in_header != false  %}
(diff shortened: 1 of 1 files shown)


PaulKinlan:
thanks!!! :)

PR #1233, opened 2015-04-29
Translations instead of Transitions

sroucheray:


sroucheray:
I signed it

2015-04-29 13:23 GMT+02:00 googlebot <notifications@github.com>:

hebbet:
i think transition is correct here.

jpmedley:
Transition seems right to me as well.

wibblymat:
In the page there is a heading that says "use translations".

In context I think translation is correct. Complicated by the fact that you can have a `transition` which is also a `translation`.

sroucheray:
I agree with @wibblymat in this page there are two similar sentences first "Use transitions to move between views" and later "Use translations to move between views". The fact that transitions and translations are close words and that there are used in similar sentences adds noise to the page.

The second point is more semantic :
_Use transitions to move between views; avoid using `left`, `top` or any other property that triggers layout._
In this sentence if the semicolon contrasts the two proposals, not using left and top means using translations ; if not, semicolon should not be used and the two proposals must be separated in two bullet points because there are not really related.

PaulKinlan:
Let's ask @paullewis exactly what he meant.

PR #1234, opened 2015-04-30
Remove double file ignore

Sitebase:
The asterisk DS_Store ignore is not required. The regular one will also match files in sub directories.


Sitebase:
I signed it!

PaulKinlan:
Good spot.  Thanks.

PR #1236, opened 2015-05-03
Media type fix

### LINKED ISSUE #1235: Feedback for: /fundamentals/performance/critical-rendering-path/render-blocking-css.html
The bullet list at the end of the page points out:
> The second declaration is also render blocking: “screen” is the default type and if you don’t specify any type, it’s implicitly set to “screen”. Hence, the first and second declarations are actually equivalent.

The default is "all" since HTML 5 and, if I remember correctly, it was the de facto standard previously as well, despite the HTML 4 spec saying it's "screen". Modern browsers seem to use "all" as the default regardless of doctype, anyway.

PontusHorn:
Fix for #1235


PaulKinlan:
cc @igrigorik for a review.

igrigorik:
FWIW, I'm inclined to keep "screen". Both "all" and "screen" are valid types [1], where "screen" is a subset of "all". Switching to "all" doesn't win us anything here, and the update would also have to be reflected in all the translations -- seems like a lot of work for nothing.

[1] http://www.w3.org/TR/CSS2/media.html#media-types

PontusHorn:
Sorry, I might have been a bit unclear in the description as I was distracted while writing it. The issue is that it says the media type "screen" is equivalent to not setting a media type at all, which isn't correct. Removing that statement would be another solution, as it's not crucial information, but I've submitted a PR changing it to "all" in all the translations if you want to go for that.

PontusHorn:
Looks like I'm still distracted. I thought you posted that on the issue, not the PR. The original problem description was on issue #1235.

src/_langs/fr/fundamentals/performance/critical-rendering-path/render-blocking-css.markdown
```diff
@@ -94,13 +94,13 @@ Prenons quelques exemples concrets :
...
 {% highlight html %}
 <link href="style.css"    rel="stylesheet">
-<link href="style.css"    rel="stylesheet" media="screen">
+<link href="style.css"    rel="stylesheet" media="all">
 <link href="portrait.css" rel="stylesheet" media="orientation:portrait">
 <link href="print.css"    rel="stylesheet" media="print">
 {% endhighlight %}
 
 * La première déclaration empêche l'affichage et respecte toutes les conditions.
-* La seconde déclaration empêche également l'affichage : 'écran' est le type par défaut, et si vous ne précisez pas le type, il sera implicitement défini sur 'écran'. La première et la deuxième déclaration sont donc en fait équivalentes.
```
igrigorik: Good catch.

igrigorik:
@PontusHorn ok, with you know. LGTM, and thanks for the fix!

PR #1240, opened 2015-05-04
Use /bin/bash over /bin/sh

Garbee:
The docker tool script currently doesn't run in a fresh Ubuntu 15.04 build from what I can tell. The script runs against *dash* in this circumstance which fails, switching to directly use bash fixes this.

### DIFF
--- a/tools/docker.sh
+++ b/tools/docker.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 image=${WF_DOCKER_IMAGE:-google/webfundamentals}
 cmd=$1


wibblymat:
@crhym3: making it run with `/bin/sh` would be more portable, but I'll let you decide.

crhym3:
yeah, I would agree with @wibblymat. 
@Garbee can you post the error here?

Garbee:
Output is:

```
tools/docker.sh: 6: tools/docker.sh: [[: not found
tools/docker.sh: 10: shift: can't shift that many
```

This is from where Ubuntu using /bin/dash for sh instead of bash.

crhym3:
Thanks @Garbee.
Lack of time. Merging it as is. If anyone wants it to be `/bin/sh` again, please create a PR.

PR #1241, opened 2015-05-04
Remove Install NodeJS step.

Garbee:
NodeJS is installed with nvm in step 3. This is a redundant instruction.


wibblymat:
Thanks!

Issue #760, opened 2014-09-12
About translation convention


chenxsan:
As far as I know, people speak English don't have such a problem. But in Chinese, we got this question: do I have to add a white space between Chinese character and English word or Chinese character and number, like `我是 Sam Chen，今年 100 岁` or `我是Sam Chen，今年100岁`. People in China don't agree on this.

When the translation really get started, you are going to see a lot variants, that would be a mess. And I suggest to document some conventions for such a language translation.

PR #1263, opened 2015-05-13
Add translation zh-cn for render performance

simongong:


chenxsan:
I once filed [this issue](https://github.com/google/WebFundamentals/issues/760) about translation convention, I'm not sure if there's one translation style guide now? @PaulKinlan. If not, then which style do this project prefer more? With white space between Chinese character and English word or without? I can see there are [white spaces between transitional Chinese and English word](https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-tree-construction?hl=zh-tw). We better keep one style instead of two.

PaulKinlan:
It seem's one space will be best.  Honestly, we haven't thought about it
too much.

PaulKinlan:
I'll merge this and if we need to change we can.

simongong:
@chenxsan @PaulKinlan 
As to the concern of `white space between mixing characters`, I vote for **one space** as well for the sake of better readability.
So I think it would be better to add a tip/guidance somewhere for future translation, especially for non-alphabet based languages.

PR #1270, opened 2015-05-20
Forms: Add sign-{in,up} autocomplete attributes.

mikewest:
Adding the following autocomplete attribute values which we use in Chrome's password manager:

* `username` (https://html.spec.whatwg.org/multipage/forms.html#attr-fe-autocomplete-username)
* `current-password` (https://html.spec.whatwg.org/multipage/forms.html#attr-fe-autocomplete-current-password)
* `new-password` (https://html.spec.whatwg.org/multipage/forms.html#attr-fe-autocomplete-new-password)

### DIFF
--- a/src/_langs/en/fundamentals/input/form/label-and-name-inputs.markdown
+++ b/src/_langs/en/fundamentals/input/form/label-and-name-inputs.markdown
@@ -239,6 +239,29 @@ The `autocomplete` attributes can be accompanied with a section name, such as **
         </ul>
       </td>
     </tr>
+    <tr>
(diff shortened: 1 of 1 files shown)


mikewest:
Hi @googlebot! I still work here, so far as I can tell.

greenido:
Yep @googlebot - I must agree with @mikewest 

@PaulKinlan not high priority - but can we do something with this? :) I remember it happened 'few' times in the past.

PaulKinlan:
Mike be signed up with the old org and system.

P

Issue #1245, opened 2015-05-06
[EN] Feedback for: /fundamentals/input/form/choose-the-best-input-type.html


brianmontana:
The highlight-module__cta class element has position absolute without a clearfix on the parent element. It's breaking outside the container and visually looks awkward.

PR #1282, opened 2015-05-22
Fix to cta button so it's not appearing outside it's container.

brianmontana:
[EN] Feedback for: /fundamentals/input/form/choose-the-best-input-type.html #1245

https://developers.google.com/web/fundamentals/input/form/choose-the-best-input-type

Fixes cta button appearing outside container.


brianmontana:
I signed it!

PR #1283, opened 2015-05-22
cta highlight button fix, so it doesn't appear outside it's container

brianmontana:
Breaks visual design flow on this page, looks appropriate on others without borders just below containers.

https://developers.google.com/web/fundamentals/input/form/choose-the-best-input-type


petele:
Thanks @brianmontana - this should get pushed to the live site either later today or sometime early next week.

brianmontana:
Thanks Pete!

PR #1288, opened 2015-05-25
fix Chinese translation

chenxsan:


petele:
Hey Sam - I had to make one change to this as it broke the build on developers.google.com - we can't have file names with unicode characters. I renamed your profile image to samchen.jpg and updated the associated files.  Sorry!

chenxsan:
@petele It's ok. Thanks.

PR #1300, opened 2015-05-29
changed the link of download button

vicky002:
In the current page The download button link is not working. It was redirecting to 404 page.


vicky002:
I signed it!

petele:
this has been fixed in another check in. thanks for the contribution, sorry for the mix up!

PR #1330, opened 2015-06-04
Add fetch() call to page speed recommendations.

jpmedley:


PaulKinlan:
Should we caveat this with information about it only working in chrome.

petele:
Yes - that would be good.  @jpmedley can you please update and resubmit?

PR #1336, opened 2015-06-08
Create theming-elements.markdown

fantompower:
@gauntface new markdown for Polycasts is up!


petele:
Looks like there's been a small error in the markdown you've been submitting. I've fixed it the last few times, but can you make the change for this one and resubmit it?

The fix is simple, you're just missing two spaces before written_on and updated_on.

Correct syntax:

article: 
  written_on: 2015-06-02
  updated_on: 2015-06-02

Thanks,
PEte

PR #1339, opened 2015-06-09
Add fetch() call to page speed recommendations.

jpmedley:


src/_langs/en/fundamentals/performance/critical-rendering-path/page-speed-rules-and-recommendations.markdown
```diff
@@ -47,6 +47,43 @@ code shows how to use `navigator.sendBeacon()` to send data to the server in the
 </script>
 {% endhighlight %}
 
+If you're on Chrome 42 or later, you can use the `fetch()` method to request data asynchronously. This method processes responses with Promises rather than multiple event handlers. Unlike the response to an XMLHttpRequest, a fetch response is a stream object starting in Chrome 43. This means that a call to `json()` also returns a Promise.
```
petele: Hey Joe - developers may be on Chrome 42, but they won't know if their users are on Chrome 42, so instead of phrasing it this way, can we rephrase to something along the lines of...

The new `fetch()` method provides an easy way to asynchronously request data. Because it is not available everywhere yet, you should use feature detection to test for its presence before use. The `fetch()` method processes responses...

petele:
Sorry to send this back again! Please ping me if my comment doesn't make sense.

jpmedley:
Is there a way to test whether a response is a stream? My preferred way to
handle the change from 42 to 43 is also to say 'test for that'.

Joe Medley | Technical Writer, DevPlat | jmedley@google.com | 816-678-7195

petele:
Hmm, @jakearchibald - do you know the answer to @jpmedley's question?

jpmedley:
I'm going to merge this. We can fix it later if somebody thinks of a better way to deal with the streams issue.

PR #1342, opened 2015-06-09
Add stricter syntax to run on Linux

jpmedley:
Here are two places where the syntax was changed so that the sass task will run on Linux. I have confirmed that this does not break on Mac. To demonstrate I've pulled the following fragments out of DevTools when running from Docker. Note that these are the same as what is output without the changes.

@import url(//fonts.googleapis.com/css?family=Roboto:400,300,500);

@-webkit-keyframes showNav {
  0% {
    opacity: 0
  }

  100% {
    opacity: 1
  }
}

@-moz-keyframes showNav {
  0% {
    opacity: 0
  }

  100% {
    opacity: 1
  }
}

@-ms-keyframes showNav {
  0% {
    opacity: 0
  }

  100% {
    opacity: 1
  }
}

@keyframes showNav {
  0% {
    opacity: 0
  }

  100% {
    opacity: 1
  }
}

### DIFF
--- a/src/_sass/_utils.scss
+++ b/src/_sass/_utils.scss
@@ -151,19 +151,19 @@
 
 
   @mixin keyframes($animation-name){
-    @-webkit-keyframes $animation-name{
+    @-webkit-keyframes #{$animation-name}{
       @content;
     }
 
-    @-moz-keyframes $animation-name{
+    @-moz-keyframes #{$animation-name}{
       @content;
     }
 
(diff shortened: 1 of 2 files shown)


jpmedley:
I assume it's not good to merge my own requests, so I have to ask. Is this something you guys don't want to do?

PaulKinlan:
I would merge yourself.  I trust you.

PR #1371, opened 2015-06-16
Helper script to publish WF samples to GitHub pages

petele:
@gauntface @PaulKinlan - can I get a quick code review/LGTM on this.

Background - I almost always forget to publish the samples after pushing to DevSite, this separates that logic out and makes it easier to do it in a quick, one off way. The code is essentially the same (copied/pasted) from the DevSite script with a few minor modification.

I've left the original logic in the DevSite script so it can still be used there if necessary, but this makes it easier for one of pushes and doesn't require G3/DevSite/etc.

### DIFF
--- a/tools/publishSamples.sh
+++ b/tools/publishSamples.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+# Script to sync the GitHub web fundamentals samples contents onto GitHub Pages.
+# Run from the location of a clone of the fundamentals repo.
+
+WEBDIR="$(pwd)"
+GHDIR="/tmp/WF-gh-pages"
+
+# - are we in the web dir?
+if [ ! -f "Gruntfile.js" ]; then
+  echo "Please make sure you're in the git checkout location."
+  exit
+fi
+
+echo "----- Getting latest sources from GitHub"
+
+# Git pull
+git checkout master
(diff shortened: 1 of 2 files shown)


tools/publishSamples.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+echo "----- Getting latest sources from GitHub"
+
+# Git pull
+git checkout master
+git pull https://github.com/Google/WebFundamentals master
+LASTCOMMIT=$(git rev-list HEAD --max-count=1)
+
+if [ $? -ne 0 ]; then
```
gauntface: Any chance of a comment explaining this does? To a shell noob like me, that is nonsense.

tools/publishSamples.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+# Script to sync the GitHub web fundamentals samples contents onto GitHub Pages.
+# Run from the location of a clone of the fundamentals repo.
+
+WEBDIR="$(pwd)"
+GHDIR="/tmp/WF-gh-pages"
```
gauntface: Could you rename this to GHSAMPLESDIR <- Just to be a tad more descriptive.

gauntface:
Overall LGTM - minor nits over naming and comments.

Issue #1385, opened 2015-06-22
_code folders do not support subdirectories


petele:
If a `_code` folder contains its content in subdirectories, the build fails when building the Samples list page.

Error seems to occur on line 55 of `sample_builder.rb`.

This will need to be fixed to support adding additional samples to the updates directory

petele:
Very low priority - going to punt.

PR #1386, opened 2015-06-22
Flattened directory structure

samdutton:


petele:
This is a work around for #1385

PR #1413, opened 2015-07-06
Fixes to bad links in header

petele:
DO not integrate until @petele gives the all clear. This update changes EVERY file in the system and will be fun to deploy.


jpmedley:
I'm build cop this week, but would like to have another coder look at this before you commit. (I'm a writer.)

PR #1446, opened 2015-07-17
RSS + ATOM feeds for UPDATES

petele:
This is a temporary fix to provide full content RSS & ATOM feeds for /updates/ and it *only* affects **updates**.

Essential changes:
* adds `atom.liquid` and `rss.liquid` 
* validated output of atom and rss feeds against the W3C feed validator
* adding the feed links to the document head is a hack: in `_includes/head.liquid`, it checks if the file path contains `updates` if it does, it uses the new feed links.
* modified the `UpdatesFeedPage` in `update_generator.rb` to take a `template` and `fname`, then when it is called, it generates the feed page based on the `template`, and outputs it to `fname`.
* For *backwards* compat it still generates the old rss feed, but that should get yanked.

### DIFF
--- a/src/_layouts/atom.liquid
+++ b/src/_layouts/atom.liquid
@@ -1,0 +1,39 @@
+---
+layout: null
+spec: https://tools.ietf.org/html/rfc4287 
+---
+<?xml version="1.0" encoding="utf-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom">
+  {% assign pages = page.items | deepsort: 'article', 'updated_on' | reverse %}
+  {% assign tld = "https://developers.google.com/web" %}
+  <title>{{ page.title | xml_escape }} - Google Developers</title>
+  <subtitle>{{ page.description | xml_escape }}</subtitle>
+  <updated>{{ site.time | date_to_xmlschema }}</updated>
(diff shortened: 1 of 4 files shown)


src/_layouts/atom.liquid
```diff
@@ -0,0 +1,39 @@
+---
+layout: null
+spec: https://tools.ietf.org/html/rfc4287
```
PaulKinlan: Any need for this?

src/_layouts/rss.liquid
```diff
@@ -0,0 +1,42 @@
+---
+layout: null
+spec: http://cyber.law.harvard.edu/rss/rss.html
```
PaulKinlan: Any need for this?

PaulKinlan:
Other than minor comments, LGTM.

src/_layouts/atom.liquid
petele: No - it's more of a comment so that people updating the feed can easily refer back to the atom spec.

src/_layouts/rss.liquid
petele: No - it's more of a comment so that people updating the feed can easily refer back to the RSS spec.

PR #1447, opened 2015-07-17
DevTools: Daily DevTips

pbakaus:
Two DevTips imported from Umar's DevTips into /web/updates.


pbakaus:
I confirm I authored these commits. (for what it's worth, I think I finally figured out how to fix that CLA message...next pull request should be fine).

pbakaus:
@petele could you take a quick look. These are just two new DevTips.

PR #1453, opened 2015-07-20
DevTools: Two new DevTips imported

pbakaus:


pbakaus:
WOOT, the CLA finally went through :)

PR #1465, opened 2015-07-21
First commit for moving new IA into Material Branch

petele:
@gauntface - this is the first cut of the new IA for the material branch. 

Included updates:
* Updates `_betterbook-wf.yaml` to the new IA
* Adds new basic landing pages for each section
* Moves some `getting-started` content to new `guides` section


gauntface:
Accepting this PR - but first thing I will need to do is rename the layout ;)

petele:
@gauntface - yep, the layout name is meant only as temporary until we get the actual layout we want to use in place. i guess i should have called it `TODO-this-is-just-a-placeholder-until-we-get-the-layouts-figured-out.liquid` ;)

PR #1470, opened 2015-07-22
Fix some wording nits in the Web Bluetooth article.

jyasskin:
@beaufortfrancois Here's my pass over the article. I haven't compiled it to HTML, so you should double-check that I didn't introduce any syntax errors.


beaufortfrancois:
LGTM!
@petele Can we merge this please?

PR #1480, opened 2015-07-24
some fixes

chenxsan:


petele:
@chenxsan - in future, would you please be a bit more descriptive about what you're fixing? "minor fix" and "another fix" isn't very clear ;)

chenxsan:
@petele ok.

PR #1484, opened 2015-07-24
DevTools: Two new imported DevTips

pbakaus:
This also fixes a small issue with the Admin Importer, making the match of what tips where already posted a little more fuzzy (slash vs. no slash at end of original source url) so they don't reappear in the importer.

@petele since this isn't original content, I'm unsure whether the instruction of getting a lgtm from a writer applies here as well. Thoughts?


petele:
@pbakaus - yes, you still need an LGTM, even if it's not original content. We also should make sure that we have a CLA on file for Umar.

petele:
Again, will merge today, but this'll be the last time ;)

pbakaus:
OK thanks for the clarification. Will get a writer on to it next time!

PR #1488, opened 2015-07-27
Updates: Small fixes/links for DevTools Digest

pbakaus:
This just adds a few links where there weren't any.


Meggin:
lgtm from writer. I accepted pull request too-- confirming with Pete that this is the right thing to do in these reviews?

PR #1502, opened 2015-08-05
Add completed /tools section

pbakaus:
This pull request merged the /tools branch and does the following:

- Add new DevTools homepage
- Add all new tutorials ported from the original DevTools homepage
- Removal of /web/fundamentals/tools section in favor of /web/tools

A number of smaller infrastructure have been done:

- Updates now support digest cards (being used in the DevTools Digest posts, fully backward compatible)
- "$ grunt develop" now accepts multiple sections to build, and largely do not depend onto each other. Doing "$ grunt develop --section=tools" will only build tools and executes in less than 5 seconds.
- Building a sub section and exiting Grunt early could sometimes not move the files back into their original place. This logic is a lot safer now and can repair itself if needed.

@Meggin and I spent the entire last three workdays to finalize the content and ensure that everything is in good state, the branch is good to go from our side.


pbakaus:
Confirm that I authored all commits (along with meggin). The CLA bot failed because I did some commits with my chromium.org email address and it wasn't yet associated with Github, apparently.

PR #1512, opened 2015-08-11
Web Manifest Density is a float, not a string

beaufortfrancois:
As per https://w3c.github.io/manifest/#density-member

R: @PaulKinlan @petele


beaufortfrancois:
There are a lot of confusion out there regarding density string. 
I'm trying to fix these. See https://github.com/GoogleChrome/multi-device/pull/21 for instance.

PR #1525, opened 2015-08-15
Publishing two new spotlights and inserted the developer name

pbakaus:
Only dates have been changed, plus the name of one dev was missing. Should be good to merge.


pbakaus:
ping!

PR #1528, opened 2015-08-18
no relation with backend

shailendra9:
web fundamentals are for only web . there is no relation with any type of backend server and mongodb


PaulKinlan:
Thanks for the PR.  Sorry but this isn't pertinent to the project.

shailendra9:
you welcome Paul

PR #1555, opened 2015-08-26
Updates to Animations; adds mentions of Web Animations with updated demos

samthor:
@paullewis as per our email thread.

Let me know if you feel the changes work. I'm not sure if I should try to do some more serious refactoring. Thanks!

### DIFF
--- a/src/_langs/en/fundamentals/look-and-feel/animations/_code/box-move-elastic-wa.html
+++ b/src/_langs/en/fundamentals/look-and-feel/animations/_code/box-move-elastic-wa.html
@@ -1,0 +1,85 @@
+<!doctype html>
+<html>
+
+  <head>
+	  <meta charset="utf-8">
+	  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+	  <meta name="viewport" content="width=device-width">
+	  <title>Simple Box Move</title>
+	  <style>
+	  	html {
+	  		background: #f2f2f2;
+	  		width: 100%;
+	  		height: 100%;
+	  	}
+
+			.button-container {
+			  text-align: center;
+			}
+
+			.box-mover {
+			  background: #4285f4;
+			  border: none;
+			  padding: 10px 12px;
+			  border-radius: 3px;
+			  color: #FFF;
+			}
+
+			.box {
+			  position: absolute;
+			  width: 162px;
+			  height: 100px;
+			  left: 50%;
+			  top: 50%;
+			  margin: -50px 0 0 -81px;
+			  border-radius: 2px;
+			  border: 1px solid #EEE;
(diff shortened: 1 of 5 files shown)


samthor:
re: #1556

paullewis:
At first blush this looks good, but I'd love to read it all in context. Wondering what the fastest way would be to do that.... Hmmm...

samthor:
There are a lot of dependencies. I can try to set it up, but it looks geared for Mac, so I can't then share that link with you - unless I can get it working on *nix.

samthor:
@paullewis Any luck bringing up a demo?

paullewis:
Fiiiiiiinally got this to build. Okay, notes:

* /web/fundamentals/look-and-feel/animations/css-vs-javascript?hl=en 
  * feels like it should have more about Web Animations mentioned near the video embed. It mentions CSS and JS, but we should talk WA up around the JS option, i.e. "You may want the control of JS, but without including a huge animation library... ta-daaaa"
  * I don't think we need the polyfill embed code, and we don't want to blind include it, either, since it's a connection we don't need to make. If anything it should be lazy-loaded, but only when we know we need it. The same is probably true for where I mention including TweenMax!
  * If we mention the Firefox flags, mention how to get there (I think about:flags, but worth checking)
  * I don't think we need "This API is built into the browser, and as such, is more performant than a framework-based solution." but if you really want to include it I think you need to include data. For example, for a simple animation, you're going to see 60fps no matter how you do it. I think we should drop it unless we want to get into the specifics a bit more.
  * Final paragraph: "Since with JavaScript animations..." feels very out of place.
* /web/fundamentals/look-and-feel/animations/animations-and-performance?hl=en
  * Nit: you added a point, but it still says "couple", i.e. "Here are a couple of points to keep in mind" when it should be few.

You should add yourself as an author.

samthor:
Ok! I've added myself as an author on the pages I've changed, let me know if you think I should do it to every page.

* `css-vs-javascript`
  * I've expanded the top section to mention WA more prominently but without dictating that it's the only way (hopefully).
  * I've dropped the code itself, but kept a link to the polyfill. (It's on my TODO list to make the polyfill site a bit more accessible, with CDN links and the like, but I've not there yet).
  * So, it looks like Firefox isn't really ready yet, despite the flag - it just exposes active CSS animations as Web Animations, which while interesting, isn't really what we're talking about. You can't create new ones. I've changed this to indicate that it's in development.
  * I've dropped "This API is built...", in favour of fewer paragraphs.
  * The "Since with..." paragraph I didn't actually change from your version :)
  
* `animations-and-performance`
  * Done, I've changed to 'few'.

samthor:
Thanks!

PR #1576, opened 2015-09-02
Remove duplicated native install code snippet

mahemoff:
Maybe I missed something, but seems like there was some unnecessarily redundant duplication here.


PaulKinlan:
I added this in so people can just copy and paste if they don't want the normal part of the manifest.

mahemoff:
I see exactly the same code block twice on https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android#native

By "normal" I think you mean the one for web app, ie you mean you wanted to explicitly show the native snippet. Which I agree is good, but I think you've done so twice.

PaulKinlan:
There are extra curly braces in the second one.

mahemoff:
I see. I did miss them. So I think you're saying that the second example is the whole manifest file - that part wasn't so clear from the preamble.

PaulKinlan:
Yes. I will clear this up though.

PR #1579, opened 2015-09-03
update symbols and add links to Web Animations updates

samthor:
@PaulKinlan @petele 

Just tweaking a few older Web Animations posts to fix naming.


PaulKinlan:
Ta dude.

PR #1634, opened 2015-09-10
Fix the link to the Bluetooth logs-gathering instructions.

jyasskin:
I moved it after Francois added the link. Sorry.

@petele @beaufortfrancois


PaulKinlan:
We will get it live asap.

PR #1746, opened 2015-09-18
fixed broken devtools link in main nav of tools

pbakaus:
..what the title says :)


pbakaus:
@gauntface can you take a look and merge?

PR #1751, opened 2015-09-18
add interview video link to the CDS 2014 case study

pbakaus:
@gauntface this needs to be merged and pushed before 09/22. Thanks much!


petele:
@pbakaus - please do not assign people, otherwise it might sit for a long period of time; also what's up with the 9/22 deadline? We're heads down on the redesign.

pbakaus:
happy to leave unassigned in the future. The 9/22 deadline is due to the fact that it correlates with a social push for the finally released video on that case study so it would be great to have it on the case study page by then.

pbakaus:
thanks Pete.

PR #1807, opened 2015-09-23
Better compression

jakearchibald:


paullewis:
:headphones: SGTM :headphones:

PR #1849, opened 2015-09-24
Meggin's suggestion on IA tweaks

petele:


petele:
@Meggin - these changes should be live in about 10 or 15 minutes.

PR #1871, opened 2015-09-25
First cut of mixed content

petele:


petele:
@johyphenel fyi - this is the push to the new branch.

PR #1902, opened 2015-09-28
Material typography fixes

pbakaus:
FTW!

### DIFF
--- a/src/static/styles/partials/_mdl-overrides.scss
+++ b/src/static/styles/partials/_mdl-overrides.scss
@@ -8,14 +8,14 @@
 .mdl-data-table td, .mdl-data-table th {
   text-align: left;
   font-size: 15px;
-  line-height: 1.625;
+  line-height: 1.625em;
   padding-bottom: 12px;
   padding-top: 12px;
 }
 
 h2 {
 	font-size: 30px;
-    line-height: 1.2381em;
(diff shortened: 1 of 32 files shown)


src/static/styles/partials/_mdl-overrides.scss
```diff
@@ -1,7 +1,52 @@
 .mdl-data-table {
   white-space: inherit;
+  font-size: 15px;
+  line-height: 1.625;
```
petele: add em's

src/static/styles/partials/_mdl-overrides.scss
```diff
@@ -1,7 +1,52 @@
...
 .mdl-data-table td, .mdl-data-table th {
   text-align: left;
+  font-size: 15px;
+  line-height: 1.625;
+  padding-bottom: 12px;
+  padding-top: 12px;
+}
+
+h2 {
+	font-size: 30px;
```
petele: tabs/spaces

src/static/styles/partials/_mdl-overrides.scss
```diff
@@ -1,7 +1,52 @@
...
+    line-height: 1.2381em;
```
petele: how about pi instead? something rounder if possible.

PR #1979, opened 2015-10-01
Fixes build break

petele:


petele:
@gauntface - this backed out the svg-minified changes as the svg-minified directory wasn't checked in. Feel free to revert once it's checked in.

PR #2013, opened 2015-10-02
IOWA case study cleanup

jeffposnick:
R: @petele @gauntface 

This removes an errant `<pre>` and `title: ` that was left behind after the IOWA case study content was migrated out of the YAML frontmatter.

There's also a bunch of whitespace that my editor auto-removed from the ends of lines, which is good, but makes for a noisy diff. It's easier to read if you ignore whitespace with https://github.com/google/WebFundamentals/pull/2013/files?w=1


petele:
The extra line spaces may have been me, thanks for cleaning up!

PR #2031, opened 2015-10-05
Fixes the jaggies in the service-workers-iowa card image

jeffposnick:
R: @petele @gauntface 
CC: @ebidel

It looks like the card image acquired some jaggies in https://github.com/google/WebFundamentals/commit/983a935b0fbf1c3c5745800e9fd4d03e9ce61c43#diff-b3bc49d0f245eb6b06ac84e2eed75393 when it was resized.

I've take the correctly-sized base image from the CDS case study and pasted in the IOWA screenshot over it, resulting in what should be an appropriately sized card without the jaggies.


ebidel:
LGTM

PR #2072, opened 2015-10-11
Fixed html entity typo on 2013-01-14-Voice-Driven-Web-Apps-Introduction-to-the-Web-Speech-API.markdown

imagentleman:


PaulKinlan:
Good spot.

Just as a quick note, we are about to do a major upgrade of the site and we will try and pull this in after that (over the next day or two)

PR #2074, opened 2015-10-12
Making url lowercase for news / devtools

gauntface:


petele:
@gauntface - we can close this now that we've made the other changes, correct?

gauntface:
Yes we can

PR #2096, opened 2015-10-13
Revision of Service Worker Primer

jpmedley:
This may not be ready to merge. I've addressed nearly everything in Jake's review, but I'm waiting on answers to a few questions.


petele:
OK - will hold off until you say it's ready

PR #2112, opened 2015-10-14
Fix heading in "Discard Tabs in Chrome" update article

kaycebasques:
image


PaulKinlan:
@kaycebasques this will need to be done against the material-branch

petele:
@kaycebasques - as @PaulKinlan mentioned, we've moved to the material-branch full time now. Can you make the change there please?

PR #2043, opened 2015-10-06
Port and update snippets content from DCC

kaycebasques:
@Meggin or @pbakaus please review

* Ported old [snippets content](https://developer.chrome.com/devtools/docs/authoring-development-workflow#snippets) from DCC.
* Added some nifty animations.

Issue #1890, opened 2015-09-28
Animation logic needs to be ported to material-design


petele:
@pbakaus's animation stuff is potentially necessary for animations.

gauntface:
Can you share more docs about what this is? @petele @pbakaus

pbakaus:
In the DevTools docs, as well as in the update section, we have the need to embed automatically running animations in some places that show a certain feature.

Animated gifs are big, have low frame rates to manage size and limited colors, plus eat a lot of memory and CPU power on mobile. The canvas based approach we've used in the old DevTools docs was efficient but a lot of JS, and every animation needed a script.

The solution I ended up implementing is the most simple I could come up with: Normal videos on the page and a small amount of JS that detects if you're on desktop, then hides the controls, sets loop to true and autoplays a video as soon as it comes into the viewport (and stops others). This ensures that your bandwidth isn't eaten and you always see the right animation running. On mobile, said logic simply does not run, and so you get normal videos that can be played back on demand.

kaycebasques:
+1

gauntface:
Personally not ok with the data hit on users especially considering serving
up mp4s isn't going to use adaptive streaming.

Is YouTube an option?

auto-play still seems like a stretch to me. I think We've had reports of
people suffering from regularly migraines when landing on a page with lots
of moving images (don't know if this was a gif or a video however)

pbakaus:
The migraine issue came from the tips section where we've had lots of animated gifs at the same time. It was not about the video animations.

The original script for the animations was carefully designed with performance in mind. It never ran on mobile and would always only auto-play a video that's currently in the viewport. Thus, the only data hit on users happens when they *actually* scroll to the video and want to see it on a *Desktop* computer (with the exception that the browser might do some funky preloading).

If YouTube can auto-play and loop then it might be an option but I worry that the overhead on page performance compared to a single video tag is significant. Would require the YT API to control playback etc.

Needless to say, I've given this a lot of thought.

PR #2113, opened 2015-10-14
Add Snippets guide

kaycebasques:
hallo @pbakaus 

* Continuation of #2043
* New commits address last round of feedback from #2043, see that PR for inline discussions
* Animations don't work, see #1890
* Question: I really don't know where to house this. Snippets resides in the Sources panel. It doesn't make sense to store it with the Command Line docs. I've stored it top-level in Debug, but I don't know how or where we would display it. It's currently buried / hidden on the Debug index page.


pbakaus:
Nice!

Instead of putting it directly in the debug/ folder, create a sub directory called "snippets", put the markdown file there and rename it to "index.markdown". Then change the title from "Set Up Snippets" to "Using Snippets" (or something even better). This will make it appear in the left nav.

Unfortunately WF currently isn't setup to skip the forth indentation, so on the snippets page you'll see an empty "topics" block at the bottom. My guess is that this will be fixed eventually and is fine for now.

Regarding the video descriptions: I still don't like em. @Meggin feel free to chime in here. I don't think they have a lot of value. I propose the following: Let's always make sure that the text on the page works standalone without videos, and use the videos as "bonus visualization". This gets us to the same place without awkwardly describing the contents of a video.

paulirish:
"Using JavaScript Snippets"  and i'd position it alongside any Using the Console stuf.

petele:
@pbakaus - when this is ready to go, please add the 'lgtm: approved' label

kaycebasques:
OK @pbakaus and @paulirish all items addressed, and then some. See last batch of commits.

pbakaus:
LGTM, nice work!

petele:
Merged, will most likely push a new build tomorrow.

PR #2127, opened 2015-10-15
Switch the featured case study to SW in Production

jeffposnick:
R: @gauntface @PaulKinlan all

Swaps out CDS 2014 case study for the I/O 2015 as featured on the landing page.

image


PaulKinlan:
+1, although this is likely to change again next week ;)

gauntface:
I would like this to be automated-a bug should be raised

Issue #2128, opened 2015-10-15
Broken images at https://developers.google.com/web/updates/2015/03/push-notifications-on-the-open-web


asos-giovannip:
All the images are broken in the article guys.

asos-giovannip:
Eg. https://developers.google.com/updates/images/2015-03-04-push-on-the-open-web/enabled-disabled-push-ux-chrome.png

Should be:
https://developers.google.com/web/updates/images/2015-03-04-push-on-the-open-web/enabled-disabled-push-ux-chrome.png

You miss /web/ in the path.

PaulKinlan:
Looking at it now.

PaulKinlan:
@gauntface it looks like {{site.baseurl}} is not set and it is used quite heavily on the site.

PaulKinlan:
@gauntface just fixing now WFBaseUrl

gauntface:
This PR has been merged (ta PK) so should be fixed in the next build - sorry it got through and thank you for the bug report :)

asos-giovannip:
Thanks to you guys @PaulKinlan @gauntface

PR #2129, opened 2015-10-15
Fixing the base url issue brought up by #2128

PaulKinlan:


PaulKinlan:
cc @petele and @gauntface

PR #2136, opened 2015-10-15
Remove blocked requests from DevTools Digest #3

pbakaus:
Feature was pulled after it was already in Canary. It's important that this goes out ASAP.


pbakaus:
Thanks Pete!

Issue #2143, opened 2015-10-16
Add language selection


schlamar:
Please add a way to select the language. I don't want to read the German version (see #976). But right now, there is no way to select the language besides changing browser defaults.

gauntface:
At the bottom of the page there is a language selection button.

Do you have any ideas that would have made this easier to have found?

![screenshot from 2015-10-16 10 15 19](https://cloud.githubusercontent.com/assets/139760/10538208/e178f44e-73ee-11e5-8187-8730400036d1.png)

schlamar:
Mhh, good question. And more interestingly, language selection doesn't work for me (Firefox 41.0.1).

gauntface:
I just tried it on 41.0.1 on ubuntu and it's working for me.

What are you seeing that's the issue??

schlamar:
Actually, there are even multiple issues:

1. Quite often the link just doesn't work. When I hover over a language there is no link shown in the browser status field (bottom left). After a click nothing happens. If I hover over a language multiple times a link is finally shown.
1. The English link has no argument in the URL (no `?hl=en`). So if German is shown as default language (no `hl` argument) and you click on the English link you are still on the German page. Manually adding `?hl=en` to the URL works.
1. Language selection should be naturally on top and not on bottom. You don't want to scroll the complete text in the "wrong" language.

gauntface:
@schlamar thanks for the feedback.

Thanks for the feedback. 1.) I've fixed in CSS and will be commiting soon.

Regarding number 2.) if your browser or computer set to German? Everything from what I can see defaults to english if no hl is set (which is why english doesn't have one). Just wondering if there is a way I can reproduce what you are seeing so I can trigger it, guarantee and fix and document it as well.

@petele what's your thoughts on comment 3.

schlamar:
re 2): Yes, my preferred language in the Browser is German.

There is even an link on the index page: Out of Date! This article's translation is out of date with the original version. Clinking on this link brings me back to the German version, too.

I guess this comes from the HTTP Header `Accept-Language:"de,en-US;q=0.7,en;q=0.3"`

gauntface:
@schlamar thanks for all the help.

Yeah looks like the production environment is taking into account that header where as dev and staging aren't.

I've got a PR open for this fix so hopefully that'll go through and fix up all links to include the hl for english versions.

petele:
Closing as this has been pushed live. Please reopen if it's still an issue.

PR #2158, opened 2015-10-16
Adding hl=en to the page URLs and making language selection full heig…

gauntface:
…ht of list items

@petele @PaulKinlan this is to address: #2143

This does mean all links will have a language code. The canonical URLs will not change with this (i.e. canonical urls in head of pages will not have a hl query param).

Is there any negative side effects to this change?


PaulKinlan:
This replicates the mostly the original way that we handled translated content in WF.

PR #2173, opened 2015-10-19
Add article about display-mode

kenchris:
Add article about display-mode which is already shipping


kenchris:
@PaulKinlan  you somehow need to confirm :)

PR #2174, opened 2015-10-19
Finishes S2 of HTTP 203; adds S3

paullewis:
PTAL @gauntface @petele

### DIFF
--- a/src/content/en/shows/http203/series-2/http-203-utopia-vs-dystopia.markdown
+++ b/src/content/en/shows/http203/series-2/http-203-utopia-vs-dystopia.markdown
@@ -1,0 +1,12 @@
+---
+layout: shows/episode
+title: "2.6. Utopia vs. Dystopia"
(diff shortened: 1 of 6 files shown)


PaulKinlan:
@gauntface don't we auto generate all this stuff now?

gauntface:
No we don't auto generate this.

Can't come up with a way to accommodates everything. For example, if we autogenerated everything, how would podcasts sit with all of this?

PaulKinlan:
Right, but this isn't the podcast... not bothered either way really just that I know we argued about it.

gauntface:
Yeah it's hard. I have a gulp task that is a GUI for creating show series
and pages

paullewis:
Actually it's not the podcast; that's going to be a separate PR. This is just for the video content.

gauntface:
Podcast was just an example that makes auto-generation difficult - that was all :)

PR #2179, opened 2015-10-20
Material branch

paullewis:
R @gauntface


src/jekyll/_plugins/custom_tags/audio.rb
```diff
@@ -0,0 +1,59 @@
...
+    def render(context)
+      # If the audio source is empty then this tag shouldn't be used
+      raise "Missing source for audio tag" if @audioUrl.empty?
+
+      # Expand this page variable to be its actual value.
+      if not context[@audioUrl].nil?
+        @audioUrl = context[@audioUrl]
+      end
+
+      out = '<audio class="embedded-audio" src="' + @audioUrl +  '" controls></audio>'
```
petele: @paullewis - can you check how much Chrome will preload here and potentially add the preload=none attribute if it's more than a few kb?

src/jekyll/_plugins/custom_tags/audio.rb
```diff
@@ -0,0 +1,59 @@
+# Copyright 2014 Google Inc. All rights reserved.
```
petele: Is there a reason why you did this as an audio tag vs an include? Personally I prefer includes, but not set absolutely.

src/jekyll/_plugins/custom_tags/ytvideo.rb
```diff
@@ -45,6 +45,7 @@ def render(context)
       # If the video source is empty then this tag shouldn't be used
       raise "Missing source for video tag" if @videoId.empty?
 
+      # Expand this page variable to be its actual value.
```
petele: If this is a TODO, please file an issue and reference it here.

src/jekyll/_plugins/custom_tags/audio.rb
paullewis: Oh I'll just add preload=none. Would rather we did everything on demand.

src/jekyll/_plugins/custom_tags/audio.rb
paullewis: Just to be in-keeping with the ytvideo embed code, and @gauntface said it was his preference, though I guess I can leave you two to fight that out? :runner:

src/jekyll/_layouts/shows/show.liquid
```diff
@@ -24,10 +24,18 @@
       Previous Seasons &amp; Episodes
     </h6>
     {% for subdirectory in page.context.subdirectories reversed %}
+      {% comment %}
+        TODO(gauntface): find better solution for podcasts in shows.
```
petele: Please file an issue for this and reference it in the code

src/jekyll/_plugins/custom_tags/audio.rb
petele: :facepunch: - nah, that's too much work. if it's his, then we'll go with that.

gauntface: *fist bump*

Generally the YouTube got out of hand and an include is to easy for someone
to jump in an alter, also include may be missing something and will still
print nothing, tags give easier error checking and throwing.

petele:
LGTM - merging.

PR #2193, opened 2015-10-21
updates data-video-id and 80 char lines

petele:
Splits some content into 80 char lines, but primarily changes `data-video-id` to the appropriate id.


gauntface:
LGTM

PR #2197, opened 2015-10-21
Moving podcast

jakearchibald:
I've uploaded the podcasts to cloud storage. Once this is merged & live, we can remove the mp3s from the repo.

### DIFF
--- a/src/_langs/en/shows/http203/podcast/feed.xml
+++ b/src/_langs/en/shows/http203/podcast/feed.xml
@@ -21,8 +21,8 @@
     <itunes:subtitle>Paul and Jake talk about CORS, Layout, and which of them has the evolutionary advantage.</itunes:subtitle>
     <itunes:summary><![CDATA[Why does nobody seem to include CORS headers on their files? And can Paul answer Jake's dreaded CORS pre-flight quiz?]]></itunes:summary>
     <itunes:image href="https://developers.google.com/web/shows/http203/podcast/http203-podcast-art.jpg" />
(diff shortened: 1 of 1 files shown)


petele:
@jakearchibald - merged!

jakearchibald:
So, we now have a choice. We can accept that there's 60mb in the repo for episode 1 & 2, and chalk it up to "let's not do that again". Or we can do something more drastic.

The mp3s can be purged from the project history. This will mean recreating all commits of the project, so anyone contributing will have to rebase their against against the new origin/master since it'd be a new history branch.

The command to do this is

```sh
git filter-branch --force --index-filter \
'git rm --cached --ignore-unmatch src/_langs/en/shows/http203/podcast/episode-*' \
--prune-empty --tag-name-filter cat -- --all
```

…which will purge the mp3s from your local checkout, the you can push that via:

```sh
git push origin --force --all
git push origin --force --tags
```

This is basically a subset of the steps in https://help.github.com/articles/remove-sensitive-data/.

Don't do this until the XML from this PR is live though!

petele:
@gauntface - is this something that we can do when we move from material-branch back to master?

gauntface:
I would probably do it before tbh. I'm nervous enough about the master move as is.

petele:
OK - sounds good. We'd probably be smart to tell (most) people to blow away their repo and start again. I'm going to create an issue to "clean up repo" so we can also remove some of the other huge assets.

gauntface:
Gone through the podcast files and purged them from the history and commited.

PR #2199, opened 2015-10-21
Fixing the broken layout of the tables

greenido:
I hope the header of the table in the markdown="block" will look good.


petele:
Hey @greenido - can you make sure that the tables are consistent, they each have a thead and a tbody, and if you're going to use ULs in one, use it in the other?

petele:
Closing as this is now in the old branch.

Issue #2058, opened 2015-10-07
Move MP3s/PodCast files to Google Cloud Storage


petele:
Right now the MP3 files are sitting in the Git repo, we should move them to a separate storage location that has better deployment support and has the CORS headers.

cc - @paullewis

PaulKinlan:
Hmm. I am loath to have things in another place. If we do this we need to
maintain control over them and have them version controlled in once place
too.

petele:
Understood, and I agree about the control part. My biggest concern right now is that by putting all of these MP3 files into the repo, we're going to blow the repo up. It's already pretty huge, the first pull is PAINFUL.  We also have no control over the headers and can't add CORS headers to the files right now, which makes Paul's podcast player a sad panda.

gauntface:
@paullewis are you happy to take this on?

paullewis:
Can you assign to @jakearchibald instead? He said he's happy to do this.

jakearchibald:
https://github.com/google/WebFundamentals/pull/2197

PR #2201, opened 2015-10-21
Updates podcast mp3 paths.

paullewis:
PTAL @petele ref: #2197 and #2058 

FYI @jakearchibald


petele:
Good catch - thanks @paullewis

PR #2206, opened 2015-10-22
Keep GUID static

jakearchibald:
Feels like we should keep the GUID static, but update the enclosure url. What do you think @paullewis?


paullewis:
Oops, yep. My mistake.

R @petele

PR #2231, opened 2015-10-26
Adds translation_priority to files

petele:


madroneropaulo:
So, what do the different priorities values mean?

petele:
It lets us specify how important it is to get a document translated. Priority 0 means it's critical, 1, important, etc.

PR #2237, opened 2015-10-26
add Web Animations Resources update

samthor:


PaulKinlan:
Suggest adding in a header image for the tile that is used in the list page of /updates/ ... Will look further tomorrow, but at a high-level, great to see this

PaulKinlan:
Ah. Just realised. This should be pulled against material-branch (sorry for confusion) the pre-amble has changed and needs to be fixed.

samthor:
Closed in favour of #2238.

PR #2239, opened 2015-10-27
Updated bluetooth javascript code in web-bluetooth article

beaufortfrancois:
Since https://codereview.chromium.org/1378013002, we don't need to check for `service` existence anymore.

R=@petele

### DIFF
--- a/src/content/en/updates/posts/2015/07/interact-with-ble-devices-on-the-web.markdown
+++ b/src/content/en/updates/posts/2015/07/interact-with-ble-devices-on-the-web.markdown
@@ -167,7 +167,6 @@ navigator.bluetooth.requestDevice({ filters: [{ services: ['battery_service'] }]
   return server.getPrimaryService('battery_service');
 })
 .then(service => {
-  if (!service) throw 'Battery Service not found';
   // Getting Battery Level Characteristic...
   return service.getCharacteristic('battery_level');
 })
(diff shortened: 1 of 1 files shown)


src/content/en/updates/posts/2015/07/interact-with-ble-devices-on-the-web.markdown
```diff
@@ -167,7 +167,6 @@ navigator.bluetooth.requestDevice({ filters: [{ services: ['battery_service'] }]
   return server.getPrimaryService('battery_service');
 })
 .then(service => {
-  if (!service) throw 'Battery Service not found';
```
petele: Even though we don't have to check, is it still good practice to check?

src/content/en/updates/posts/2015/07/interact-with-ble-devices-on-the-web.markdown
```diff
@@ -199,10 +198,7 @@ page](https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicV
 navigator.bluetooth.requestDevice({ filters: [{ services: ['heart_rate'] }] })
 .then(device => device.connectGATT())
 .then(server => server.getPrimaryService('heart_rate'))
-.then(service => {
-  if (!service) throw 'Heart Rate Service not found';
```
petele: Again - is it best practice to check for the service before returning? If not, no worries, happy to approve and merge.

src/content/en/updates/posts/2015/07/interact-with-ble-devices-on-the-web.markdown
beaufortfrancois: Nope, because if `service` is `null`, it will raise an exception that will be caught by the last `catch`.

src/content/en/updates/posts/2015/07/interact-with-ble-devices-on-the-web.markdown
beaufortfrancois: ditto.

petele: Perfect! thank you.

PR #2242, opened 2015-10-27
BTR

PaulKinlan:
R: @petele


src/content/en/showcase/case-study/beyond-the-rack.markdown
```diff
@@ -0,0 +1,71 @@
+---
+layout: showcase/casestudy
+title: "Beyond the Rack"
+subtitle: "Beyond the Rack re-engages users on the mobile web, increasing revenue per visit by 26% with push notifications"
+description: "&quot;<span style='font-style: italic;'>Push notifications allowed us to bring one of the most compelling capabilities from our native app to our mobile site. We see a direct 20% click through rate from push notifications—having another channel to reach our users is a game changer.</span>&quot; <br><br> &mdash;  Richard Cohene, VP Marketing, Beyond the Rack"
```
petele: any reason you're using `font-style: italic` instead of <i>?

src/content/en/showcase/case-study/beyond-the-rack.markdown
```diff
@@ -0,0 +1,71 @@
...
+for its 14 million members globally. Until recently, the company primarily depended on 
+targeted, daily emails to re-engage users. Recent internal research showed that **52%** of 
+users were visiting their site via the mobile web, so they looked for ways to improve 
+their mobile engagement and sales. At this point Beyond the Rack reached out to their 
+mobile shopping platform partner Mobify. Working with Mobify, Beyond the Rack implemented 
+push notifications on their mobile website and saw a **50%** increase in repeat visits within 
+3 months. The company also found push notifications delivered higher value visits, with 
+members spending **26%** more per visit on average.
+
+<img src="images/beyond-the-rack/btr-steps.jpg">
```
petele: Don't forget the alt attribute and figcaption if necessary.

src/content/en/showcase/case-study/beyond-the-rack.markdown
PaulKinlan: Because you can't have markdown in this part.

PaulKinlan: Sorry, I just looked at your code in the comment :) < i > is not the way we should be italicised as it is topical subtext.... thinking about it < em > is probably the best to use.

PR #2274, opened 2015-11-07
Apple status bar style doesn't support hex color

jrbasso:
According to [Apple's documentation](https://developer.apple.com/library/safari/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html) and some tests on the simulator the only options available are `default`, `black` and `black-translucent`.

Also, in order to Safari to use this meta information, the `apple-mobile-web-app-capable` needs to be set to `yes`.


jrbasso:
I signed it!

PR #2294, opened 2015-11-17
Add app shell to updates

gauntface:
@addyosmani


addyosmani:
LGTM. There's one last set of edits I've just pinged over but other than that this looks good to ship.

beaufortfrancois:
@addyosmani You may want to merge it to https://github.com/google/WebFundamentals/tree/material-branch instead I believe.

PR #2303, opened 2015-11-24
Ignore .bundle directory

kaycebasques:
I'm using a slightly different build / preview setup. Rather than RVM I'm just installing the bundle stuff via `bundler install --path vendor/bundle`, which automatically creates a config file at `.bundle/config`.


gauntface:
I can't see any obvious problems with this.

@petele and @PaulKinlan in case they can think of any thing

PR #2351, opened 2015-12-11
Switch from Travis to private drone instance

crhym3:
R: @PaulKinlan @petele


gauntface:
It looks good to me.

petele:
Sweet - besides merging this, is there anything else that needs to be done or will it just automagically switch over?

crhym3:
Yeah, it just works.

Both drone and travis kick in only when their config file (.drone.yml and .travis.yml respectively) is present. This PR removes .travis.yml and adds .drone.yml which effectively disables one and enables the other one.

Another way to verify this is click on "Show all checks" - you'll see Drone but you won't see Travis.

PR #2385, opened 2015-12-22
Adds the "Instant Loading with Service Workers" video embed

jeffposnick:
R: @PaulKinlan @petele @gauntface 

I don't know whether there's any editorial policy around video embeds on WF, but it seemed like the CDS 2015 video/deck would be a good resource under "Learn More" on the service worker libraries page.


PaulKinlan:
Contentwise I support this!! Have you checked the rendering?

jeffposnick:
image

gauntface:
Could you switch to the ytvideo include for the video?

PR #2386, opened 2015-12-23
add security panel guide

kaycebasques:
@pbakaus please review (doc is in correct location, etc.)

content has already been reviewed by two members of chrome security team

:christmas_tree:


PaulKinlan:
I believe Paul is on paternity leave now. I can help review.

kaycebasques:
ok @PaulKinlan please review!

src/content/en/tools/chrome-devtools/security/index.markdown
```diff
@@ -0,0 +1,93 @@
...
+A non-secure page is indicated with the message `This page is not secure.`
+
+The Security Panel distinguishes between two types of non-secure pages. 
+If the requested page is served over HTTP, then the main origin is flagged as 
+not secure. 
+
+![security overview, non-secure main origin](/web/tools/chrome-devtools/security/images/overview-non-secure.png)
+
+If the requested page is retrieved over HTTPS, but the page then goes on 
+to retrieve content from other origins using HTTP, then the page is still
```
Garbee: > content from other origins

Should read "content from any origin". Since, even if you serve the page over HTTPS, if you serve any resources from the same origin over HTTP (say by a hard-coded URL and it doesn't redirect) then it would still trigger mixed-content.

Garbee:
Doesn't mean much, but LGTM other than the one inline nit-pick. I'd rather we be absolutely specific on how things work here. Security is nothing to accidentally confuse someone on.

PaulKinlan:
@Garbee technically an origin includes scheme, so https is different to http, and means that the text is still accurate.

https://tools.ietf.org/html/rfc6454
>   Q: Why not just use the host?
>
>   A: Including the scheme in the origin tuple is essential for
>   security.  If user agents did not include the scheme, there would be
>   no isolation between http://example.com and https://example.com
>   because the two have the same host.  However, without this isolation,
>   an active network attacker could corrupt content retrieved from
>   http://example.com and have that content instruct the user agent to
>   compromise the confidentiality and integrity of content retrieved
>   from https://example.com, bypassing the protections afforded by TLS
>   [RFC5246].

All that being said, I do think it will be good to ensure that this is extremely clear.  @kaycebasques I will merge this but open another issue about @Garbee's comment.

PR #2414, opened 2016-01-05
Fix Issues with Push Notification Content

jpmedley:
@petele @PaulKinlan 

Gentlemen,

When the new /web debuted last fall, Pete merged content from two section address a deficiency in one. This PR does three things to correct that.

* Restores the Primer materials to the Primers section. 
* Provides a proper introduction to the subject for the Engage and retain section.
* Adds a new page to same.


src/content/en/fundamentals/engage-and-retain/push-notifications/personal-timely-relevant.markdown
```diff
@@ -0,0 +1,44 @@
+---
```
jpmedley: NOTE: This is not actually a new page. This is the content that was previously at src/content/en/fundamentals/engage-and-retain/push-notifications/index.markdown

jpmedley:
@PaulKinlan 

Have you had a chance to look at this?

PaulKinlan:
@jpmedley I was thinking about this recently, I think we can certainly do with a lot more depth and code samples when it comes to a lot of this.  I will leave a whole heap of comments in (including in some of the files that have moved location).  I am also open to landing this and then opening a bigger issue.

PR #2419, opened 2016-01-12
Fixes to image sizes, etc.

samdutton:


PaulKinlan:
Can you also make the image resolution physically smaller and ensure they are imageOptim'ed

samdutton:
Will do.

PR #2468, opened 2016-01-27
Update DevTools elements docs to cover gray entries.

### LINKED ISSUE #2444: document computed properties in devtools
http://stackoverflow.com/questions/34712218/what-does-it-mean-when-chrome-dev-tools-shows-a-computed-property-greyed-out

Garbee:
Addresses a common question about the gray items in the styles panel.

Via: https://stackoverflow.com/questions/34712218/what-does-it-mean-when-chrome-dev-tools-shows-a-computed-property-greyed-out

Fixes #2444


petele:
@pbakaus / @kaycebasques - can you please review and merge if it looks good to you.

pbakaus:
Nice job, thanks @Garbee!

PR #2482, opened 2016-02-01
Added WF Update: Record audio and video with MediaRecorder

samdutton:
@jpmedley PTAL

@miguelao, @niklasenbom,  FYI

@PaulKinlan I've left in the link to the screen recording Chrome App, but can remove if necessary.


PaulKinlan:
@samdutton merging, but please remove anything that is not open web. i.e, Chrome App.  We shouldn't be pointing to things that will never work in Firefox.

samdutton:
Fair point — removed and did another PR.

PR #2484, opened 2016-02-01
Added a step to enable web inspector in iOS for Safari

jitendravyas:
From the steps to debug iOS Safari from Safari Desktop, one important step was not mentioned, and I found via googling that we first need to enable "Web Inspector" to make this work. I have added this step I think it might save other's time too.


jitendravyas:
I signed it!

PR #2488, opened 2016-02-01
Adds line highlighting support to {% highlight blocks %}

petele:
In a `{% highlight %}` block, this will make certain lines bold so that they are more apparent.

Use: `{% highlight javascript hl_lines="3 4 5 6" %}`

### DIFF
--- a/src/jekyll/_includes/page-structure/close-page.liquid
+++ b/src/jekyll/_includes/page-structure/close-page.liquid
@@ -54,4 +54,4 @@
 
 </main>
 </div>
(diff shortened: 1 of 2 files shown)


jpmedley:
@petele 

Is there a way to implement this relies on wrapping the actual lines instead of naming them by number? Granting that we're not a wiki, MDN uses a naming approach and it has a horrible tendency to wrong. Food for thought.

petele:
No, sadly not. The feature is provided by Pygments, all I've done is change the highlight from nothing to bold. :(

PR #2508, opened 2016-02-03
update console-ui

kaycebasques:
@pbakaus plz review

note: custom object formatters has been documented (added a row to the table at the bottom with link to doc).


Garbee:
It would be nice if in the [filter section](https://github.com/kaycebasques/WebFundamentals/blob/console-ui/src/content/en/tools/chrome-devtools/debug/console/console-ui.markdown#filtering-the-console-output) we could get a note added. It should inform users if they do "Ctrl + click" on multiple filter types, then it will select multiple filters to show. This is [asked often enough](https://stackoverflow.com/questions/32740174/how-to-filter-out-logs-using-filter-with-two-categories/32741217#32741217) that we should float the ability in docs.

pbakaus:
Mostly LGTM, but some of the images feature too much whitespace / irrelevant info (i.e. the console images are too tall, no need to waste that much space with white). 

Also, they need better compression. For example, [this image](https://github.com/kaycebasques/WebFundamentals/blob/console-ui/src/content/en/tools/chrome-devtools/debug/console/images/console-panel.png) can easily be shrunken to 16.7k by using the free software ImageAlpha (png8+alpha, 256 colors). Happy to show you the tool in person if you like.

Garbee:
The deployment toolchain compresses images automatically.Should we compare its output vs manually doing it to see if any extra shrinkage happens?

pbakaus:
ah, wasn't aware. Prob not needed then, although in my experience, doing it manually always results in better compression with ImageAlpha, as your eye is still the best indicator on how low on colors you can go.

PR #2542, opened 2016-02-09
Fix incorrect attribution

eligrey:
I wrote this post, and my license allows you to host it here freely, but only given that the attribution is correct.


eligrey:
@petele Will <https://developers.google.com/web/updates/2011/08/Saving-generated-files-on-the-client-side> automatically regenerate soon or will that require manual action? Thanks.

petele:
It requires a manual push which I'm just about to do. It should be live within a few hours.

Issue #2540, opened 2016-02-09
Fix the image source 


osvaldoabel:
[In this article] (https://github.com/google/WebFundamentals/blob/master/src/content/pt-br/fundamentals/getting-started/your-first-multi-screen-site/index.markdown) the image's source is not working.

PR #2551, opened 2016-02-10
Fix image Source

Kanguanda:
Address of the issue  https://github.com/google/WebFundamentals/issues/2540

The name I used to sign the CLA is Osvaldo Abel


petele:
I don't think this solution will work on our production site, the file paths aren't quite the same.

Kanguanda:
i understood , @petele . i didn't know it. i was just readying and i found that error. 
if i

PR #2560, opened 2016-02-12
Fix a broken link on the webfont optimization page

ricbit:
Former link points to a 404, linking to project's main page instead.


ricbit:
I just signed it!

PR #2567, opened 2016-02-16
Fix duplicate shortcut.

Garbee:
Fixes [crbug 486222](https://code.google.com/p/chromium/issues/detail?id=486222). Shows both possible shortcuts to find by a filename.

cc @kaycebasques or @pbakaus for review.


Garbee:
Thank you!

PR #2572, opened 2016-02-18
Explains no-store in the correct way.

fabrik42:
Before this, the german text explained the exact opposite of what the `no-store` attribute does.

Before: "The browser and all intermediate caches can simply cache a copy of any version of the response."

After: "The browser and all intermediate caches are simply not allowed to cache a copy of any version of the response."


fabrik42:
I signed it!

petele:
@pbakaus - can you take a super quick look at this and verify it looks good? If it's good, please LGTM

pbakaus:
LGTM

PR #2576, opened 2016-02-21
Fixed typos and clarified some explanation

phistuck:
- Fixed typos.
- Called out that cssFloat is still around.
- Changed the text that implied that the error handler did not exist at some point, since every shipped version of RTCPeerConnection, from the beginning, had both of the handlers as parameters.
- Removed a meaningless and partial quote.


PaulKinlan:
Great. Thanks.  This should be going live in our Monday push

PR #2580, opened 2016-02-22
fix: checkout created new local branch

teachtyler:
`git checkout -b code-lab` created a new branch from master for me locally, after deleting the folder and re-cloning I excluded `-b` and it switched fine. This might be specific to me.

### DIFF
--- a/src/content/en/fundamentals/getting-started/your-first-offline-web-app/step-4.markdown
+++ b/src/content/en/fundamentals/getting-started/your-first-offline-web-app/step-4.markdown
@@ -14,7 +14,7 @@ authors:
 Go back to the command line and switch from `master` to the `code-lab` branch:
 
 {% highlight bash %}
-git checkout -b code-lab
+git checkout code-lab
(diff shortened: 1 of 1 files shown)


teachtyler:
I signed it!

jpmedley:
@teachtyler 

Before I merge I want to understand your issue. The instruction is to switch to a branch called code-lab, which hasn't been previously created. the -b switch is what creates the code-lab branch. 

My question is, what made you think the result was incorrect?

(I think I know what the problem is already. Please humor me.)

Thanks

teachtyler:
>which hasn't been previously created

https://github.com/GoogleChrome/airhorn/tree/code-lab
Switching to this is better than creating a new one, right?

>I think I know what the problem is already

https://xkcd.com/1597/

Garbee:
An alternate solution might be `-B` instead of outright removing it. This flag will use an existing branch (and reset it) if found, otherwise it will create the branch.

PR #2615, opened 2016-03-02
March 2016 update for increasing-engagement-with-app-install-banners-in-chrome-for-android

dandv:
- removed the Cr-UI-Browser-AppShortcuts paragraph because the link no longer points to any issues
- added having a `start_url` in the manifest as a requirement for the app install banner

The other changes are pretty straight-forward.


dandv:
Is this post useful now that we have https://developers.google.com/web/fundamentals/engage-and-retain/app-install-banners/ ?

dandv:
Related: https://github.com/google/WebFundamentals/pull/2617

dandv:
Thanks for the merge. Is this post useful now that we have https://developers.google.com/web/fundamentals/engage-and-retain/app-install-banners/ ? Should it link to the latter?

petele:
Yep, at some point we should add a banner to the top of the updates doc and point to the fundamentals doc.

Essentially the workflow should be new feature in updates, then when it becomes stable and is in other browsers, it goes to fundamentals.

Issue #2628, opened 2016-03-06
Fix typo in Udacity upsell


tachk:
E.g. Udacity course upsell shown on this page: https://developers.google.com/web/fundamentals/performance/critical-rendering-path/

Expected text:
"Check out our companion course [...]"

Actual text:
"Check out or companion course [...]"

PR #2629, opened 2016-03-06
Fixing typo in Udacity course upsell.

tachk:
Fix for #2628 

Following the precedent set by other PRs for typo fixes, this change has been scoped just to the base EN strings (though this change can be updated to fix this string in other languages' files, also).


PaulKinlan:
thanks!

PR #2647, opened 2016-03-08
update timeline-tool

kaycebasques:
please review @Garbee or @pbakaus 

This is just a baseline update so that the doc isn't horrendously out of date. There is probably a lot more content to cover. @pbakaus, @paulirish, and I are planning to meet to discuss how to build out the profiling docs. 

cc @paulirish


pbakaus:
LGTM.

PR #2649, opened 2016-03-09
Content layout/typo optimizations

pbakaus:
These are small, but powerful optimizations that improve the look and feel across Web Fundamentals. Explanation:

 - no margin on li's: They usually don't have them, and this has been blowing up list-heavy pages like crazy.
 - no letter spacing: Removed letter spacing from three places where they've been used. Weirdly, they come from MDL, yet in the Material Design spec, I couldn't find any note about letter spacing. Letter spacing is arguably a bad idea, as it messes with the font and decreases readability, and should be reserved for big, bold headings.
- The indentation of the second level in the nav drawer was redundant, as the title above from the first level nav is already bold, suggesting it's the title. W3C Specs have been doing something similar, [looking quite nice](https://drafts.css-houdini.org/worklets/).  This gives us a few much needed horizontal pixels back.

**Before:**
before

**After**
after

@petele and @gauntface, would be awesome if you could take a look.

### DIFF
--- a/src/static/styles/partials/_base.scss
+++ b/src/static/styles/partials/_base.scss
@@ -75,7 +75,6 @@ img, video {
 
 li {
   line-height: $base-line;
-  margin-bottom: $base-line / 2;
 }
 
 pre {
@@ -203,8 +202,11 @@ blockquote {
 .intro {
   @include typo-subhead-2($usePreferred: false);
   font-weight: normal;
+  letter-spacing: 0;
   margin-bottom: $base-line;
   margin-top: $base-line * 2;
+  padding-bottom: $base-line / 2;
+  border-bottom: 1px solid #eee;
 }
 
 .mdl-cell > .intro {
--- a/src/static/styles/partials/components/_wf-border-container.scss
+++ b/src/static/styles/partials/components/_wf-border-container.scss
@@ -20,3 +20,8 @@
   border-bottom-width: 1px;
   border-bottom-color: $grey-border-color;
 }
+
+p.intro + .wf-border-container .wf-border-container__content {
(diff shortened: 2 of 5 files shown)


pbakaus:
cc @paulirish (we've discussed some of these this morning)

paulirish:
+1 these

the letter spacing in particular really threw me for a loop.

petele:
In general I like most of the changes here, but it looks like there were a few side effects that need to be accounted for...

* Double line when there's no content in some pages [example](https://pr-2649-dot-weasel-dot-web-central.appspot.com/web/fundamentals/design-and-ui/browser-customization/)
* Landing page lists are a bit too tight [example](https://pr-2649-dot-weasel-dot-web-central.appspot.com/web/fundamentals/design-and-ui/?hl=en)
* Other lists are too tight [example](https://pr-2649-dot-weasel-dot-web-central.appspot.com/web/fundamentals/discovery-and-monetization/search-optimization/url-structure?hl=en)

Before we merge I'd like to see these kinds of issues address. Or split this into two PRs, one for letter spacing and one for li's.

pbakaus:
Thanks for finding these Pete! I fixed 1 and 2, but 3 is hard as there is no "right" solution for all scenarios: Without margin, the list items look a little tight when they have multiple lines of text, but with margin, the look bad in most other scenarios.

I've checked out some other pages like Smashing Mag and List Apart, and they all have margins but 1/3 less than what we had. I'll play around with it and update this PR so you can take a look.

pbakaus:
Oh just saw you merged, ok will do another PR and address 3) in that one ;)

PR #2650, opened 2016-03-09
update edit-styles

kaycebasques:
@pbakaus or @Garbee please review

Added a section on the **.cls** button in the **Styles** pane and did a general update of the doc. I didn't look at the preprocessor section. Me thinks that should be moved to its own doc.

cc @paulirish


Garbee:
LGTM

PR #2652, opened 2016-03-09
document PR staging sites

kaycebasques:
@petele @crhym3 WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,36 @@ These can be combined like so:
 
 We have a resource area which you can access at [https://web-central.appspot.com/web/resources/?hl=en](https://web-central.appspot.com/web/resources/?hl=en) that should get you up and running with a lot of the custom and useful info.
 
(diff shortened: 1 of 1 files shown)


crhym3:
Thanks for doing this!

One suggestion: maybe instead of pointing to a fixed URL like `pr-<NUMBER>-dot-weasel-dot-web-central.appspot.com`, provide a screenshot or tell people the link will be available under "Show all checks"?

screen shot 2016-03-09 at 4 50 27 pm

Although the URL pattern is quite stable, it's not immutable and may change in the future. I guess the readme could be updated as well in that case. Just wondering what approach to document this is better.

Another note: staged versions are also available for any branch in this repo at `<branch>-dot-weasel-dot-web-central.appspot.com`.

kaycebasques:
Hi @crhym3, updated per your feedback.

crhym3:
LGTM :star2:

PR #2665, opened 2016-03-11
update color picker docs

kaycebasques:
PTAL @pbakaus


pbakaus:
Nice, what you could include on this page are the little buttons that appear when you hover over the little plus on a style rule, that will then allow you to quickly add a background or foreground color.

Anyway, LGTM.

PR #2711, opened 2016-03-16
Use ... in window for feature detection

jeffposnick:
R: @petele @PaulKinlan @gauntface 

As per @webreflection's [recommendation](https://plus.google.com/108461295186427019068/posts/g71kJbSqxA7).

@timvolodine, you're using `window.hasOwnProperty()` in https://github.com/timvolodine/deviceorientation-test/blob/master/index.html#L218, as a heads-up.


timvolodine:
thanks Jeff for the heads-up, also fixed this in https://timvolodine.github.io/deviceorientation-test/ now

Issue #2716, opened 2016-03-17
create a settings directory for devtools docs


kaycebasques:
Problems:
* The **settings** link on /web/tools/chrome-devtools links out to DCC.
* The keyboard shortcuts doc is awkwardly organized under the iterate docs.
* There's no natural place to document general appearance features, like being able to reorder the panel tabs.
Solution:
* Create a directory /web/tools/chrome-devtools/settings and house the stuff there. We could name it "configure" so that it's consistent with the rest of action-oriented stuff, but I think "settings" is more straightforward. Or we can just store the docs directly under chrome-devtools. E.g. chrome-devtools/shortcuts, chrome-devtools/settings.

Thoughts @pbakaus?

cc @paulirish

pbakaus:
We've left these things on the old page for some reason, but I think it makes sense to migrate them all over. That would only leave the "Let's work on DevTools themselves" docs on the old site, which is fine for now. Thoughts, @paulirish and @Meggin?

paulirish:
URL wise i don't care very much... but "just store the docs directly under chrome-devtools. E.g. chrome-devtools/shortcuts, chrome-devtools/settings."  is my favorite.

Left-nav wise I'd be fine if they never showed up, but if they only showed at top-level when open, that'd be cool too.

But regardless they should be linked from the homepage.

kaycebasques:
That was my preferred approach too. I'm going with that

PR #2763, opened 2016-03-29
add devtools settings doc

kaycebasques:
PTAL @pbakaus 

See #2716

There's potentially a lot of stuff that we could cover in here. Right now the goal is to just have a home for general settings stuff that doesn't fit into iterate / debug / profile.  We'll add content ad-hoc as needed.


src/content/en/tools/chrome-devtools/settings.markdown
```diff
@@ -0,0 +1,114 @@
...
+the **Undock into separate window** 
+(![undock button](images/undock.png){:.inline})
+button, **Dock to bottom** 
+(![dock to bottom button](images/dock-bottom.png){:.inline})
+button, or 
+**Dock to right** 
+(![dock to right button](images/dock-right.png){:.inline})
+button. 
+
+## Set dark theme {#dark-theme}
```
pbakaus: "set" is a weird word here. I propose "Use" or "Switch to"

src/content/en/tools/chrome-devtools/settings.markdown
```diff
@@ -0,0 +1,114 @@
...
+
+To use a black and orange DevTools theme, [open DevTools' Settings](#settings),
```
pbakaus: Nothing really orange about it. Orange are just a few of the inverted colors. I'd prefer not to mention orange, as it might change.

src/content/en/tools/chrome-devtools/settings.markdown
```diff
@@ -0,0 +1,114 @@
...
+To enable Experiments, go to `chrome://flags/#enable-devtools-experiments`
+and click **Enable**. Click the **Relaunch Now** button at the bottom of the
+page. 
+
+You should now see a new page called **Experiments** when you open DevTools
+Settings.
+
+![DevTools Experiments](images/experiments.png)
+
+While DevTools is in focus, press <kbd>Shift</kbd> six times to view extremely
```
pbakaus: Don't mention this in the docs. It's supposed to be hidden by design :)

pbakaus:
Cool, good start! A few things that should be addressed before merging, but nothing very major. It might make sense in the future to have two docs: One for actual settings, and one for configuration/appearance.

PR #2765, opened 2016-03-29
Update article to match modifications made to source code

rwaldron:
Probably pending these: 

- https://github.com/01org/webbluetooth-edison-demo/pull/1
- https://github.com/01org/webbluetooth-edison-demo/pull/2

### DIFF
--- a/src/content/en/updates/posts/2016/03/web-enabled-internet-of-things.markdown
+++ b/src/content/en/updates/posts/2016/03/web-enabled-internet-of-things.markdown
@@ -60,7 +60,7 @@ of PWAs. PWAs are site that have a responsive app like user
 experience, can work offline and can be added to the device home screen.
 
 As a proof of concept, I have been building a small device using
-the Intel&reg; Edison Arduino breakout board. Teh device contains
+the Intel&reg; Edison Arduino breakout board. The device contains
 a temperature sensor (TMP36) as well as an actuator (colored LED
 cathode). The schematics for this device can be found at the end
 of this article.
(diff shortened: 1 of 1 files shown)


rwaldron:
So... I definitely read the article and wrote these patches right after waking up from a nap, which I took after a long day at the American Museum of Natural History—in the future I'll avoid doing that ;)

Here's the dumb thing that I did... 

I _did_ try _some_ code before submitting the PR—it's a practice that I'm pretty diligent about, since I don't ever want to claim that something works if I haven't actually seen it doing so. _However_ the code I tried was correct, but the code I wrote here is incorrect. What is the difference? I didn't copy/paste, I just made a few minor edits and in the process, failed to write the correct thing here: 

```js
temperatureCharacteristic.valueChange(this.value);
```

Which should've been: 

```js
temperatureCharacteristic.valueChange(this.celsius);
```

Here's the code I used when I was reality checking, before making the mistake in updating your example: 

```js
var five = require("johnny-five");
var Edison = require("edison-io");

var board = new five.Board({
  io: new Edison()
});

board.on("ready", function() {
  var temperature = new five.Thermometer({
    controller: "TMP36",
    pin: "A0"
  });

  temperature.on("data", function() {
    console.log(this.celsius + "°C", this.fahrenheit + "°F");
  });
});
```

Note that this is nearly identical to the example shown in the Johnny-Five docs—because I just copied that as a starting point. 

Ok, longest explanation ever? Updates to follow :)

kenchris:
Actually what was returning undefined for me what the this.celcius

But if you tested this on the Edison then I am fine with submitting this :-) I would just like to know what I did wrong.

PaulKinlan:
@kenchris are we good to pull?

PaulKinlan:
As a total aside, here is the preview: https://pr-2765-dot-weasel-dot-web-central.appspot.com/web/updates/2016/03/web-enabled-internet-of-things?hl=en

rwaldron:
>  this.celcius

Did you try `this.celsius`? ;) <3

kenchris:
If that is the error, I am embarrassed :)

kenchris:
@PaulKinlan yes sure! I have some other patch myself which I want to push soon.

PR #2768, opened 2016-03-31
Updated Implementation Bugs URL

beaufortfrancois:
R=@petele


PaulKinlan:
We shouldn't be pointing to an http url, why was https removed?

beaufortfrancois:
My bad! This is fixed.

PR #2807, opened 2016-04-12
Update step-07.markdown

emmynet:
Wrong curl request params

### DIFF
--- a/src/content/en/fundamentals/getting-started/push-notifications/step-07.markdown
+++ b/src/content/en/fundamentals/getting-started/push-notifications/step-07.markdown
@@ -63,7 +63,7 @@ From your terminal, run the cURL command below — but make sure to use your
 own API key and subscription ID, which you created earlier:
 
 {% highlight bash %}
-curl --header "Authorization: key=XXXXXXXXXXXX" --header "Content-Type: application/json" https://android.googleapis.com/gcm/send -d "{\"registration_ids\":[\"fs...Tw:APA...SzXha\"]}"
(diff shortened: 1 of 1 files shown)


PaulKinlan:
Hey, thanks for the PR. I know it is only a small commit, but we really need you to sign the CLA

emmynet:
. I signed it!
Il 15/Apr/2016 19:55, "Paul Kinlan" <notifications@github.com> ha scritto:

PaulKinlan:
Now, iirc we've changed this a number of times in the past.

Every example that I have seen of this has the " around all of the Content-Type header, and not after the ":"..... which is correct and why?

emmynet:
I've tested curl on MacOsx and it doesn't work.

After many tests I have succeeded by doing the same thing that you have
done in your Demo:

https://simple-push-demo.appspot.com/

I thing that the problem is curl's version on MacOsx because now i tried on
linux (curl version 7.47.0) and it works in both ways!


Il 15/Apr/2016 21:51, "Paul Kinlan" <notifications@github.com> ha scritto:

Now, iirc we've changed this a number of times in the past.

Every example that I have seen of this has the " around all of the
Content-Type header, and not after the ":"..... which is correct and why?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/google/WebFundamentals/pull/2807#issuecomment-210616204>

petele:
It looks like the [proper syntax](http://stackoverflow.com/questions/356705/how-to-send-a-header-using-a-http-request-through-a-curl-call) for curl is to quote the whole header, not just the value.

I'm going to close for now, if you can find something that indicates otherwise, please re-open.

PR #2831, opened 2016-04-19
Credential Management API update #2830

PaulKinlan:


jpmedley:
Line 43 needs to be addressed before we can publish this.

PR #2845, opened 2016-04-21
Credential management api article fix

agektmr:


agektmr:
This one seems to include some unnecessary changes. Closing.

PR #2886, opened 2016-05-03
Fixes #2775 - adding a note

### LINKED ISSUE #2775: TLS version fallback deprecation post should include instructions for administrators
In https://developers.google.com/web/updates/2016/03/chrome-50-deprecations#remove-insecure-tls-version-fallback, the TLS version fallback section ideally would include a sentence or two for administrators. Something generic like: "Administrators should ensure their server software is up-to-date." should do. In the Intent email I also added "If this is still unresolved, reach out to the software vendor". I don't know how useful that is.

(This is caused by bugs in, usually, old server software. I'm not aware of any concrete bugs remaining, apart from one Windows issue fixed years ago. But there might be some I'm not aware of.)

PaulKinlan:


PaulKinlan:
@petele we might need a deploy to get this updated.  That page has a lot of traffic.

petele:
Pushing now.

PR #2932, opened 2016-05-23
Corrected a misplaced underscore

YarmoM:
The misplaced underscore was causing a wrong placement of em tags in the resulting HTML page.


PaulKinlan:
Thanks

PR #3003, opened 2016-06-15
Moved files to 06 directories, updated M51 to M52, tweaked text

samdutton:


samdutton:
@PaulKinlan @petele  Didn't realise this had already been published :). I can change back the publication date and image/post directories if you'd prefer.

PaulKinlan:
Let's just go with this.

PR #3010, opened 2016-06-19
Fixed a small typo

tarungarg546:
Fixed a small typo changed an `and` to `an`.

Result :-

- Before :- 
>It may be cool to know that the browser can execute one version of a thing 100 times faster               than another thing, like that requesting `and` element’s offsetTop is faster than computing getBoundingClientRect().

- After :- 
>It may be cool to know that the browser can execute one version of a thing 100 times faster than another thing, like that requesting `an` element’s offsetTop is faster than computing getBoundingClientRect()


tarungarg546:
@googlebot I signed it!

PR #3048, opened 2016-07-03
Remove gendered pronoun

razzius:
Not sure if Google has a standard for gender neutral pronouns, but this is what I'd use.


PaulKinlan:
Thanks.  This should be reflected on the site in the next push.

PR #3053, opened 2016-07-05
Fix link in infinite-scroller article

surma:
@petele @jeffposnick Just a quick link fix.


surma:
thanks @oldergod for letting me know :)

PR #3080, opened 2016-07-07
Long nav items wrap

ebidel:
R: @petele 

screen shot 2016-07-07 at 2 55 34 pm

### DIFF
--- a/src/static/styles/partials/components/_wf-nav-drawer.scss
+++ b/src/static/styles/partials/components/_wf-nav-drawer.scss
@@ -54,6 +54,9 @@
 
 .wf-main-nav__item {
   margin-bottom: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


petele:
Hey @ebidel, I'm hesitant to merge this one because I'm afraid it'll hide too much of the navigation. I'm worried that you won't necessarily be able to find what you're looking for if it's cut off. 

Thoughts?

ebidel:
I was worried about that too. For longer items it really helps but for
shorter ones it could be harder to find things. Your call.

jpmedley:
Option 3: Edit the titles to make them shorter where possible and where it
isn't, follow our push message advice and "Put important information up
front". After the later, Eric's suggestion would be more reasonable.

 816-678-7195

ebidel:
Proposed a `shortname` metadata field: https://github.com/google/WebFundamentals/issues/3101

PR #3089, opened 2016-07-11
Translated optimize-encoding-and-transfer EN to KO

joshua1988:
My first translation work

PR #3088, opened 2016-07-11
Update _contributors.yaml

joshua1988:
My personal profile added

PR #3087, opened 2016-07-11
Patch 1

joshua1988:
My First Translation Work as Google Contributor.
Need to figure out how to make it neat in term of the includes, and takeaways.

PR #3090, opened 2016-07-11
Revert 1 joshua1988 patch 1

joshua1988:


petele:
@joshua1988 - looks like this isn't building, can you please take a look. And can you also add a description about what this does?

petele:
Hey @joshua1988, I'm going to close this as it looks like it duplicates #3088, #3089 and #3087 .

joshua1988:
@petele - I think there is something wrong with the contents I translated. To bring you up to speed, I copied the original markdown from "en" folder and translated it in Korean. Do you think it caused the problem with this building?

Another thing is that,
the code `{% include shared/remember.liquid list=page.notes.gzip %}` inside the document doesn't seem to work. Please let me know what I can do for that.

Thanks!

PR #3110, opened 2016-07-17
Create index.markdown

moinshaikh:
Translation of index.markdown file into Hindi (Indian-Hindi) language.

PR #3111, opened 2016-07-17
Update _langnames.yaml

moinshaikh:
Added 'Hindi' language.

### DIFF
--- a/src/content/_langnames.yaml
+++ b/src/content/_langnames.yaml
@@ -50,3 +50,6 @@ tr:
 
 'zh-tw':
   name: '中文（繁體）'
+
+hi:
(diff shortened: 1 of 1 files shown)


src/content/_langnames.yaml
```diff
@@ -50,3 +50,6 @@ tr:
 
 'zh-tw':
   name: '中文（繁體）'
+
```
petele: Would you please ensure that these are added to the yaml alphabetically?

petele:
See comments in #3110

src/content/_langnames.yaml
moinshaikh: @petele I have added this alphabetically now.

Thanks,
Moin

Issue #2954, opened 2016-06-03
[EN] Feedback for: /web/updates/2015/07/interact-with-ble-devices-on-the-web?hl=en


jyasskin:
Linking directly to https://developers.google.com/web/updates/2015/07/interact-with-ble-devices-on-the-web?hl=en#user-gesture-required doesn't scroll to the right place on the page.

Garbee:
Duplicate of #2253

PaulKinlan:
@sgomes @petele please sort asap, this is getting beyond a joke now.  I know we are planning to migrate back into devsite soon but right now we need to get this fixed, people can't use our site.

sgomes:
This is not an MDL issue, as it only affects Web Fundamentals. WF seems to be modifying page and MDL layout styles, and is triggering a Safari bug. Workaround on #3118.

jyasskin:
I'm using Chrome.

sgomes:
Bad dupe, sorry. The follow up to this is on #2291.

Issue #2291, opened 2015-11-13
MDL is breaking #urls


PaulKinlan:
If we send a user to an anchor in the page it scrolls to the top.  We really need to fix this as every user who is landing from devtools is not seeing the content they needed.

gauntface:
I believe this is a MDL 1.0.6 bug - but not sure if our production build is using that yet or not.

Garbee:
It is an MDL 1.0 bug period, however it is due to how we handle layout (along with quite a few other problems.) V2 is going to fix it since we will be doing less stuff for developers, thereby avoiding the layout hacks we currently employ.

kaycebasques:
Duplicate of #2253

sgomes:
Can we have an example URL to reproduce the bug?

Garbee:
https://developers.google.com/web/tools/chrome-devtools/profile/network-performance/understanding-resource-timing#diagnosing-network-issues

petele:
Yep - I see it happen, page loads at the anchor, then some event fires and the page jumps to the top. If you click on the anchor again, it'll scroll down to that part.

petele:
This should be fixed now. Sorry about that!!

PR #3118, opened 2016-07-19
Update _mdl-overrides.scss

sgomes:
Workaround for Safari scrolling bug.

@petele @PaulKinlan Please test.

### DIFF
--- a/src/static/styles/partials/_mdl-overrides.scss
+++ b/src/static/styles/partials/_mdl-overrides.scss
@@ -11,6 +11,8 @@
 // From Git Issue https://github.com/google/WebFundamentals/issues/2022
 .mdl-layout__content {
    will-change: transform;
+   // From Git issue https://github.com/google/WebFundamentals/issues/2949
+   transform: translateZ(0);
 }
 
 @media screen and (max-width: $grid-desktop-breakpoint) {


gauntface:
I can't see any reason not to go with this. LGTM

petele:
Merged and pushed live.

ricardobeat:
@petele the issue is still there. Is there a delay before it reaches live servers?

petele:
This only fixed the scroll prevention, not the fact it jumps. We're working on moving to a new infrastructure that'll eliminate that issue completely.

ricardobeat:
@petele not sure what you mean - scrolling in Safari is still not possible at all: https://cdn.pbrd.co/images/e55f14PTf.gif

petele:
@sgomes - can you take another look please?

sgomes:
@petele The fix doesn't seem to be live, as far as I can tell.

petele:
Just did another push, and I see the transform in the CSS and with Safari 9.1.1 on my Mac, I'm able to scroll just fine. I *think* that did it.

jyasskin:
https://developers.google.com/web/updates/2015/07/interact-with-ble-devices-on-the-web?hl=en#user-gesture-required still doesn't work in Chrome, even though #2954 was duped to this PR.

sgomes:
That's a bad dupe, sorry about that @jyasskin. The follow up to that one is #2291

PR #3181, opened 2016-08-02
Typo and style fixes

dandv:
Space between "function" and parentheses (anonymous functions) [per Crockford](http://stackoverflow.com/questions/9765942/space-after-function-name-is-wrong/11076228#11076228).


PaulKinlan:
iirc we are using the Google Style guide on this site and we have no spaces between function and parenthesis.

petele:
@PaulKinlan is correct, we use the [Google JavaScript Style Guide](https://google.github.io/styleguide/javascriptguide.xml), which inherits no spaces from the Google C++ Style Guide.  The way I read the [SO post](http://stackoverflow.com/questions/9765942/space-after-function-name-is-wrong/11076228#11076228), it looks like they're saying no spaces as well.

@dandv - Can you create a new PR with only the typo fixes please?

PR #3193, opened 2016-08-04
Speech synthesis only works while online

dandv:
There's also a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=634545) that if Chrome goes offline and back online, speech synthesis stops.


petele:
@dandv is that true on all platforms? i thought on Mac it could/would use the system speech synthesis stuff.

dandv:
You might be right:

    speechSynthesis.getVoices().forEach(voice => console.log(voice.localService));

shows "true" 64 times and "false" 19 times on Mac OS X, but only false 19x on Ubuntu.

PR #3202, opened 2016-08-08
Add adoptedCallback to CE article. Add custom event dispatch to SD article

ebidel:
R: @domenic @azakus @justinfagnani

cc @petele

### DIFF
--- a/src/content/en/fundamentals/primers/customelements/index.markdown
+++ b/src/content/en/fundamentals/primers/customelements/index.markdown
@@ -547,7 +547,7 @@ let promises = [...undefinedButtons].map(socialButton => {
   return customElements.whenDefined(socialButton.localName);
 ));
 
-// Wait for all the social-buttons to to be upgraded.
+// Wait for all the social-buttons to be upgraded.
 Promise.all(promises).then(() => {
   // All social-button children are ready.
 });
@@ -812,9 +812,9 @@ Defines a new custom element in the browser.
 Example
 
 {% highlight javascript %}
-customElements.define('my-app', class extends HTMlElement { ... });
+customElements.define('my-app', class extends HTMLElement { ... });
 customElements.define(
-  'fancy-button', class extends HTMlButtonElement { ... }, {extends: 'button'});
+  'fancy-button', class extends HTMLButtonElement { ... }, {extends: 'button'});
 {% endhighlight %}
 
 **`get(tagName)`**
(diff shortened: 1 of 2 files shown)


src/content/en/fundamentals/primers/customelements/index.markdown
```diff
@@ -369,6 +373,35 @@ class AppDrawer extends HTMLElement {
...
+Promise.all([createWindow(), createWindow()])
+  .then(([w1, w2]) => {
+    // 2. Define a custom element in w1.
+    w1.customElements.define('x-adopt', class extends w1.HTMLElement {
+      adoptedCallback() {
+        console.log('Adopted!');
+      }
+    });
+    let a = w1.document.createElement('x-adopt');
+    w2.document.adoptNode(a); // 3. Adopt the custom element into w2.
```
domenic: There are actually other much more common situations where adoption occurs. Just inserting an element defined in one document into the other will adopt it. So e.g. I'd do `w2.document.body.appendChild(a)` as a good illustration of how this might happen in practice.

src/content/en/fundamentals/primers/shadowdom/index.markdown
```diff
@@ -680,9 +678,15 @@ The events that **do** cross the shadow boundary are:
...
+**Tips**
+
 If the shadow tree is open, calling `event.composedPath()` will return an array of nodes that the event traveled through.
-{: .wf-talkinghead }
 
+If you want to fire custom events from a custom element that uses shadow DOM, be sure to set the `composed` flag when creating the event:
+
+    this.dispatchEvent(new Event('special-thing', {bubbles: true, composed: true});
+
+If `composed: false` (default), consumers won't be able to listen for the event outside of your shadow root.
```
domenic: Hmm, is this true? I thought it would just retarget it to your shadow host. Which is probably what you want to do in most cases, so suggesting that people always set `composed: true` for firing custom events inside a shadow DOM seems wrong.

domenic: I am probably wrong here, reading more. Interesting. Still, I think you should not need to set the composed flag unless you are firing the event *on something inside the shadow DOM*, which in most cases is not true---you are firing the event on shadow host custom element itself.

ebidel: This was all news to me as of yesterday :) cc @azakus @hayatoito

src/content/en/fundamentals/primers/customelements/index.markdown
ebidel: Done

src/content/en/fundamentals/primers/shadowdom/index.markdown
ebidel: updated to be more clear

domenic: Really nice update, thanks.

petele:
Eric - is this ready to go? I'm planning a push for about 8pm Wednesday PT.

ebidel:
Yea!

PR #3225, opened 2016-08-15
Fixes #3221 and fixes #3222

### LINKED ISSUE #3221: Confusing example in shadow dom article
https://developers.google.com/web/fundamentals/primers/shadowdom/?hl=en

> The spec defines a list of elements that can’t host a shadow tree…

```js
let shadowRoot = document.createElement('div').attachShadow({mode: 'open'});
// Error. `<input>` cannot host shadow dom.
// document.createElement('input').attachShadow({mode: 'open'});
```

It's not clear why the first line is uncommented & stores a variable, while the third line is commented and doesn't.

Since you've already provided a working example, should this only show a failing example? Eg:

```js
// This doesn't work:
let shadowRoot = document.createElement('input').attachShadow({mode: 'open'});
// Error. `<input>` cannot host shadow dom.
```

+@ebidel

### LINKED ISSUE #3222: Inconsistent let/const usage in shadow dom article
https://developers.google.com/web/fundamentals/primers/shadowdom/?hl=en

Sometimes `const` is used, sometimes `let` is used. All uses of `let` appear like they could be `const`, with the exception of:

```js
for (let i = 0, el; el = nodes[i]; ++i) {
```

…which could be (IMO) better written as:

```js
for (const el of nodes) {
```

+@ebidel

ebidel:
R: @jakearchibald @petele

### DIFF
--- a/src/content/en/fundamentals/primers/shadowdom/index.markdown
+++ b/src/content/en/fundamentals/primers/shadowdom/index.markdown
@@ -142,12 +142,11 @@ The spec [defines a list of elements](http://w3c.github.io/webcomponents/spec/sh
 - The browser already hosts its own internal shadow DOM for the element (`<textarea>`, `<input>`).
 - It doesn't make sense for the element to host a shadow DOM (`<img>`).
 
-For example:
+For example, this doesn't work:
 
 {% highlight javascript %}
-let shadowRoot = document.createElement('div').attachShadow({mode: 'open'});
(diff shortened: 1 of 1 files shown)


jakearchibald:
Looks good to me. I see you're keeping your old school for-loop, is that just personal preference? (@aerotwist is curmudgeonly when it comes to for-of too)

ebidel:
If it ain't broke...

PR #3255, opened 2016-08-22
Add focus section to shadow dom article

ebidel:
R: @robdodson @hayatoito @petele


src/content/en/fundamentals/primers/shadowdom/index.markdown
```diff
@@ -689,21 +689,108 @@ If the shadow tree is open, calling `event.composedPath()` will return an array
...
+                  this.shadowRoot.activeElement);
+    });
+  }
+});
+</script>
+{% endhighlight %}
+
+**Result**
+
+<img src="images/delegateFocusTrue.png" title="delegatesFocus: true behavior">
```
robdodson: is the blue border supposed to be the default focus outline?

robdodson:
+@alice

src/content/en/fundamentals/primers/shadowdom/index.markdown
```diff
@@ -689,21 +689,108 @@ If the shadow tree is open, calling `event.composedPath()` will return an array
...
+
+{% highlight html %}
+<x-focus>
+  #shadow-root
+    <input type="text" placeholder="Input inside shadow dom">
+{% endhighlight %}
+
+ The `focus` event will look like it came from `<x-focus>`, not the `<input>`. Similarly, `document.activeElement` will be `<x-focus>`. If the shadow root was created with `mode:'open'` (see [closed mode](#closed)), you'll also be able access the internal node that gained focus:
+
+    document.activeElement.shadowRoot.activeElement // only works with open mode.
```
robdodson: what if the focused element is like 10 shadow roots deep? do I have to keep calling activeElement.shadowRoot.activeElement?

src/content/en/fundamentals/primers/shadowdom/index.markdown
```diff
@@ -689,21 +689,108 @@ If the shadow tree is open, calling `event.composedPath()` will return an array
...
+
+Another option is `delegatesFocus: true`, which expands the focus behavior of element's within a shadow tree:
+
+- If you click on a node inside shadow DOM and the node is not a focusable area, the first focusable area becomes focused.
```
robdodson: So if I click on a `p` that's next to an `input` it will focus the `input` and apply the default outline style?

src/content/en/fundamentals/primers/shadowdom/index.markdown
```diff
@@ -689,21 +689,108 @@ If the shadow tree is open, calling `event.composedPath()` will return an array
...
+- When a node inside shadow DOM gains focus, `:focus` applies to the host in addition to the focused element.
```
robdodson: If I haven't provided a :focus style, would I not see anything on the host and just the default outline on the actually focused element?

src/content/en/fundamentals/primers/shadowdom/index.markdown
```diff
@@ -689,21 +689,108 @@ If the shadow tree is open, calling `event.composedPath()` will return an array
...
+If you were to set `delegatesFocus: false`, here's what you would see instead:
+
+<figure>
+  <img src="images/delegateFocusFalse.png">
+  <figcaption>The internal <code>&lt;input></code> is focused.</figcaption>
+</figure>
+
+<figure>
+  <img src="images/delegateFocusFalseFocus.png">
+  <figcaption><code>&lt;x-focus></code> gains focus (e.g  <code>focus()</code>).</figcaption>
```
robdodson: Maybe say: "If delegateFocus is false and the user focuses the `input`, focus will be redirected to the containing element" or something along those lines

src/content/en/fundamentals/primers/shadowdom/index.markdown
```diff
@@ -689,21 +689,108 @@ If the shadow tree is open, calling `event.composedPath()` will return an array
...
+
+<img src="images/delegateFocusTrue.png" title="delegatesFocus: true behavior">
+
+Above is the result when `<x-focus>` is focused (user click, tabbed into, `focus()`, etc.), "Clickable Shadow DOM text" is clicked, or the internal `<input>` is focused (including `autofocus`).
+
+If you were to set `delegatesFocus: false`, here's what you would see instead:
+
+<figure>
+  <img src="images/delegateFocusFalse.png">
+  <figcaption>The internal <code>&lt;input></code> is focused.</figcaption>
```
robdodson: Maybe say something like "if delegateFocus is false and the internal input is focused, focus will be redirected to the parent"

src/content/en/fundamentals/primers/shadowdom/index.markdown
robdodson: How would x-focus gain focus without a tabindex?

src/content/en/fundamentals/primers/shadowdom/index.markdown
robdodson: argh, nevermind I am reading these images wrong. I thought the caption was applying to the image underneath it because they're physically closer

src/content/en/fundamentals/primers/shadowdom/index.markdown
hayatoito: Yes, we have to call it, recursively, until it reaches the end.

There is no *deep* version for activeElement.
The issue is filed here https://github.com/w3c/webcomponents/issues/104.

hayatoito:
R: @TakayoshiKochi for focus.

petele:
Looks like there are still some comments. @ebidel can you sort and verify everything looks as it should [https://pr-3255-dot-weasel-dot-web-central.appspot.com/](https://pr-3255-dot-weasel-dot-web-central.appspot.com/) then remove the DO NOT MERGE label and i'll merge and push.

src/content/en/fundamentals/primers/shadowdom/index.markdown
ebidel: Yes.

src/content/en/fundamentals/primers/shadowdom/index.markdown
ebidel: You'd see the default focus ring around the host and the element.

src/content/en/fundamentals/primers/shadowdom/index.markdown
ebidel: Yea, I want to illustrate the how the styles are affected inside SD and outside of it.

src/content/en/fundamentals/primers/shadowdom/index.markdown
ebidel: Done.

ebidel:
PTAL

src/content/en/fundamentals/primers/shadowdom/index.markdown
robdodson: That wasn't immediately clear to me. You might add a note mentioning that

src/content/en/fundamentals/primers/shadowdom/index.markdown
robdodson: Wow really? That's super weird. Does it actually apply focus or just move the focus navigation start point (https://developers.google.com/web/updates/2016/03/focus-start-point?hl=en)

If it were to focus the input that would be atypical of how the browser works today. Think about a page with no shadow dom. If you click somewhere on the page it suddenly focuses the nearest input. That would feel broken to me.

src/content/en/fundamentals/primers/shadowdom/index.markdown
robdodson: I'd see two focus rings??? That doesn't seem to make sense. The concept of focus, as I understand it, is pretty specific in that only 1 thing on your entire computer ever "has focus".

src/content/en/fundamentals/primers/shadowdom/index.markdown
ebidel: This actually felt right to me but who knows. It's also opt-in so maybe the use case is the "widget", where you have focusable items inside the shadow dom. When you tab to the widget, you should be focused on the first focusable item.

cc @TakayoshiKochi

src/content/en/fundamentals/primers/shadowdom/index.markdown
ebidel: That's according to spec.

robdodson: Can you link me? How would someone know where their keyboard events are being directed? That just seems bizarre...

ebidel: Pretty sure the input gets the events. The other part only applies to CSS:

> When a node inside shadow DOM gains focus, `:focus` applies to the host in addition to the focused element.

https://www.w3.org/TR/shadow-dom/

johnliz4ever:
Its awesome

src/content/en/fundamentals/primers/shadowdom/index.markdown
alice: I think the issue Rob's pointing out is that say you have an input inside of a custom element which is composed into another custom element... etc., with default styling for `:focus` at each level you're going to get a blue ring, and you could end up with a case where things are absolute positioned or something such that you have blue rings all over the page, creating a really confusing experience where you don't know which is the "real" keyboard target.

robdodson: Yeah

> If :focus pseudo-class applies to HOST, and HOST is in a shadow root of another shadow host HOST2 which also delegates focus, :focus pseudo-class applies to HOST2 as well.

That seems like you could create an N-level deep focused styled UI which would be really confusing.

ebidel: Yea, this is massively confusing: http://jsbin.com/kuqepubihi/edit?html,output

Try that in Canary.

alice: It looks like a guidance system for spacecraft docking in a really sad videogame.

ebidel: AFA the demos in this article are concerned, I think they're ok according to the spec and illustrate what the feature does. At the very least, they show what Blink has implemented. Whether that's good UX or not is another story :\

I'd like to merge this so @petele can make the push later today.  Shall we continue the focus conversation over on a spec bug? I filed https://github.com/w3c/webcomponents/issues/554

alice: Works for me.

src/content/en/fundamentals/primers/shadowdom/index.markdown
alice: This seems odd to me too. If shadow DOM is used purely as an encapsulation mechanism (i.e. it's not a self-contained interactive near-leaf-node widget like `<date>`, but something more like `<topeka-app>`), why would a user expect clicking a part of the page to move focus to something "near" it in some cases but not others? Also, in some cases it would move focus to something earlier in the page than the thing that was clicked, and in some cases it would be later, totally unpredictably.

PR #3288, opened 2016-08-29
Update README.md

diddierh:
Update start URL to http://localhost:7331/web/fundamentals which is the right


diddierh:
I signed it!

PR #3304, opened 2016-09-01
PWA Landing page: Point Service Workers to primer

addyosmani:
For the average developer, pointing to the primer is going to be a lot more illuminating than a spec page/repo imo :) cc @jpmedley


ebidel:
LGTM

jpmedley:
@gauntface @petele 

"Unable to validate Yaml checksum." 

This something different from the ignorable error we usually get.

PR #3368, opened 2016-09-10
Fix formatting

arthurevans:
Missing an empty line before the bulleted list—formats correctly on GitHub, but not on WF:

https://developers.google.com/web/tools/chrome-devtools/profile/evaluate-performance/timeline-tool?hl=en#overview-pane


addyosmani:
LGTM

PR #3369, opened 2016-09-10
090916 work

shanayres:


PaulKinlan:
Do you have a summary of what is actually in this PR?  It looks like it is pulling in from an incredibly old set of changes and merging them.  It might be best to re-base or re-checkout your branch and then commit your two changes that you authored.  Right now, we can't pull this in as is.

petele:
Hey Shannon - it looks like this is trying to merge your changes into MASTER not into next-version, can you resubmit the PR to next-version instead.  Thanks!

PR #3377, opened 2016-09-12
CacheQueryOptions post

jeffposnick:
R: @petele @PaulKinlan etc.

This is Markdown conversion of a post that was previously reviewed internally.


jpmedley:
Did a quick edit.

PR #3405, opened 2016-09-16
Corrected a few things in a few files.

shanayres:


jpmedley:
Your updates are merged. The next time you do this you'll get them.

git checkout next-version
git pull

shanayres:
I did those two commands as my last step last night. I think they worked.

I always leave it too late to do them before I leave work, and then I
forget to do them when I first get home. I always seem to do them fairly
late at night, when I finally remember!

jpmedley:
The best time to do them is right before you start in the morning, right before
git checkout -b todaysbranch next-version

PR #3429, opened 2016-09-20
Fix MediaRecorder example mimeType

gotik:
`video/webm, codecs=vp8` and `video/webm, codecs=vp9` don't work as mime types, they must be `video/webm;codecs=vp8` and `video/webm;codecs=vp9`


gotik:
I signed it!

PR #3510, opened 2016-09-26
Housing case study

petele:


src/content/en/showcase/2016/housing.md
```diff
@@ -0,0 +1,65 @@
...
+platforms in India. After building [Housing Go](https://housing.com/), a
+Progressive Web App (PWA), they saw a 38% increase in total conversions
+across browsers. The new PWA also delivered higher-value users, with
+visitors spending 10% longer per session and returning more often.
+
+## Challenges
+Housing.com knew their mobile users woudn’t tolerate slow load times, which
+significantly affect conversions. They measured the impact of users’ average
+page-load time and found that even a one-second improvement brought a
+significant boost in the conversion rate.
```
jpmedley: "They measured the impact of users’ average page-load time and found that even a one-second improvement brought a significant boost in the conversion rate."

Can we do pullquotes?

petele: Yes, but this is for master and a quick push. Will look at for next-version

PR #3531, opened 2016-09-28
Tweaks to touch docs

gauntface:
cc @RByers @patrickhlauke

Thoughts on these changes? Anything we should be adding in?

### DIFF
--- a/src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
+++ b/src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
@@ -227,15 +227,8 @@
 
           // Add the move and end listeners
           if (window.PointerEvent) {
-            // Pointer events are supported.
-            document.addEventListener('pointermove', this.handleGestureMove, true);
-            document.addEventListener('pointerup', this.handleGestureEnd, true);
+            evt.target.setPointerCapture(evt.pointerId);
           } else {
-            // Add Touch Listeners
-            document.addEventListener('touchmove', this.handleGestureMove, true);
-            document.addEventListener('touchend', this.handleGestureEnd, true);
-            document.addEventListener('touchcancel', this.handleGestureEnd, true);
-
             // Add Mouse Listeners
             document.addEventListener('mousemove', this.handleGestureMove, true);
             document.addEventListener('mouseup', this.handleGestureEnd, true);
@@ -252,6 +245,10 @@
         /* // [START handle-move] */
         this.handleGestureMove = function (evt) {
           evt.preventDefault();
+
+          if(!initialTouchPos) {
+            return;
+          }
 
           lastTouchPos = getGesturePointFromEvent(evt);
 
@@ -278,21 +275,16 @@
 
           // Remove Event Listeners
           if (window.PointerEvent) {
-            // Remove Pointer Event Listeners
-            document.removeEventListener('pointermove', this.handleGestureMove, true);
-            document.removeEventListener('pointerup', this.handleGestureEnd, true);
+            evt.target.releasePointerCapture(evt.pointerId);
           } else {
-            // Remove Touch Listeners
-            document.removeEventListener('touchmove', this.handleGestureMove, true);
-            document.removeEventListener('touchend', this.handleGestureEnd, true);
-            document.removeEventListener('touchcancel', this.handleGestureEnd, true);
-
             // Remove Mouse Listeners
             document.removeEventListener('mousemove', this.handleGestureMove, true);
             document.removeEventListener('mouseup', this.handleGestureEnd, true);
           }
 
           updateSwipeRestPosition();
+
+          initialTouchPos = null;
         }.bind(this);
         /* // [END handle-end-gesture] */
 
@@ -355,7 +347,6 @@
           var point = {};
 
           if(evt.targetTouches) {
-            console.log(evt.targetTouches[0]);
             point.x = evt.targetTouches[0].clientX;
             point.y = evt.targetTouches[0].clientY;
           } else {
@@ -391,9 +382,15 @@
         if (window.PointerEvent) {
           // Add Pointer Event Listener
           swipeFrontElement.addEventListener('pointerdown', this.handleGestureStart, true);
+          swipeFrontElement.addEventListener('pointermove', this.handleGestureMove, true);
+          swipeFrontElement.addEventListener('pointerup', this.handleGestureEnd, true);
+          swipeFrontElement.addEventListener('pointercancel', this.handleGestureEnd, true);
         } else {
           // Add Touch Listener
           swipeFrontElement.addEventListener('touchstart', this.handleGestureStart, true);
+          swipeFrontElement.addEventListener('touchmove', this.handleGestureMove, true);
+          swipeFrontElement.addEventListener('touchup', this.handleGestureEnd, true);
+          swipeFrontElement.addEventListener('touchcancel', this.handleGestureEnd, true);
 
           // Add Mouse Listener
           swipeFrontElement.addEventListener('mousedown', this.handleGestureStart, true);
--- a/src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
+++ b/src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
@@ -65,9 +65,9 @@
 
         z-index: 2;
 
+        -ms-touch-action: none;
         /* // [START touch-action-example] */
         /* Pass all touches to javascript */
-        -ms-touch-action: none;
         touch-action: none;
         /* // [END touch-action-example] */
       }
@@ -172,14 +172,14 @@
       })();
 
       /* // [START pointereventsupport] */
-      var pointerDownName = 'MSPointerDown';
-      var pointerUpName = 'MSPointerUp';
-      var pointerMoveName = 'MSPointerMove';
+      var pointerDownName = 'pointerdown';
+      var pointerUpName = 'pointerup';
+      var pointerMoveName = 'pointermove';
 
-      if(window.PointerEvent) {
-        pointerDownName = 'pointerdown';
-        pointerUpName = 'pointerup';
-        pointerMoveName = 'pointermove';
+      if(window.navigator.msPointerEnabled) {
+        pointerDownName = 'MSPointerDown';
+        pointerUpName = 'MSPointerUp';
+        pointerMoveName = 'MSPointerMove';
       }
 
       // Simple way to check if some form of pointerevents is enabled or not
@@ -198,7 +198,7 @@
         var STATE_RIGHT_SIDE = 3;
 
         var swipeFrontElement = element.querySelector('.swipe-front');
-        var isAnimating = false;
+        var rafPending = false;
         var initialTouchPos = null;
         var lastTouchPos = null;
         var currentXPosition = 0;
@@ -226,10 +226,10 @@
           }
 
           // Add the move and end listeners
-          if (window.PointerEventsSupport) {
+          if (window.PointerEvent) {
             // Pointer events are supported.
-            document.addEventListener(pointerMoveName, this.handleGestureMove, true);
-            document.addEventListener(pointerUpName, this.handleGestureEnd, true);
+            document.addEventListener('pointermove', this.handleGestureMove, true);
+            document.addEventListener('pointerup', this.handleGestureEnd, true);
           } else {
             // Add Touch Listeners
             document.addEventListener('touchmove', this.handleGestureMove, true);
@@ -248,19 +248,22 @@
         /* // [END handle-start-gesture] */
 
         // Handle move gestures
+        //
+        /* // [START handle-move] */
         this.handleGestureMove = function (evt) {
           evt.preventDefault();
 
           lastTouchPos = getGesturePointFromEvent(evt);
 
-          if(isAnimating) {
+          if(rafPending) {
             return;
           }
 
-          isAnimating = true;
+          rafPending = true;
 
           window.requestAnimFrame(onAnimFrame);
         }.bind(this);
+        /* // [END handle-move] */
 
         /* // [START handle-end-gesture] */
         // Handle end gestures
@@ -271,13 +274,13 @@
             return;
           }
 
-          isAnimating = false;
+          rafPending = false;
 
           // Remove Event Listeners
-          if (window.PointerEventsSupport) {
+          if (window.PointerEvent) {
             // Remove Pointer Event Listeners
-            document.removeEventListener(pointerMoveName, this.handleGestureMove, true);
-            document.removeEventListener(pointerUpName, this.handleGestureEnd, true);
+            document.removeEventListener('pointermove', this.handleGestureMove, true);
+            document.removeEventListener('pointerup', this.handleGestureEnd, true);
           } else {
             // Remove Touch Listeners
             document.removeEventListener('touchmove', this.handleGestureMove, true);
@@ -364,8 +367,9 @@
           return point;
         }
 
+        /* // [START on-anim-frame] */
         function onAnimFrame() {
-          if(!isAnimating) {
+          if(!rafPending) {
             return;
           }
 
@@ -373,18 +377,20 @@
 
           var newXTransform = (currentXPosition - differenceInX)+'px';
           var transformStyle = 'translateX('+newXTransform+')';
-          swipeFrontElement.style['-webkit-transform'] = transformStyle;
-          swipeFrontElement.style['-moz-transform'] = transformStyle;
+          swipeFrontElement.style.webkitTransform = transformStyle;
+          swipeFrontElement.style.MozTransform = transformStyle;
+          swipeFrontElement.style.msTransform = transformStyle;
           swipeFrontElement.style.transform = transformStyle;
 
-          isAnimating = false;
+          rafPending = false;
         }
+        /* // [END on-anim-frame] */
 
         /* // [START addlisteners] */
         // Check if pointer events are supported.
-        if (window.PointerEventsSupport) {
+        if (window.PointerEvent) {
           // Add Pointer Event Listener
-          swipeFrontElement.addEventListener(pointerDownName, this.handleGestureStart, true);
+          swipeFrontElement.addEventListener('pointerdown', this.handleGestureStart, true);
         } else {
           // Add Touch Listener
           swipeFrontElement.addEventListener('touchstart', this.handleGestureStart, true);
--- a/src/content/en/fundamentals/design-and-ui/input/touch/index.md
+++ b/src/content/en/fundamentals/design-and-ui/input/touch/index.md
@@ -15,45 +15,30 @@ description: Touchscreens are available on more and more devices,  from phones u
   </iframe>
 </div>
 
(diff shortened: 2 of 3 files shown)


src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
```diff
@@ -65,9 +65,9 @@
 
         z-index: 2;
 
+        -ms-touch-action: none;
```
RByers: Edge doesn't support the ms prefixed APIs anymore, still worth including?

src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
```diff
@@ -172,14 +172,14 @@
...
-
-      if(window.PointerEvent) {
-        pointerDownName = 'pointerdown';
-        pointerUpName = 'pointerup';
-        pointerMoveName = 'pointermove';
+      var pointerDownName = 'pointerdown';
+      var pointerUpName = 'pointerup';
+      var pointerMoveName = 'pointermove';
+
+      if(window.navigator.msPointerEnabled) {
```
RByers: you're not using these variable below anymore, right? Remove all this?

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -123,121 +75,60 @@ Safari and Chrome add a tap highlight color which can be prevented with the
...
+### Disabling user-select
 
+If you are crafting an experience for touch you may want to make parts of
+your UI which can't have text select, elements like buttons or form input.
+You can do this by setting  the `user-select` CSS property, but beware that
+doing this on content users want to select can be *extremely* infuriating
+for users.
 
-    -moz-user-select: none;
-    -webkit-user-select: none;
```
RByers: FYI, we only just [shipped unprefixed user-select](https://bugs.chromium.org/p/chromium/issues/detail?id=605419) in Chrome 54 - so not quite stable yet.  Fine with me though.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -247,79 +138,66 @@ If, however, you expect users to interact with multiple elements at the same
...
+the recommended approach for implementing custom gestures.
+
+In other browsers `TouchEvents` and `MouseEvents` are the approach to take..
 
-The event names you need to implement are `touchstart`, `touchmove`,
-`touchend` and `touchcancel`.
+The great feature of `PointerEvents` is that it merges multiple types of input,
+including mouse, touch and pen events, into one set of
+callbacks. The events to list for are `pointerdown`, `pointermove` and
+`pointerup`.
```
RByers: If you're going to mention `touchcancel` below then you should probably also mention `pointercancel` which serves a similar purpose (both are pretty obscure these days - eg. if the tab changes while a finger is on the screen).

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -247,79 +138,66 @@ If, however, you expect users to interact with multiple elements at the same
...
-In the short snippet of code above you may have noticed that we only add the starting event listener, this is a conscious decision.
-
-By adding the move and end event listeners once the gesture has
-started on the element itself, the browser can check if the touch occured in a
-region with a touch event listener and if it’s not, can handle it faster
-by not having to run any additional javascript.
+By adding the "move" and "end" event listeners *after* a gesture has
+started, the browser can quickly check if the touch occurred in a
+region with an event listener and if not, it can skips calling
+any JavaScript, which can be a slow process.
```
RByers: I'm not aware of a real performance reason to do this (it's possible that adding/removing the listeners is actually more expensive than just having them there - it's complicated).  But see below.

jpmedley:
@gauntface FYI, I'm doing a copy edit.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -247,79 +138,66 @@ If, however, you expect users to interact with multiple elements at the same
...
 
 The steps taken to implement this are:
 
 1. Add the start events listener to an element.
-1. Inside your touch start method, bind the move and end elements to the
+1. Inside your start gesture callback, bind the move and end elements to the
    document. The reason for binding the move and end events to the
    document is so that we receive all events regardless of whether they
    occur on the original element or not.
```
RByers: Yeah I think this hints at the real reason to generally prefer adding move/end listeners from within the start callback.  Though it's different depending on the type of event (probably more detail than you want to go into here):
- For mouse events, this is the only way you'll get events if the pointer gets ahead of the element you're dragging (as you've described)
- For pointer events you're probably better off using 'setPointerCapture' to indicate you want the target to capture all events. 
- For touch, we do the setPointerCapture implicitly (both for touch events and pointer events, with Edge changing to match us in their pointer events implementation).

So maybe to make this more accurate, just clarify "so that we receive all _mouse_ events regardless ...".

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -328,32 +206,29 @@ and end events to the document:
...
-events no longer firing. By adding the move event to the document, we'll continue to get mouse movements regardless of where they are on the page.
-
-You can use the
-[Show potential scroll bottlenecks](https://developer.chrome.com/devtools/docs/rendering-settings#show-potential scroll bottlenecks)
-feature in Chrome DevTools to show how the touch events behave:
+<div class="attempt-left">
+  <p>By following this pattern of adding the move event to the document, if the
+  user starts interacting with an element and moves their gesture outside of
+  the element, we'll continue to get mouse movements regardless of where they
+  are on the page, because the events are being received from the document.</p>
```
RByers: Ah good, this is exactly what I was saying above.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -328,32 +206,29 @@ and end events to the document:
...
+![Illustrating Binding Touch Events to Document in touchstart](images/scroll-bottleneck.gif)
 
 <div class="clearfix"></div>
 
 
-### Handle Multi-Element Interaction
+#### Handle Multi-Element Interaction
 
 If you expect your users to use multiple elements at once, you can add the
 move and end events listeners directly to the elements themselves. This
```
RByers: The issue with `mouse` events also applies to pointer events for mouse devices (both are "uncaptured" by default).  You can either follow the same pattern of pointer adding handlers to the document, or you can just call `target.setPointerCapture(event.pointerId)` in your `pointerdown` handler to ensure that even if your mouse leaves the element, you'll continue to get `pointermove` events on the target you're listening to.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -457,7 +340,6 @@ of `pan-x | pan-y | manipulation` you prevent the default double-tap
 behavior.
```
RByers: This "|" syntax is possibly confusing.  Maybe just say "manipulation" (since that's the one that disables double-tap without impacting anything else)?  `manipulation` is also [supported by](https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away?hl=en) Chrome, Safari and Firefox now too.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -457,7 +340,6 @@ of `pan-x | pan-y | manipulation` you prevent the default double-tap
...
 
 This allows you to implement a double-tap gesture yourself.
-In the case of IE10+, it also eliminates the 300ms click delay.
 
 Below is a list of the available parameters for *touch-action*.
```
RByers: There are some more values now, but the only one I'd mention explicitly is `pinch-zoom`.  Sadly it's [not in the spec yet](https://github.com/w3c/pointerevents/issues/29), but IE, Edge, Chrome and Firefox will all support it.  I.e. `touch-action: pan-y pinch-zoom` allows vertical scrolling without also disabling pinch-zoom.  This should arguably be recommended over bare `pan-y` for accessibility reasons.

RByers:
Looks great to me, thanks!  Just a few (mostly small) suggestions.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -247,79 +138,66 @@ If, however, you expect users to interact with multiple elements at the same
...
 
-### Handle Single-Element Interaction
-
-In the short snippet of code above you may have noticed that we only add the starting event listener, this is a conscious decision.
-
-By adding the move and end event listeners once the gesture has
-started on the element itself, the browser can check if the touch occured in a
-region with a touch event listener and if it’s not, can handle it faster
-by not having to run any additional javascript.
+By adding the "move" and "end" event listeners *after* a gesture has
```
jpmedley: Also, the generic names move and end should be consistently quoted or consistently _not_ quoted. The simplest thing is to remove them from this line, which I've done in my editing PR. If you feel strongly that it should be the other way, change it back.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
jpmedley: Can you find a way to split or rewrite this sentence. It's a bit of a mouthful.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -328,32 +206,29 @@ and end events to the document:
...
+  user starts interacting with an element and moves their gesture outside of
+  the element, we'll continue to get mouse movements regardless of where they
+  are on the page, because the events are being received from the document.</p>
 
-<img src="images/scroll-bottleneck-devtool.png" alt="Enable Scroll Bottleneck in DevTools">
-
-![Illustrating Binding Touch Events to Document in touchstart](images/scroll-bottleneck.gif){: .attempt-right }
+  <p>This diagram demonstrates what the touch events are doing as we add the
+  *move* and *end* events to the document once a gesture begins.</p>
+</div>
```
jpmedley: I've removed the bolding for the same reason give at line 186.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -378,66 +253,74 @@ remove the mouse event listeners to the document.
...
-little code as possible in the callback to keep our frame rate high,
-preventing jank.
+little code as possible in the callbacks for our events, keeping our frame
+rate high and preventing jank.
 
-Use `requestAnimationFrame` to change the UI in response to
-an event. This gives you an opportunity to update the UI when the browser is intending to draw a frame and will help you move some work out of your callback.
+Using `requestAnimationFrame` we have an opportunity to update the UI when
+the browser is intending to draw a frame and will help us move some work out
+of our event callbacks.
```
jpmedley: This sentence needs some work. I recommend this:

The `requestAnimationFrame()` method lets us update the UI when the browser is drawing a frame and helps us move work out of our callbacks.

petele:
@gauntface - let me know when this is ready to go.

src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
gauntface: Moved it into the older IE support section.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -15,103 +15,55 @@ description: Touchscreens are available on more and more devices,  from phones u
   </iframe>
 </div>
 
-Touchscreens are available on more and more devices, ranging from phones up to desktop screens. When your users choose to interact with your UI, your app should respond to their touch in intuitive and beautiful ways.
+Touchscreens are available on more and more devices, ranging from phones to
+desktop screens. When your users choose to interact with your UI, your app
+should respond to their touch in intuitive and beautiful ways.
```
patrickhlauke: personally, i'd drop "and beautiful" (and in the matching description), but that's just taste i guess

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -123,121 +75,60 @@ Safari and Chrome add a tap highlight color which can be prevented with the
...
 
 ## Implement custom gestures
 
 If you have an idea for custom interactions and gestures for your site, there
-are two topics to keep in mind: how to support the range of mobile browsers
-and how to keep your frame rate high. In this article, we'll look at exactly
-these topics.
+are two topics to keep in mind:
 
+1. How to support the all browsers.
```
patrickhlauke: remove "the"

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -123,121 +75,60 @@ Safari and Chrome add a tap highlight color which can be prevented with the
...
+1. How to keep your frame rate high.
 
-### Respond to Touch Input Using Events
+In this article, we'll look at exactly these topics covering the API's we need
+to support to hit all browsers and then cover how we use these events
+efficiently.
 
-Depending on what you would like to do with touch, you’re likely to fall into
-one of two camps:
+Depending on what you would like your gesture to do, you’re likely want the
```
patrickhlauke: you're likely -> you likely

src/content/en/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html
gauntface: Moved into a seperate html file just for the older IE support section.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
gauntface: Argh great spot - I mis-read http://caniuse.com/#search=user-select

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -247,113 +138,97 @@ If, however, you expect users to interact with multiple elements at the same
...
 
-Touch events and mouse events are implemented on most mobile browsers.
+In other browsers `TouchEvents` and `MouseEvents` are the correct approach.
 
-The event names you need to implement are `touchstart`, `touchmove`,
-`touchend` and `touchcancel`.
+The great feature of `PointerEvents` is that it merges multiple types of input,
+including mouse, touch and pen events, into one set of
+callbacks. The events to listen for are `pointerdown`, `pointermove` and
+`pointerup`.
```
patrickhlauke: and `pointercancel` (if you're mentioning `touchcancel` below)

src/content/en/fundamentals/design-and-ui/input/touch/index.md
gauntface: Done.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -367,80 +242,88 @@ move and end listeners for touch and pointer events to the element straight away
...
 <pre class="prettyprint">
-{% includecode content_path="web/fundamentals/design-and-ui/input/touch/_code/touch-demo-2.html" region_tag="on-anim-frame" adjust_indentation="auto" %}
+{% includecode content_path="web/fundamentals/design-and-ui/input/touch/_code/touch-demo-1.html" region_tag="on-anim-frame" adjust_indentation="auto" %}
 </pre>
 
-### Control Gestures using Touch Actions
+### Control gestures using touch actions
 
 The CSS property `touch-action` allows you to control the default touch
 behavior of an element. In our examples, we use `touch-action: none` to
```
patrickhlauke: maybe add a little warning that `touch-action: none` is a bit of a nuclear option, as it will prevent ANY browser default behavior, and that it should be used with caution/surgically on specific elements, as otherwise you end up having to implement scrolling, long-press, etc yourself

src/content/en/fundamentals/design-and-ui/input/touch/index.md
gauntface: Done.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
```diff
@@ -557,7 +506,7 @@ application:
   </tbody>
 </table>
```
patrickhlauke: Maybe somewhere in here it's worth talking about / pointing to some feature detection/event binding example a la https://w3c.github.io/pointerevents/#examples (which omits the legacy MS prefixed versions, but otherwise gives a rough reference for transitional handling of Pointer/Touch/Mouse and a reminder about Keyboard again)

src/content/en/fundamentals/design-and-ui/input/touch/index.md
gauntface: Should I drop this table completely and just add a note on `touch-action: pan-y pinch-zoom` or just drop and leave it out?

src/content/en/fundamentals/design-and-ui/input/touch/index.md
gauntface: Just to double check:

1. Mouse events need the move and end listeners on the body - best to do in the down event.
1. Touch events capture so no need to add the listener to the body, everything can be on the element itself all times
2. PointerEvents can call setPointerCapture meaning all move events will be tied to the the target element setPointerCapter is called against (i.e. works for a single element and multi-touch scenario)

src/content/en/fundamentals/design-and-ui/input/touch/index.md
gauntface: I think this is going to change heavily based on discussion below.

src/content/en/fundamentals/design-and-ui/input/touch/index.md
RByers: Maybe just change the table from pretending to be an exhaustive list, to just listing the four most common scenarios.  I.e.:
- touch-action: none;    For surfaces where all touches are handled by JavaScript (maps, games, etc.)
- touch-action: pinch-zoom;   Like the above but leaves browser pinch-zoom enabled (eg. maybe your drag and drop example)
- touch-action: pan-y pinch-zoom;   Handle horizontal scrolls in JavaScript without disabling vertical scrolling or pinch-zooming (eg. image carousels).
- touch-action: manipulation;   Just disable double-tap to avoid the click delay, leave all scrolling and pinch-zooming up to the browser.

WDYT?

src/content/en/fundamentals/design-and-ui/input/touch/index.md
RByers: Right.  One more detail: now PointerEvents (in Level 2) also implicity capture (like touch events do) when using a touchscreen. But if you recommend using explicit setPointerCapture (which you probably want for mouse input anyway) then there's no need to mention this detail.

RByers: But it's OK to keep this simple by putting ALL your listeners on the document as you've done - that'll still work.  Just clarify that the only reason you really need to do that is for mouse events (where there's no better option).

petele:
@gauntface - please assign to me when this is ready to go.

gauntface:
Just so people know I haven't forgotten this PR and I will get all the feedback in.

gauntface:
Closing this PR to open a new one against new master (Basically avoiding the hell of the massive git history changes)

PR #3562, opened 2016-09-30
Capturing images from Web Cam

### LINKED ISSUE #371: Capturing an image from the user
TL;DR
Sites such as auto-trader let the user snap and upload a picture of their users cars for inclusion in a listing.
-  Prefer getUserMedia (with camera selection and constraints) allows in-page camera experience
-  Use input type=file accept image/\* capture when not supported.

Open questions, do we also want to document the best way to get the file up to the server, i.e, populate the value of a form field.

PaulKinlan:
Fixes #371


PaulKinlan:
@petele Whilst I would love to get it into for the launch, there is no rush on this one.

PR #3617, opened 2016-10-06
Article: async functions

jakearchibald:
I haven't proof-read this myself, and I don't really have an ending yet, but here it is.

### DIFF
--- a/src/content/en/fundamentals/getting-started/_toc.yaml
+++ b/src/content/en/fundamentals/getting-started/_toc.yaml
@@ -25,6 +25,8 @@ toc:
       path: /web/fundamentals/getting-started/primers/service-workers
     - title: "JavaScript Promises: an Introduction"
       path: /web/fundamentals/getting-started/primers/promises
+    - title: "JavaScript Async Functions"
+      path: /web/fundamentals/getting-started/primers/async-functions
     - title: "Shadow DOM v1: Self-Contained Web Components"
       path: /web/fundamentals/getting-started/primers/shadowdom
     - title: "Custom Elements v1: Reusable Web Components"
--- a/src/content/en/fundamentals/getting-started/primers/async-functions.md
+++ b/src/content/en/fundamentals/getting-started/primers/async-functions.md
@@ -1,6 +1,6 @@
 project_path: /web/_project.yaml
 book_path: /web/fundamentals/_book.yaml
-description: TODO
+description: Async functions allow you to write promise-based code as if it were synchronous
 
 {# wf_published_on: 2016-10-07 #}
 {# wf_updated_on: 2016-10-07 #}
@@ -9,7 +9,10 @@ description: TODO
 
 {% include "web/_shared/contributors/jakearchibald.html" %}
 
-Async functions are enabled by default in Chrome 55 and they're quite frankly marvelous. They allow you to write promise-based code as if it were synchronous, but without blocking the main thread. They make your asynchronous code less "clever" and more readable.
+Async functions are enabled by default in Chrome 55 and they're quite frankly
+marvelous. They allow you to write promise-based code as if it were synchronous,
+but without blocking the main thread. They make your asynchronous code less
+"clever" and more readable.
 
 Async functions look like this:
 
@@ -22,11 +25,16 @@ Async functions look like this:
       }
     }
 
-If you use the `async` keyword before a function definition, you can then use `await` within the function. When you `await` a promise, the function is paused until the promise settles. If the promise fulfills, you get the value back. If the promise rejects, the rejected value is thrown.
+If you use the `async` keyword before a function definition, you can then use
+`await` within the function. When you `await` a promise, the function is paused
+until the promise settles. If the promise fulfills, you get the value back. If
+the promise rejects, the rejected value is thrown.
 
-Note: If you're unfamiliar with promises & promise terminology, check out [our promises guide](/web/fundamentals/getting-started/primers/promises).
+Note: If you're unfamiliar with promises & promise terminology, check out [our
+promises guide](/web/fundamentals/getting-started/primers/promises).
 
-Calling an async function returns a promise for whatever the function returns or throws. So with:
+Calling an async function returns a promise for whatever the function returns or
+throws. So with:
 
     // wait ms milliseconds
     function wait(ms) {
@@ -47,11 +55,14 @@ Calling an async function returns a promise for whatever the function returns or
 
 …calling `foo()` returns a promise that *rejects* with `Error('bar')`.
 
-Note: You can only use `await` directly inside an async function. There's some discussion to allow top-level `await` in modules, but it's [somewhat controversial](https://gist.github.com/Rich-Harris/0b6f317657f5167663b493c722647221).
+Note: You can only use `await` directly inside an async function. There's some
+discussion to allow top-level `await` in modules, but it's [somewhat
+controversial](https://gist.github.com/Rich-Harris/0b6f317657f5167663b493c722647221).
 
 ## Example: Logging a fetch
 
-Say we wanted to fetch a URL and log the response as text. Here's how it looks using promises:
+Say we wanted to fetch a URL and log the response as text. Here's how it looks
+using promises:
 
     function logFetch(url) {
       return fetch(url)
@@ -75,11 +86,13 @@ And here's the same thing using async functions:
       }
     }
 
-It's the same number of lines, but all the callbacks are gone. This makes it way easier to read, especially for those less familiar with promises.
+It's the same number of lines, but all the callbacks are gone. This makes it way
+easier to read, especially for those less familiar with promises.
 
 ## Other async function syntax
 
-We've seen `async function() {}` already, but the `async` keyword can be used with other function syntax:
+We've seen `async function() {}` already, but the `async` keyword can be used
+with other function syntax:
 
 ### Arrow functions
 
@@ -89,7 +102,9 @@ We've seen `async function() {}` already, but the `async` keyword can be used wi
       return response.json();
     });
 
-Note: `array.map(func)` doesn't care that I gave it an async function, it just sees it as a function that returns a promise. It won't wait for the first function to complete before calling the second.
+Note: `array.map(func)` doesn't care that I gave it an async function, it just
+sees it as a function that returns a promise. It won't wait for the first
+function to complete before calling the second.
 
 ### Object methods
 
@@ -122,7 +137,8 @@ Note: Class constructors cannot be async.
 
 ## Example: Streaming a response
 
-The benefit of async functions increases in more complex examples. Say we wanted to stream a response while logging out the chunks and returning the final size.
+The benefit of async functions increases in more complex examples. Say we wanted
+to stream a response while logging out the chunks and returning the final size.
 
 Note: The phrase "logging out the chunks" made me sick in my mouth.
 
@@ -145,7 +161,11 @@ Here it is with promises:
       });
     }
 
-Check me out, Jake "wielder of promises" Archibald. See how I'm calling `processResult` inside itself to set up an asynchronous loop? Writing that made me feel *very smart*. But like most "smart" code, you have to stare at it for ages to figure out what it's doing, like one of those magic-eye pictures from the 90's.
+Check me out, Jake "wielder of promises" Archibald. See how I'm calling
+`processResult` inside itself to set up an asynchronous loop? Writing that made
+me feel *very smart*. But like most "smart" code, you have to stare at it for
+ages to figure out what it's doing, like one of those magic-eye pictures from
+the 90's.
 
 Let's try that again with async functions:
 
@@ -164,13 +184,22 @@ Let's try that again with async functions:
       return total;
     }
 
-All the "smart" is gone. The asynchronous loop that made me feel so smug is replaced with a trusty, boring, while-loop. Much better. In future, we'll get [async iterators](https://github.com/tc39/proposal-async-iteration){: .external}, which would [replace the above `while` loop](https://gist.github.com/jakearchibald/0b37865637daf884943cf88c2cba1376){: .external}.
+All the "smart" is gone. The asynchronous loop that made me feel so smug is
+replaced with a trusty, boring, while-loop. Much better. In future, we'll get
+[async iterators](https://github.com/tc39/proposal-async-iteration){:
+.external}, which would [replace the above `while`
+loop](https://gist.github.com/jakearchibald/0b37865637daf884943cf88c2cba1376){:
+.external}.
 
-Note: I'm sort-of in love with streams. If you're unfamiliar with streaming, [check out my guide](https://jakearchibald.com/2016/streams-ftw/#streams-the-fetch-api){: .external}.
+Note: I'm sort-of in love with streams. If you're unfamiliar with streaming,
+[check out my
+guide](https://jakearchibald.com/2016/streams-ftw/#streams-the-fetch-api){:
+.external}.
 
 ## Example: Avoid going too sequential
 
-Say we wanted to fetch a series URLs and log them as soon as possible, in the correct order.
+Say we wanted to fetch a series URLs and log them as soon as possible, in the
+correct order.
 
 *Deep breath* - here's how that looks with promises:
 
@@ -187,9 +216,11 @@ Say we wanted to fetch a series URLs and log them as soon as possible, in the co
       }, Promise.resolve());
     }
 
-Yeah, that's right, I'm using `reduce` to chain a sequence of promises. I'm *so smart*. But this is a bit of *so smart* coding we're better off without.
+Yeah, that's right, I'm using `reduce` to chain a sequence of promises. I'm *so
+smart*. But this is a bit of *so smart* coding we're better off without.
 
-However, when converting the above to an async function, it's tempting to go *too sequential*:
+However, when converting the above to an async function, it's tempting to go
+*too sequential*:
 
 <span class="compare-worse">Not recommended</span> - too sequential
 
@@ -200,7 +231,9 @@ However, when converting the above to an async function, it's tempting to go *to
       }
     }
 
-Looks much neater, but my second fetch doesn't begin until my first fetch has been fully read, and so on. This is much slower than the promises example that performs the fetches in parallel. Thankfully there's an ideal middle-ground:
+Looks much neater, but my second fetch doesn't begin until my first fetch has
+been fully read, and so on. This is much slower than the promises example that
+performs the fetches in parallel. Thankfully there's an ideal middle-ground:
 
 <span class="compare-better">Recommended</span> - nice and parallel
 
@@ -217,11 +250,13 @@ Looks much neater, but my second fetch doesn't begin until my first fetch has be
       }
     }
 
-In this example, the URLs are fetched and read in parallel, but the "smart" `reduce` bit is replaced with a standard, boring, readable for-loop.
+In this example, the URLs are fetched and read in parallel, but the "smart"
+`reduce` bit is replaced with a standard, boring, readable for-loop.
 
 ## Browser support & workarounds
 
-At time of writing, async functions are enabled by default in Chrome 55, but they're being developed in all the main browsers:
+At time of writing, async functions are enabled by default in Chrome 55, but
+they're being developed in all the main browsers:
 
 * Edge - [In build 14342+ behind a flag](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/asyncfunctions/)
 * Firefox - [active development](https://bugzilla.mozilla.org/show_bug.cgi?id=1185106)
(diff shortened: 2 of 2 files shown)


src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+description: TODO
+
+{# wf_published_on: 2016-10-07 #}
+{# wf_updated_on: 2016-10-07 #}
+
+# Async functions - making promises friendly {: .page-title }
+
+{% include "web/_shared/contributors/jakearchibald.html" %}
+
+Async functions are enabled by default in Chrome 55 and they're quite frankly marvelous. They allow you to write promise-based code as if it were synchronous, but without blocking the main thread.
```
addyosmani: `Async functions` link to spec or MDN page?

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+      yield wait(1000);
+      return val;
+    });
+
+Note that you have to pass in a generator (`function*`) and use `yield` instead of `await`. Other than that it works the same.
+
+### Workaround - regenerator
+
+If you're targeting older browsers, Babel can also transpile generators, allowing you to use async functions all the way down to IE8.
+
```
addyosmani: Thanks for calling out the path to older browser support here too.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+# Async functions - making promises friendly {: .page-title }
+
+{% include "web/_shared/contributors/jakearchibald.html" %}
+
+Async functions are enabled by default in Chrome 55 and they're quite frankly marvelous. They allow you to write promise-based code as if it were synchronous, but without blocking the main thread.
+
+Async functions look like this:
+
+    async function myFirstAsyncFucntion() {
```
ebidel: Fucntion -> Function

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+      try {
+        const resolvedValue = await aPromise;
+      }
+      catch (rejectedValue) {
+        // …
+      }
```
ebidel: so because async functions always return a promise, you don't explicitly need a return val? Might want to mention that.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+If you use the `async` keyword before a function definition, you can then use `await` within the function. When you `await` a promise, the function is paused until the promise settles. If the promise fulfills, you get the value back. If the promise rejects, the rejected value is thrown.
```
philipwalton: This is a nit, but do you want to say "resolves" instead of "fulfills"?

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+    function wait(ms) {
+      return new Promise(r => setTimeout(r, ms));
+    }
+
+    async function hello() {
+      await wait(500);
+      return 'world';
+    }
+
+…calling `hello()` returns a promise that *fulfills* with `"world"`.
```
ebidel: would be good to show the examples of the calls for each of these.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+    };
+
+### Class methods
+
+    class Storage {
+      constructor() {
+        this.cachePromise = caches.open('avatars');
+      }
+
+      async getAvatar(name) {
```
ebidel: show an example usage of `getAvatar()`

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+      }
+
+      return total;
+    }
+
+The asynchronous loop that made me feel so smug is replaced with a trusty, boring, while-loop. Much better.
+
+Note: The phrase "logging out the chunks" made me sick in my mouth.
+
+## Example: Avoid going too linear
```
philipwalton: I didn't know what you meant by "too linear" when first reading this heading. It took me reading the examples to figure it out.

I wonder if sequential is a better word.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+
+      for (const textPromise of textPromises) {
+        console.log(await textPromise);
+      }
+    }
+
+In this example, the URLs are fetched and read as text in parallel, but the "smart" `reduce` bit is replaced with a standard, readable, for-loop.
+
+## Browser support & workarounds
+
```
ebidel: What do you think about using a framebox here to embed the chromestatus info:
https://www.chromestatus.com/feature/5643236399906816?embed

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: The opposite of reject is "fulfill". A promise can resolve to a rejected promise, eg `Promise.resolve(Promise.reject(new Error("foo")))` resolves to a rejected promise.

src/content/en/fundamentals/getting-started/primers/async-functions.md
PaulKinlan: lol - I am rubbing off on Jake.  I said this to him yesterday that I can write function properly.

jakearchibald:
Oh, I need to wrap all the lines too. Oops.

src/content/en/fundamentals/getting-started/primers/async-functions.md
philipwalton: Interetesting. TIL.

For anyone else seeing this, I also found this helpful:
https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md

jeffposnick: You could use this space to explicitly call out that top-level `await` [isn't supported](https://github.com/tc39/ecmascript-asyncawait/issues/9#issuecomment-127427447), and that the **only** time you can use `await` is inside an `async` function.

That's something that I found non-obvious.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,249 @@
...
+        return cache.match(`/avatars/${name}.jpg`);
+      }
+    }
+
+Note: Class constructors cannot be async.
+
+## Example: Streaming a response
+
+Note: If you're unfamiliar with streaming, [check out my guide](https://jakearchibald.com/2016/streams-ftw/#streams-the-fetch-api){: .external}.
+
```
jakearchibald: Keep on streamin'

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Yeah, much better

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Hah, yep. Good catch.

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Good idea

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Isn't this a bit huge for an embed?

jakearchibald:
I'm pretty happy with this now. @addyosmani anyone need to review this?

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,317 @@
...
+    async function myFirstAsyncFunction() {
+      try {
+        const fulfilledValue = await promise;
+      }
+      catch (rejectedValue) {
+        // …
+      }
+    }
+
+If you use the `async` keyword before a function definition, you can then use
```
ithinkihaveacat: Maybe not for this article, but why is `async` even necessary? (Why is `await` appearing in the body not sufficient?) Is this a convenience to the JS engine?

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,317 @@
...
+Calling an async function returns a promise for whatever the function returns or
+throws. So with:
+
+    // wait ms milliseconds
+    function wait(ms) {
+      return new Promise(r => setTimeout(r, ms));
+    }
+
+    async function hello() {
+      await wait(500);
```
ithinkihaveacat: Might be worth explaining why adding an additional `await wait(500)` will result in this function taking 1s to run, and why something like

```js
const p1 = wait(500);
const p2 = wait(500);
const v1 = await p1;
const v2 = await p2;
```

(which seems more or less equivalent) takes 0.5s.

jakearchibald:
@ithinkihaveacat 

> Perhaps add a section on async/await gotchas? Also, are there any situations in which raw promises are preferable or easier to understand?

The biggest gotcha is forgetting that you're async, or making thinks too sequential, which I cover.

Generally it's ok to use async functions for everything you'd previously use promises for.

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: I guess the reasons are similar to why `function()*` is necessary for generators, but I'm not 100% sure.

FWIW I've been caught out by generators in Python in the past. It wasn't clear to me that a function would behave very differently because it has a `yield` hidden away in it somewhere.

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Ohh, this is a good example. Let me see if I can work it in.

jakearchibald: Added this at the start of the series vs parallel bit. Cheers!

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+      async getAvatar(name) {
+        const cache = await this.cachePromise;
+        return cache.match(`/avatars/${name}.jpg`);
+      }
+    }
+
+    const storage = new Storage();
+    storage.getAvatar('jaffathecake').then(…);
+
+Note: Class constructors cannot be async.
```
beaufortfrancois: `get` and `set` as well. However it looks like `static` works fine.
Shall we be exhaustive there @jakearchibald?

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+      });
+    }
+
+Check me out, Jake "wielder of promises" Archibald. See how I'm calling
+`processResult` inside itself to set up an asynchronous loop? Writing that made
+me feel *very smart*. But like most "smart" code, you have to stare at it for
+ages to figure out what it's doing, like one of those magic-eye pictures from
+the 90's.
+
+Let's try that again with async functions:
```
beaufortfrancois: Shall we also add the `catch` to show how to handle errors?

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+
+    async function parallel() {
+      const wait1 = wait(500);
+      const wait2 = wait(500);
+      await wait1;
+      await wait2;
+      return "done!";
+    }
+
+…the above takes 500ms to complete, because both waits happen at the same time.
```
beaufortfrancois: We may want to call out `observables` are different than `promises` as they always start no matter what when you call them unlike `observables`.

src/content/en/fundamentals/getting-started/primers/async-functions.md
hemanth: You might like [this](https://h3manth.com/new/blog/2015/are-you-async-yet/) post 😇

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+
+Async functions look like this:
+
+    async function myFirstAsyncFunction() {
+      try {
+        const fulfilledValue = await promise;
+      }
+      catch (rejectedValue) {
+        // …
+      }
```
littledan: I see a lot of code samples like this, with the try/catch. I worry that presenting async functions like this may make people think they always need a try/catch block inside their async function body, which definitely isn't true. Maybe the initial example should be a chain of awaiting multiple things, or even a loop, and then explain the exceptions/rejection duality later.

src/content/en/fundamentals/getting-started/primers/async-functions.md
littledan: There are a few reasons:
- When you have such drastically different modes, it's a benefit to readability to have it called out when the function starts (as Jake says above).
- You can have an async function which doesn't contain any await; the semantics will be kinda sorta like if `return foo` was return `Promise.resolve(foo)`
- `await` is a new keyword, but we can't "just" add keywords. They might already be used as variables. Scoping the keyword-ness to async functions is one technique that `await` uses. (`async` uses other techniques, e.g., that if it's followed by `function`, it would've been a syntax error.)

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+If you use the `async` keyword before a function definition, you can then use
+`await` within the function. When you `await` a promise, the function is paused
+until the promise settles. If the promise fulfills, you get the value back. If
+the promise rejects, the rejected value is thrown.
```
littledan: Is this the right paragraph to lead with? Maybe the article should open with an explanation and examples of, "you can await a bunch of stuff in a row!" and then get into the nitty gritty of exception handling later.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+
+Note: If you're unfamiliar with promises & promise terminology, check out [our
+promises guide](/web/fundamentals/getting-started/primers/promises).
+
+Calling an async function returns a promise for whatever the function returns or
+throws. So with:
+
+    // wait ms milliseconds
+    function wait(ms) {
+      return new Promise(r => setTimeout(r, ms));
```
littledan: Generally, I like this example because it shows a simple async/await usage without exceptions. However, could we open with something that uses platform-provided Promises, rather than the Promise constructor (whose usage, long-term, should be minimal)?

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+
+    async function logFetch(url) {
+      try {
+        const response = await fetch(url);
+        console.log(await response.text());
+      }
+      catch (err) {
+        console.log('fetch failed', err);
+      }
+    }
```
littledan: Awesome example, much cleaner and gives a good use case for exception handling/rejection. Maybe you should move that opening try/catch example to just above this.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+We've seen `async function() {}` already, but the `async` keyword can be used
+with other function syntax:
+
+### Arrow functions
+
+    // map some URLs to json-promises
+    const jsonPromises = urls.map(async url => {
+      const response = await fetch(url);
+      return response.json();
+    });
```
littledan: Another cool thing you can do here is `Promise.all(foo.map(async elt => fn(elt)))`.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+
+Let's try that again with async functions:
+
+    async function getResponseSize(url) {
+      const response = await fetch(url);
+      const reader = response.body.getReader();
+      let result;
+      let total = 0;
+
+      while ((result = await reader.read()) && !result.done) {
```
littledan: I like how this uses a loop, and how it uses a bunch of await calls, and seems really practical. However, maybe it would be easier to read if you didn't use the pattern of assignment `&&` another expression, instead breaking it out into another line (otherwise it still may have the smell of "wielder of something").

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+      }
+
+      return total;
+    }
+
+All the "smart" is gone. The asynchronous loop that made me feel so smug is
+replaced with a trusty, boring, while-loop. Much better. In future, we'll get
+[async iterators](https://github.com/tc39/proposal-async-iteration){:
+.external}, which would [replace the above `while`
+loop](https://gist.github.com/jakearchibald/0b37865637daf884943cf88c2cba1376){:
```
littledan: You could write inline, "would replace the while loop with a for-of loop".

src/content/en/fundamentals/getting-started/primers/async-functions.md
littledan: Not sure why observables would be mentioned here.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+      // fetch all the URLs in parallel
+      const textPromises = urls.map(async url => {
+        const response = await fetch(url);
+        return response.text();
+      });
+
+      // log them in sequence
+      for (const textPromise of textPromises) {
+        console.log(await textPromise);
+      }
```
littledan: Great example

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Yeahhhh why not

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: I'll try and pick up on this elsewhere in the article. I kinda like opening with how fulfill/reject values map in async functions.

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Yeah, moving this to after the first example

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Shall do

jakearchibald:
I've reordered the article a little so readers get to the examples quicker. I'm happy for this to be merged btw.

src/content/en/fundamentals/getting-started/_toc.yaml
```diff
@@ -25,6 +25,8 @@ toc:
       path: /web/fundamentals/getting-started/primers/service-workers
     - title: "JavaScript Promises: an Introduction"
       path: /web/fundamentals/getting-started/primers/promises
+    - title: "JavaScript Async Functions"
```
petele: You only need to quote if you're using a 'fun' character like `:`

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+        // get the next result
+        result = await reader.read();
+      }
+
+      return total;
+    }
+
+All the "smart" is gone. The asynchronous loop that made me feel so smug is
+replaced with a trusty, boring, while-loop. Much better. In future, we'll get
+[async iterators](https://github.com/tc39/proposal-async-iteration){:
```
petele: don't break `{: something }` attributes, the parser isn't smart enough to handle this. :( keep them all on one line.

src/content/en/fundamentals/getting-started/primers/async-functions.md
```diff
@@ -0,0 +1,345 @@
...
+        const response = await fetch(url);
+        console.log(await response.text());
+      }
+    }
+
+Looks much neater, but my second fetch doesn't begin until my first fetch has
+been fully read, and so on. This is much slower than the promises example that
+performs the fetches in parallel. Thankfully there's an ideal middle-ground:
+
+<span class="compare-better">Recommended</span> - nice and parallel
```
petele: 👍  nice use

petele:
Closing this and replacing with #3716 to more easily merge into master.

src/content/en/fundamentals/getting-started/primers/async-functions.md
jakearchibald: Damnit, that's what my auto-wrapping plugin does. Will watch out for this, cheers!

PR #3631, opened 2016-10-06
[lighthouse] add moar audit references

kaycebasques:
@paullewis @paulirish @addyosmani @brendankenny here's the last of the reference docs for the currently-active Lighthouse audits

cc @jpmedley in case you want to review


kaycebasques:
see #3693

PR #3661, opened 2016-10-11
update devtools geolocation guidance

kaycebasques:
@PaulKinlan the DevTools section in your geolocation doc was a little old, so I gave it a touch up


PaulKinlan:
My guidance is timeless!!! :)

PaulKinlan:
LGTM.

I think this is great and can be merged.  One thing that might be good in another PR is to show the actual locations drop-down with the list of locations later in the doc.

PR #3663, opened 2016-10-12
Make video always take 100% width

surma:
Not sure if this is the right CSS file to put this into, but currently `<video>` is not responsive.

screenshot 2016-10-12 12 20 36
screenshot 2016-10-12 12 20 26

Let me know if I should move the styles somewhere else.


petele:
Sadly this will only fix it in staging, I've got a but open against me that I need to fix this on DevSite. Will prioritize.

surma:
Thanks @petele

petele:
Closing, will use #3688 to track.

PR #3678, opened 2016-10-13
ci: deploy to/from master instead of next-version

x1ddos:


petele:
@x1ddos - Thanks for jumping on this without even being asked!

x1ddos:
@petele I've just realized there's a missing bit, `gae/lib` after next-version => master migration: https://github.com/google/WebFundamentals/tree/next-version/gae

It was handy because then there was no need to do "pip install" - everything was in `gae/lib` already.
Do you think it would possible to reinstate it in master, too?

PR #3679, opened 2016-10-13
Replaced incorrect `cardholderNumber` with `cardNumber` credit card f…

reaktivo:
…ield name.

PaymentResponse details for credit cards `cardholderNumber` field should be replaced with `cardNumber` reflecting Chrome 53 on Android implementation and the [W3C spec](https://www.w3.org/TR/payment-method-basic-card/#basiccardresponse).


reaktivo:
I signed it!

PR #3687, opened 2016-10-13
Tweaks to capturing-images article

samdutton:
I've changed the examples to use `srcObject`, which is widely implemented now.

Couple of nitpicks: the `capture` attribute is boolean, and the stream object from a `getUserMediaI()` call is a `MediaStream`.

A few other suggestions below.

* Maybe add catch clauses?

* Might be worth mentioning the video option for the <input> method, i.e: 

`<input type="file" accept="video/*" capture />`

* Maybe mention that you can set the canvas dimensions from videoWidth and videoHeight — and it's probably better to set the canvas dimensions in JavaScript rather than in the HTML.

* With the examples (e.g. upload and image on a canvas to a server) I wonder if it would be worth adding how-to suggestions?

* You might want to link to samples:
  * [simpl.info/mediacapture](http://simpl.info/mediacapture/)
  * [webrtc.github.io/samples/src/content/getusermedia/gum](https://webrtc.github.io/samples/src/content/getusermedia/gum/)
  * [webrtc.github.io/samples/src/content/getusermedia/canvas](https://webrtc.github.io/samples/src/content/getusermedia/canvas/)

### DIFF
--- a/src/content/en/fundamentals/native-hardware/capturing-images/index.md
+++ b/src/content/en/fundamentals/native-hardware/capturing-images/index.md
@@ -201,3 +201,9 @@ settings.
 
 Warning: Asking for access to the camera on page load will result in most of 
 your users rejecting access to it.
+
+## Compatibility
+
+More information about mobile and desktop browser implementation:
+* [srcObject](https://www.chromestatus.com/feature/5989005896187904)
+* [navigator.mediaDevices.getUserMedia()](https://www.chromestatus.com/features/5755699816562688)
--- a/src/content/en/fundamentals/native-hardware/capturing-images/index.md
+++ b/src/content/en/fundamentals/native-hardware/capturing-images/index.md
@@ -17,21 +17,21 @@ experience, or it could be delegated to another app on the user's device.
 
 The easiest thing to do is simply ask the user for a pre-recorded
 file. Do this by creating a simple file input element and adding 
-an `accept` filter that indicates we can only accept audio files and ideally we 
+an `accept` filter that indicates we can only accept image files and ideally we 
 will get them directly from the camera.
 
-    <input type="file" accept="image/*" capture="camera">
+    <input type="file" accept="image/*" capture>
 
 This method works on all platforms. On desktop it will prompt the user to 
-upload an image file from the file system (ignoring `capture="camera"`). In Safari
-on iOS it will open up the camera app, allowing you to capture an image and 
-then send it back to the web page; on Android it will give the user the 
-choice of which app to use to capture the image before sending it back to the
+upload an image file from the file system. In Safari
+on iOS this method will open up the camera app, allowing you to capture an image 
+and then send it back to the web page; on Android this method will give the user 
+a choice of which app to use to capture the image before sending it back to the
 web page.
 
 The data can then be attached to a `<form>` or manipulated with JavaScript by 
 listening for an `onchange` event on the input element and then reading 
-the `files` property of the event object.
+the `files` property of the event `target`.
 
 ### Capture a single frame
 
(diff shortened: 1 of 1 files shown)


src/content/en/fundamentals/native-hardware/capturing-images/index.md
```diff
@@ -53,51 +53,47 @@ Getting access to the image file is simple.
...
+specification called `getUserMedia()`. This will prompt the user for 
 access to their connected microphones and cameras.
 
-If successful the API will return a `Stream` that will contain the data from
-the camera, and we can then either attach it to an `<video>` element and play it
-to show a real time preview or we could attach it to a `<canvas>` to get a
+If successful the API will return a `MediaStream` that contains data from
+the camera, and we can then either attach it to a `<video>` element and play it
+to show a real time preview, or attach it to a `<canvas>` to get a
 snapshot.
```
fippo: the promise needs a catch. Maybe show how to check if a camera is present as well or at least hint to it and point to https://webrtc.github.io/samples/src/content/devices/input-output/ ?

src/content/en/fundamentals/native-hardware/capturing-images/index.md
```diff
@@ -194,11 +191,13 @@ If the user has not previously granted your site access to the camera then
...
+Users hate getting prompted for access to powerful devices on their machine 
+and they will frequently block the request, or they will ignore it if they don't 
+understand the context for which the prompt has been created. It is best 
+practice to only ask to access the camera when first needed. Once the user has
+granted access they won't be asked again. However, if the user rejects access, 
+you can't get access again, unless they manually change camera permission 
+settings.
 
-Warning: Asking for access to the camera on page load will result in most of your users rejecting access to it.
+Warning: Asking for access to the camera on page load will result in most of
```
fippo: anecdotally: this kind of works if you explain why you want access. We did an A/B experiment on appear.in a while back asking for GUM permissions only after a button click and that did not significantly lower the permission denied errors.

PaulKinlan:
srcObject from what I can tell is not well supported given we still have a large number of people still pre-52 I don't think we should be suggesting it yet.

src/content/en/fundamentals/native-hardware/capturing-images/index.md
PaulKinlan: we have a lot of other data on permissions requests

src/content/en/fundamentals/native-hardware/capturing-images/index.md
```diff
@@ -107,16 +103,16 @@ it back.
...
 
 The process is as follows:
 
 1. Create a canvas object that will hold the frame from the camera
 2. Get access to the camera stream
 3. Attach it to a video element
-4. When you want to capture a precise frame, add the video element to a 
-   canvas object using `addImage()` and it will appear in the canvas element.
+4. When you want to capture a precise frame, add the data from the video element 
+   to a canvas object using `drawImage()`.
```
PaulKinlan: urgh. good spot.

PaulKinlan:
Whilst I am nit-picking about srcObject I these changes are good.

petele:
@samdutton - can you please address and then assign back to me when you're satisfied.

fippo:
@PaulKinlan native navigator.mediaDevices.getUserMedia was only landed in M53 [here](https://bugs.chromium.org/p/chromium/issues/detail?id=503227). Probably worth adding a compability note about that.

samdutton:
@PaulKinlan 

> srcObject from what I can tell is not well supported given we still have a large number of people still pre-52 I don't think we should be suggesting it yet.

> native navigator.mediaDevices.getUserMedia was only landed in M53 here. Probably worth adding a compatibility note about that.

Hi Paul — how about a compatibility note as @fippo suggests, pointing at chromestatus for `srcObject` and `navigator.mediaDevices.getUserMedia()`? 

I guess we'll have at least a small numbers of pre-52 users for a fair while (currently ~2.5% according to [StatCounter](http://gs.statcounter.com/#all-browser_version-ww-monthly-201509-201610), usual caveats). We would like to encourage devs to move to `srcObject` if possible.

PaulKinlan:
Sg

beaufortfrancois:
Last nits:

- Replace `// Do something with the audio file.` comment with `// Do something with the image file.`
- Replace `navigator.mediaDevices.getUserMedia({audio: false, video: true})` with `navigator.mediaDevices.getUserMedia({video: true})` 
- You may want to highlight `var videoTracks;` as well in the last example.

PR #3704, opened 2016-10-18
Updates post for NIC54

petele:


petele:
@jpmedley - This is a new updates post format that I'm trying for the New In Chrome video series. I've literally copied and pasted the script used on the teleprompter here. I don't want to change it too much, since it's the script I used, but there may be spelling mistakes, etc.

PR #3733, opened 2016-10-20
Correcting babel reference

jakearchibald:


hzoo:
👍 , can also change the other stage-3 reference below

jakearchibald:
@hzoo oops, updated.

PR #3749, opened 2016-10-21
[Italian] - Translation of the "caution: article may be outdated" paragraph

mendaomn:
It's a really small change, but I hope it can help!

Keep up!


petele:
Thanks so much - that's awesome!

PR #3752, opened 2016-10-21
Remove needless `scriptElement.async = true`

mathiasbynens:
See https://mathiasbynens.be/notes/async-analytics-snippet#async.


petele:
@ebidel - can you take a peek? If it looks good, please approve and assign to me.

ebidel:
Yea not needed but I always do it for good measure. Lgtm

PR #3771, opened 2016-10-24
Add ericlawrence to Contributors

ericlaw1979:


src/data/_contributors.yaml
```diff
@@ -702,6 +702,21 @@ sethladd:
   role:
     - author
 
+ericlawrence
```
petele: missing colon at end of key

PR #3776, opened 2016-10-25
Add styles image for CSS articles on updates/

jpmedley:
Add an image for use on CSS articles and any other place it may be appropriate.


jpmedley:
Update the commit name.

PR #3795, opened 2016-10-27
Update "Avoid Not Secure Warning" with latest text

ericlaw1979:
Since writing the original article, we've landed a few tweaks to Canary. My updates keep the WebFundamentals article in sync with what we're shipping.


src/content/en/updates/2016/10/avoid-not-secure-warn.md
```diff
@@ -24,17 +24,20 @@ this warning.
...
-first version to include detection of sensitive fields. You can see an
-example of the browser’s warning behavior on 
-[this page](http://http-password.badssl.com/) if you open it in Chrome
-56.0.2895.0 or later, then view the console in DevTools.
-
-When the Not Secure state is shown, the DevTools console shows the message `In
-Chrome M56 (Jan 2017), this page will be marked as “not secure” in the URL bar.
-For more information, see [https://goo.gl/zmWq3m](https://security.googleblog.com/2016/09/moving-towards-more-secure-web.html).`
+To test the upcoming user experience before that time, install the latest 
+[Google Chrome Canary](https://www.google.com/chrome/browser/canary.html) build.
```
jpmedley: You need to mention the specific version of Chrome that is in canary so that it will make sense a year from now.

ericlaw1979:
The team requested I remove the version info in favor of simply saying "Latest Canary."

After Chrome 56 launches, the entire paragraph will get removed, because none of it will be necessary and the option mentioned will likely vanish too.

PR #3801, opened 2016-10-28
introduce interactive tutorial in DOM breakpoints section

kaycebasques:
Hey hey @petele I'm piloting interactive tutorials in the DevTools docs. Pretty excited about that. Did I use `framebox` correctly here? Also, I added a little styling to make inline images display more nicely. I'd like to add this as a general rule but I believe I'm blocked on that (you can ping me if you don't know what I'm talking about). 

cc @jpmedley in case you want to writer's review

Notes for posterity:
- I'm also testing out using less screenshots and more emphasis on UI labels / small inline images of relevant UI icons. If the docs are still usable like this, then this will aid doc longevity.


src/content/en/tools/chrome-devtools/javascript/add-breakpoints.md
```diff
@@ -123,25 +130,95 @@ breakpoints** button
...
+   button](imgs/resume-script-execution.png){:.inline} twice to resume
+   script execution. You need to press it twice because the breakpoint is
+   triggered once when the count text is deleted, and then again when the
+   text is updated with the new count.
+
+To break when an attribute of the selected node is changed, or when the 
+selected node is deleted, just select **Attributes modifications** or
+**Node Removal** instead of **Subtree Modifications** in step 2 above.
+
+Tip: These breakpoints are not exclusive. You can have two or all three of these breakpoints enabled on a single node at the same time.
```
petele: We don't have a tip style, maybe note or success?

src/content/en/tools/chrome-devtools/javascript/add-breakpoints.md
```diff
@@ -5,6 +5,13 @@ description: Use breakpoints to pause your JavaScript code and investigate the v
 {# wf_updated_on: 2016-07-17 #}
 {# wf_published_on: 2015-04-13 #}
 
+<style>
```
petele: DevSite already has an `.inline` style, should prob use a different name.

kaycebasques:
✔️ Renamed `.inline` to `.devtools-inline`

src/content/en/tools/chrome-devtools/javascript/add-breakpoints.md
kaycebasques: I tried note, but it's not really jiving. I'll just leave it plaintext and maybe advocate for a new tip style.

PR #3811, opened 2016-11-01
performance/prpl: Adjust wording around HTTP/2 push

surma:
r: @addyosmani @jpmedley 
fyi: @samccone

### DIFF
--- a/src/content/en/fundamentals/performance/prpl-pattern/index.md
+++ b/src/content/en/fundamentals/performance/prpl-pattern/index.md
@@ -224,9 +224,9 @@ means they can be cached efficiently and be shared between pages.
 
 HTTP/2 Push needs to be utilized with care, as it forces data to the browser,
 even if the file is already in the browser’s local cache or bandwidth is
-already saturated. If done wrong, performance can suffer. [Resource hints]
-might be a good alternative to allow the browser to make smart decisions about
-the prioritization of these requests.  
+already saturated. If done wrong, performance can suffer. 
+[`<link rel="preload">`][Resource hints] might be a good alternative to allow 
+the browser to make smart decisions about the prioritization of these requests.  
 
 ## Conclusion
 
--- a/src/content/en/fundamentals/performance/prpl-pattern/index.md
+++ b/src/content/en/fundamentals/performance/prpl-pattern/index.md
@@ -119,7 +119,7 @@ needed by the shell.
 
 The build process builds a graph of all of these dependencies, and the server
 uses this information to serve the files efficiently. It also builds a set of
-vulcanized bundles, for browsers that don't support HTTP2 push.
+vulcanized bundles, for browsers that don't support HTTP2.
 
 ### App entrypoint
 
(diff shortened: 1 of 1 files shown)


src/content/en/fundamentals/performance/prpl-pattern/index.md
```diff
@@ -221,6 +222,12 @@ The combination of HTTP/2 and HTTP/2 server push provides the _benefits_ of
 bundling (reduced latency) without actual bundling. Keeping resources separate
 means they can be cached efficiently and be shared between pages.
 
+HTTP/2 Push needs to be utilized with care, as it forces data to the browser,
```
samccone: sooo there is even more nuance here. If you are following PRPL 100% you will have a sw which will intercept any future connections to the server so you will actually not be pushing since there is no open h2 connection to say index.html

src/content/en/fundamentals/performance/prpl-pattern/index.md
```diff
@@ -162,7 +163,7 @@ Your server logic should deliver the appropriate build for each browser.
 
 ### Bundled build
 
-For browsers that don't handle HTTP2 Push, the build process could produce a set
+For browsers that don't handle HTTP2, the build process could produce a set
```
addyosmani: We should probably stay consistent with HTTP/2 being used (vs HTTP2)

src/content/en/fundamentals/performance/prpl-pattern/index.md
surma: True. This is very high level in any case, not sure this list needs to be exhaustive here. I am mainly worried about making PUSH look like a golden bullet. But happy to adjust if you think it’s the better choice.

src/content/en/fundamentals/performance/prpl-pattern/index.md
surma: I was asking myself the same question, but it was “HTTP2” before :D Happy it fix that

src/content/en/fundamentals/performance/prpl-pattern/index.md
```diff
@@ -221,6 +222,12 @@ The combination of HTTP/2 and HTTP/2 server push provides the _benefits_ of
 bundling (reduced latency) without actual bundling. Keeping resources separate
 means they can be cached efficiently and be shared between pages.
 
+HTTP/2 Push needs to be utilized with care, as it forces data to the browser,
+even if the file is already in the browser’s local cache or bandwidth is
+already saturated. If done wrong, performance can suffer. [Resource hints]
+might be a good alternative to allow the browser to make smart decisions about
+the prioritization of these requests.  
+
```
addyosmani: Can we call out link rel=preload? I think resource hints include enough possible things you could explore that it may be worth getting more concrete with which part we're specifically talking about

surma: Good call.

src/content/en/fundamentals/performance/prpl-pattern/index.md
graynorton: My sense is that there is a place to warn people about the dangers of PUSH, but that the PRPL doc is not necessarily that place. I'd rather see that discussion happen in some doc specific to H/2 PUSH, with just a brief aside in this PRPL doc, along the lines of "H/2 PUSH works well in the context of PRPL, but you should exercise caution when adopting PUSH in other contexts -- see [other doc] for more."

graynorton: Taking a closer look, I guess this edit is fine given the current form of the doc, but I'd like to work on getting the whole H/2 background section lifted out and put somewhere else...

igrigorik:
I'm late to the game here, but a nitpick to address in a different CL...  Resource Hints != Preload.

We should say "preload" in this doc, since that's what we're referring to. The difference is that preload is _not a hint_ but a directive to fetch the resource, whereas a hint (preconnect, prefetch, prerender) is a suggestion to the browser that it _should_ fetch that resource.

surma:
@igrigorik I’ll whip up another PR.

PR #3817, opened 2016-11-03
Update transferable objects post

ebidel:
R: @petele


petele:
Adding details about what was updated is a very nice to have, but not a requirement.

PR #3830, opened 2016-11-06
Fix Japanese Translation

sotayamashita:
It is a PR for Japanese translation. :jp: I fixed it because there were differences between ja and en. See original sentence below:

English:

```
The CSSOM and DOM are independent data structures!
```

Japanese:

```
CSSOM と CSSOM は独立したデータ構造です。
```

In Japanese sentence, 2nd "CSSOM" should be "DOM" so I fixed it.
Please review it. 

cc:/ @agektmr


agektmr:
LGTM

PR #3875, opened 2016-11-16
Swap last years CDS presentation for this years.

jpmedley:
@petele I plan to do more of this kind of thing. I stumbled on this and thought, it's (1) a critical issue and (2) the vids are a pretty apples-to-apples swap.


jpmedley:
PTAL

Didn't even think to look if it were anywhere else. I'll do that in the future.

PR #3886, opened 2016-11-22
add lighthouse manifest audits docs

kaycebasques:
Authored these a while ago but forgot to PR...

### DIFF
--- a/src/content/en/tools/lighthouse/audits/manifest-contains-192px-icon.md
+++ b/src/content/en/tools/lighthouse/audits/manifest-contains-192px-icon.md
@@ -1,0 +1,53 @@
+project_path: /web/_project.yaml
+book_path: /web/tools/_book.yaml
+description: Reference documentation for the "Manifest Contains Icons at Least 192px" Lighthouse audit.
+
+{# wf_updated_on: 2016-09-21 #}
+{# wf_published_on: 2016-09-21 #}
+
+# Manifest Contains Icons at Least 192px  {: .page-title }
+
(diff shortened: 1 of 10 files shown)


kaycebasques:
Actually I did PR this already but the content might have got lost in the migration...

https://github.com/google/WebFundamentals/pull/3439

FYI @petele 

We can abandon that old PR and use this one... I made some minor updates to the docs...

petele:
Hey @kaycebasques - Would it make sense to merge these files into larger blocks, for example, all of these would go under one file called `manifest.md`? I worry a little bit that the TOC will start to get really long as more tests are eventually added.

Will be happy either way, just a thought.

src/content/en/tools/lighthouse/audits/manifest-contains-192px-icon.md
```diff
@@ -0,0 +1,53 @@
...
+      ...
+      "icons": [{
+        "src": "images/homescreen192.png",
+        "sizes": "192x192",
+        "type": "image/png"
+      }],
+      ...
+    }
+
+Check out [Manifest Exists](manifest-exists#how)
```
petele: If you wanted, you could put this into an [include file](https://developers.google.com/web/resources/widgets#includes), so that you only have to write it once, and when/if it's ever updated, you can simply update the include.

kaycebasques:
@petele I've thought about the problem of the long ToC too... I think we're OK for now, but can revisit later... a single, ridiculously long page with all the references would probably work

PR #3887, opened 2016-11-23
add note about input latency probabilities

kaycebasques:
cc @brendankenny


jpmedley:
@kaycebasques Do you need a sign-off from Brendan? Otherwise, I'll OK this.

Issue #3585, opened 2016-10-03
Payments articles are not in the menu


PaulKinlan:
![screenshot from 2016-10-03 17-16-52](https://cloud.githubusercontent.com/assets/45510/19044692/4b3e5a7c-898d-11e6-8d61-944bc4e93877.png)

Note: For some reason these Articles are in the getting started guide, but reference from the Overview.

agektmr:
This is sorted. Closing.

PR #3889, opened 2016-11-28
Moved Payment Request stuff to "Discovery and Monetization"

agektmr:


agektmr:
As per #3585

src/content/en/fundamentals/discovery-and-monetization/_toc.yaml
```diff
@@ -9,4 +9,10 @@ toc:
       path: /web/fundamentals/discovery-and-monetization/social-discovery/
     - title: Monetization with Ads
       path: https://www.google.com/adsense
+    - title: Payment Request API
```
jpmedley: Let's call this 'Handling Payments'. The title shouldn't give the impression this is an API reference.

agektmr: Fixed.

PR #3892, opened 2016-11-28
update section title in lighthouse references

kaycebasques:
I got some feedback that the purpose of this section was confusing, so I updated the section title and note. I may need to change it again, so I'm using an include.

@petele the [includes](https://developers.google.com/web/resources/widgets#includes) references says that an include must be HTML, and that Markdown is not supported. However when I run the site locally the Markdown is working just fine. So, is that doc outdated? Can I use Markdown?

### DIFF
--- a/src/content/en/tools/lighthouse/audits/cache-contains-start_url.md
+++ b/src/content/en/tools/lighthouse/audits/cache-contains-start_url.md
@@ -14,12 +14,12 @@ homescreen while offline.
 
 ## How to pass the audit {: #how }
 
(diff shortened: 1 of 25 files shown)


kaycebasques:
Nevermind... after submitting the PR I can see that the CI is spitting out errors about the Markdown file missing tags, etc. It's probably possible to hack up the Markdown to include all of the necessary tags, but it's easier to just use HTML instead. So I switched to HTML.

kaycebasques:
@jpmedley unless anything stands out to you about this, this can be merged

PR #3893, opened 2016-11-29
clean up devtools shortcuts doc

kaycebasques:
'Twas in a nonsensical location in the IA (under the "Inspect CSS and HTML" section, when it should be top-level) and included redundant mouse info and irrelevant Chrome browser shortcuts.

### DIFF
--- a/src/content/en/tools/_redirects.yaml
+++ b/src/content/en/tools/_redirects.yaml
@@ -1,5 +1,8 @@
 redirects:
 
+- from: /web/tools/chrome-devtools/inspect-styles/shortcuts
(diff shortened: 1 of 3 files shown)


kaycebasques:
note: look into mentioning deleted mouse shortcuts

kaycebasques:
Yeah @jpmedley the deleted section was mouse shortcuts. It's nowhere near a comprehensive list of mouse shortcuts, and these shortcuts are covered in the relevant sections in the other docs. So I'm comfortable with deleting them. If anyone raises an issue, I can figure out a solution. But I doubt it'll come up.

PR #3894, opened 2016-11-29
rephrase rail guidelines

kaycebasques:
This started in an email thread with @tdresser. The RAIL doc was confusingly referring to a budget of 16ms for animations. This PR rephrases the stuff to make it clear that animation code should finish in 10ms, so that the browser has time to paint the new frame to the screen within the 16ms budget. 

PR also includes some minor copyedit updates

@tdresser please review

### DIFF
--- a/src/content/en/fundamentals/performance/rail.md
+++ b/src/content/en/fundamentals/performance/rail.md
@@ -9,15 +9,16 @@ description: RAIL is a user-centric performance model. Every web app has these f
 
 {% include "web/_shared/contributors/megginkearney.html" %}
 
-RAIL is a user-centric performance model. Every web app has these four distinct aspects to its life cycle, and performance fits into them in very different ways:
+RAIL is a user-centric performance model. Every web app has these four distinct aspects to its life cycle, and performance fits into them in different ways:
 
 ![RAIL performance model](images/rail.png)
 
 
 ### TL;DR {: .hide-from-toc }
-- Focus on the user; the end goal isn't to make your site perform fast on any specific device, it''s to ultimately make users happy.
+
+- Focus on the user; the end goal isn't to make your site perform fast on any specific device, it's to ultimately make users happy.
 - Respond to users immediately; acknowledge user input in under 100ms.
-- Render each frame in under 16ms and aim for consistency; users notice 'jank'.
+- When animating or scrolling, produce a frame in under 10ms.
 - Maximize main thread idle time.
 - Keep users engaged; deliver interactive content in under 1000ms.
 
@@ -36,11 +37,16 @@ Understand how users perceive performance delays:
   <tbody>
     <tr>
       <td data-th="Delay">0 - 16ms</td>
-      <td data-th="User Reaction">Given a screen that is updating 60 times per second, this window represents the time to get a single frame to the screen (Professor Math says "1000/60 ~= 16"). People are exceptionally good at tracking motion, and they dislike it when the expectation of motion isn’t met, either through variable frame rates or periodic halting.</td>
+      <td data-th="User Reaction">People are exceptionally good at tracking
+      motion, and they dislike it when animations aren't smooth. Users
+      perceive animations as smooth so long as 60 new frames are rendered
+      every second. That's 16ms per frame, including the time it takes for
+      the browser to paint the new frame to the screen, leaving an app
+      about 10ms to produce a frame.</td>
     </tr>
     <tr>
       <td data-th="Delay">0 - 100ms</td>
-      <td data-th="User Reaction">Respond to a user action within this time window and they will feel like the result was immediate. Any longer, and the connection between action and reaction is broken.</td>
+      <td data-th="User Reaction">Respond to a user action within this time window and users feel like the result is immediate. Any longer, and the connection between action and reaction is broken.</td>
     </tr>
     <tr>
       <td data-th="Delay">100 - 300ms</td>
@@ -64,8 +70,9 @@ Understand how users perceive performance delays:
 ## Response: respond in under 100ms
 
 You have 100ms to respond to user input before they notice a lag.
-This applies to any input, whether they are clicking a button,
-toggling form controls, or starting an animation.
+This applies to most inputs, such as clicking buttons, toggling form
+controls, or starting animations. This does not apply to touch drags or
+scrolls.
 
 If you don't respond, the connection between action and reaction is broken. Users will notice.
 
@@ -76,20 +83,25 @@ If possible, do work in the background.
 
 For actions that take longer than 500ms to complete, always provide feedback.
 
-Success: Respond to user's touchmoves and scrolling in under 16ms.
+## Animation: produce a frame in 10ms
 
-## Animation: render frames every 16ms
+Animations aren't just fancy UI effects. For example, scrolling and touch
+drags are types of animations.
 
-Animations aren't trivial actions that web apps can opt into.
-For example, scrolling and touchmoves are types of animation.
-Your users will really notice if the animation frame rate varies.
+Users notice when the animation frame rate varies.
 Your goal is to produce 60 frames per second, and every frame has to go through all of these steps:
 
 ![Steps to render a frame](images/render-frame.png)
 
-From a purely mathematical point of view, every frame has a budget of 16.66ms (divide 1 second by 60) but, because browsers have housekeeping to do, the reality is that there is a window of 10ms for your code during animations.
+From a purely mathematical point of view, every frame has a budget of about 
+16ms (1000ms / 60 frames per second = 16.66ms per frame). However, because
+browsers need some time to paint the new frame to the screen, **your code
+should finish executing in under 10ms**. 
 
-In high pressure points like animations, the key is to do nothing where you can, and where you can’t, do the absolute minimum. Whenever possible, make use of the 100ms response to pre-calculate expensive work so that you maximize your chances of hitting 60fps.
+In high pressure points like animations, the key is to do nothing where you
+can, and the absolute minimum where you can't. Whenever possible, make use of
+the 100ms response to pre-calculate expensive work so that you maximize your
+chances of hitting 60fps.
 
 For more information, see
 [Rendering Performance](/web/fundamentals/performance/rendering/).
@@ -101,15 +113,14 @@ Use idle time to complete deferred work. For example, keep pre-load data to a mi
 Deferred work should be grouped into blocks of about 50ms. Should a user begin interacting, then the highest priority is to respond to that. 
 
 To allow for <100ms response,
-the app must yield control back to main thread every <100ms,
+the app must yield control back to main thread every <50ms,
 such that it can execute its pixel pipeline, react to user input, and so on.
 
 Working in 50ms blocks allows the task to finish while still ensuring instant response.
 
 ## Load: deliver content under 1000ms
 
-Load your site in under 1 second.
-If you don't, your user's attention wanders,
+Load your site in under 1 second. If you don't, user attention wanders,
 and their perception of dealing with the task is broken.
 
 Focus on
@@ -118,9 +129,9 @@ to unblock rendering.
 
 You don't have to load everything in under 1 second to produce the perception of a complete load. Enable progressive rendering and do some work in the background. Defer non-essential loads to periods of idle time (see this [Website Performance Optimization Udacity course](https://www.udacity.com/course/website-performance-optimization--ud884) for more information).
 
-## Summary of key rail metrics
+## Summary of key RAIL metrics
 
-To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline tool](/web/tools/chrome-devtools/profile/evaluate-performance/timeline-tool) to record user actions. Then check the recording times in the Timeline against these key rail metrics:
+To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline tool](/web/tools/chrome-devtools/profile/evaluate-performance/timeline-tool) to record user actions. Then check the recording times in the Timeline against these key RAIL metrics:
 
 <table>
   <thead>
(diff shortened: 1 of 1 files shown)


src/content/en/fundamentals/performance/rail.md
```diff
@@ -36,11 +37,15 @@ Understand how users perceive performance delays:
   <tbody>
     <tr>
       <td data-th="Delay">0 - 16ms</td>
```
kaycebasques: Note that this section is talking about user perception, so I left the reference to 16ms here. However, I clarify below that this number includes browser work time.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -78,18 +83,24 @@ For actions that take longer than 500ms to complete, always provide feedback.
 
 Success: Respond to user's touchmoves and scrolling in under 16ms.
 
-## Animation: render frames every 16ms
+## Animation: execute animation code in under 10ms
```
kaycebasques: How's this phrasing?

src/content/en/fundamentals/performance/rail.md
```diff
@@ -132,21 +142,16 @@ To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline to
...
+      <td data-th="User Test">User taps a button (for example, opening navigation).</td>
     </tr>
     <tr>
       <td data-th="RAIL Step"><strong>Response</strong></td>
-      <td data-th="Key Metric">Input latency (from tap to paint) < 16ms.</td>
-      <td data-th="User Test">User drags their finger and the app's response is bound to the finger position (for example, pull to refresh, swiping a carousel).</td>
-    </tr>
-    <tr>
-      <td data-th="RAIL Step"><strong>Animation</strong></td>
       <td data-th="Key Metric">Input latency (from tap to paint) < 100ms for initial response.</td>
```
kaycebasques: I nixed the metric above because it's a confusing example. Is dragging a finger a response or an animation? Maybe we need to add this back, but I need help categorizing and phrasing this.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -132,21 +142,16 @@ To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline to
...
       <td data-th="RAIL Step"><strong>Response</strong></td>
       <td data-th="Key Metric">Input latency (from tap to paint) < 100ms.</td>
-      <td data-th="User Test">User taps on an icon or button (for example, opening the nav menu, tapping Compose).</td>
+      <td data-th="User Test">User taps a button (for example, opening navigation).</td>
     </tr>
     <tr>
       <td data-th="RAIL Step"><strong>Response</strong></td>
-      <td data-th="Key Metric">Input latency (from tap to paint) < 16ms.</td>
-      <td data-th="User Test">User drags their finger and the app's response is bound to the finger position (for example, pull to refresh, swiping a carousel).</td>
-    </tr>
```
kaycebasques: I nixed the metric above because it's a confusing example. Is dragging a finger a response or an animation? Maybe we need to add this back, but I need help categorizing and phrasing this.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -132,21 +142,16 @@ To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline to
...
-    <tr>
-      <td data-th="RAIL Step"><strong>Animation</strong></td>
       <td data-th="Key Metric">Input latency (from tap to paint) < 100ms for initial response.</td>
       <td data-th="User Test">User initiates page scroll or animation begins.</td>
     </tr>
```
kaycebasques: I changed the metric above from Animation to Response. I think the current published version is a mistake.

Here's the original message from @tdresser that prompted these changes:

The response row says from tap to paint (which should have 100ms as the threshold) on the left, but then talks about drag effects (which should have ~10ms as the threshold) on the right.
The animation row says from tap to paint (with the correct 100ms threshold) on the left, but then talks about scroll and animation start on the right, which are somewhat different.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -9,15 +9,16 @@ description: RAIL is a user-centric performance model. Every web app has these f
...
 
 
 ### TL;DR {: .hide-from-toc }
-- Focus on the user; the end goal isn't to make your site perform fast on any specific device, it''s to ultimately make users happy.
+
+- Focus on the user; the end goal isn't to make your site perform fast on any specific device, it's to ultimately make users happy.
 - Respond to users immediately; acknowledge user input in under 100ms.
-- Render each frame in under 16ms and aim for consistency; users notice 'jank'.
+- Execute animation code in under 10ms.
 - Maximize main thread idle time.
```
addyosmani: IIRC we've been telling developers to render a frame in 16ms, but to complete specific pieces of work in far less time than that (10ms). Is there a way for us to still highlight frame budget of 16.6ms but also say execution time for animation code should be 10ms?.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -108,8 +119,7 @@ Working in 50ms blocks allows the task to finish while still ensuring instant re
 
 ## Load: deliver content under 1000ms
 
-Load your site in under 1 second.
-If you don't, your user's attention wanders,
+Load your site in under 1 second. If you don't, user attention wanders,
```
addyosmani: Note for a future set of edits: the RAIL guidelines we published to /web didn't factor in 3G and mobile devices as much as they could and so the 1s guideline here is a little nebulous (today we focus more on being interactive in <5s and hitting a Speed Index of < 4000ms on 3G.

src/content/en/fundamentals/performance/rail.md
cwilso: And yes, this is Response, not Animation.

src/content/en/fundamentals/performance/rail.md
cwilso: Yeah, this is pretty clearly defined in the expanded version (above this in the page) - animation needs to be <16.6ms, but because of browser housekeeping, keep your budget <10ms.

"From a purely mathematical point of view, every frame has a budget of 16.66ms (divide 1 second by 60) but, because browsers have housekeeping to do, the reality is that there is a window of 10ms for your code during animations."

tdresser: "animation code" is a bit ambiguous. For example, the continuous part of a touch drag is animation. Perhaps something like:

"Produce each frame of an animation in under 10ms." or something like that?

src/content/en/fundamentals/performance/rail.md
```diff
@@ -36,11 +37,15 @@ Understand how users perceive performance delays:
   <tbody>
     <tr>
       <td data-th="Delay">0 - 16ms</td>
-      <td data-th="User Reaction">Given a screen that is updating 60 times per second, this window represents the time to get a single frame to the screen (Professor Math says "1000/60 ~= 16"). People are exceptionally good at tracking motion, and they dislike it when the expectation of motion isn’t met, either through variable frame rates or periodic halting.</td>
+      <td data-th="User Reaction">People are exceptionally good at tracking
+      motion, and they dislike it when animations aren't smooth. Users
+      perceive animations as smooth so long as 60 new frames are rendered
+      every second. That's 16ms per frame, including the time it takes for
+      the browser to paint the new frame to the screen.</td>
```
tdresser: Might be worth adding something along the lines of:

"including the time it takes for the browser to paint the new frame to the screen, leaving a web app about 10ms to produce a frame"

src/content/en/fundamentals/performance/rail.md
tdresser: I prefer the "produce a frame" phrasing, but that may be less understandable to developers. My concern with this is that folks will think "animation code" just means rAF callbacks.

src/content/en/fundamentals/performance/rail.md
tdresser: Once we've nailed down our definition of Time To Interactive, I agree it would be good to use TTI here. Until then, this seems good enough to me.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -78,18 +83,24 @@ For actions that take longer than 500ms to complete, always provide feedback.
```
tdresser: On line 72:
You have 100ms to respond to user input before they notice a lag.

This is a bit confusing, because it isn't true for touch drag, or scroll for example.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -78,18 +83,24 @@ For actions that take longer than 500ms to complete, always provide feedback.
 
 Success: Respond to user's touchmoves and scrolling in under 16ms.
```
tdresser: This should be in the animation section, shouldn't it?

src/content/en/fundamentals/performance/rail.md
```diff
@@ -78,18 +83,24 @@ For actions that take longer than 500ms to complete, always provide feedback.
...
 
-## Animation: render frames every 16ms
+## Animation: execute animation code in under 10ms
 
 Animations aren't trivial actions that web apps can opt into.
 For example, scrolling and touchmoves are types of animation.
```
tdresser: I don't understand the above. Animations sometimes are trivial things developers trigger (CSS animations for example). 

Should this be "Animations aren't just trivial..."?

src/content/en/fundamentals/performance/rail.md
```diff
@@ -108,8 +119,7 @@ Working in 50ms blocks allows the task to finish while still ensuring instant re
```
tdresser: This section is a bit confusing.

"To allow for <100ms response, the app must yield control back to **the** main thread every <100ms"
Didn't we just say 50ms?

We allow 100 ms of event processing time, and 50 ms of idle processing, which can result in a maximum event handling time of 150ms, if an event shows up right at the beginning of the idle period.

That's a bit subtle though...

src/content/en/fundamentals/performance/rail.md
tdresser: This is a bit hairy actually. Based on the folks I asked to try out a test page, most people actually find about 50ms the threshold for when touch drag or scroll start becomes too slow, which doesn't fit nicely into the RAIL model.

I'd be happy just removing the scroll start / drag start case here.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -132,21 +142,16 @@ To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline to
...
-    <tr>
-      <td data-th="RAIL Step"><strong>Animation</strong></td>
       <td data-th="Key Metric">Input latency (from tap to paint) < 100ms for initial response.</td>
       <td data-th="User Test">User initiates page scroll or animation begins.</td>
     </tr>
     <tr>
       <td data-th="RAIL Step"><strong>Animation</strong></td>
-      <td data-th="Key Metric">Each frame's work (JS to paint) completes < 16ms.</td>
+      <td data-th="Key Metric">Each frame's work (from JS to paint) completes < 16ms.</td>
       <td data-th="User Test">User scrolls the page or sees an animation.</td>
```
tdresser: Can we add touch drag as another example here (not the start of touch drag, but the continuous portion of the drag?)

src/content/en/fundamentals/performance/rail.md
```diff
@@ -132,21 +142,16 @@ To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline to
...
     </tr>
     <tr>
```
tdresser: On line 169 - Satisfy the Response goals during the full page load process.

That's pretty intense! Do we actually expect developers to be able to accomplish that? I don't think I've heard anyone claiming this.

src/content/en/fundamentals/performance/rail.md
kaycebasques: @tdresser opinion?

tdresser: I'd be fine with this as an animation, as long as folks won't get confused between the start of a touch drag, and the continuous phase of a touch drag.

The continuous phase of a touch drag is definitely animation.

petele:
@kaycebasques - when this is ready to go, remove the DO NOT MERGE tag

src/content/en/fundamentals/performance/rail.md
kaycebasques: ✔️

src/content/en/fundamentals/performance/rail.md
kaycebasques: :heavy

src/content/en/fundamentals/performance/rail.md
kaycebasques: I think it's backwards to emphasize the 16ms in the TLDR rather than the 10ms number. 16ms is a technicality. 10ms is the real target that we aim for developers to learn.

src/content/en/fundamentals/performance/rail.md
kaycebasques: ✔️

src/content/en/fundamentals/performance/rail.md
kaycebasques: ✔️

src/content/en/fundamentals/performance/rail.md
kaycebasques: ✔️ 

I think the `Success` header was mistakenly created when we migrated the site content to its latest infrastructure. Also, yes, this didn't make sense here, so I deleted it

src/content/en/fundamentals/performance/rail.md
kaycebasques: FYI I didn't write this article, just updating it. My guess here is something along the lines of "When you hear animations, you may think this only applies to fancy effects. But scrolling and touchmoves are animations"

kaycebasques:
@tdresser @addyosmani @cwilso I updated per last round of feedback

I'm deferring to Tim here and mostly lining up the doc to his guidance

src/content/en/fundamentals/performance/rail.md
```diff
@@ -145,16 +146,21 @@ To evaluate your site against RAIL metrics, use the Chrome DevTools [Timeline to
...
-      <td data-th="User Test">User initiates page scroll or animation begins.</td>
-    </tr>
-    <tr>
       <td data-th="RAIL Step"><strong>Animation</strong></td>
       <td data-th="Key Metric">Each frame's work (from JS to paint) completes < 16ms.</td>
       <td data-th="User Test">User scrolls the page or sees an animation.</td>
     </tr>
     <tr>
+      <td data-th="RAIL Step"><strong>Animation</strong></td>
+      <td data-th="Key Metric">Input latency (from tap to paint) < 16ms.</td>
```
tdresser: "from tap to paint" is incorrect. This should refer to scroll or drag, probably.

Or perhaps we can just merge the two Animation rows?

kaycebasques: I merged the two rows

kaycebasques:
@addyosmani & @cwilso LMK if this is OK... I'm blocked on your reviews

addyosmani:
LGTM!

PR #3919, opened 2016-12-05
Tweak to the SW primer to use the delayed SW registration pattern.

### LINKED ISSUE #3883: Guidance about deferred service worker registration
I'm planning on putting together an article explaining a pattern for deferring service worker registration until after the initial uncontrolled page has had a change to load, to make it less likely that any precaching requests compete for bandwidth with requests for critical assets needed to display the page.

There will be various bits of supporting text going into more detail, but the gist of it would be to recommend using

```js
if ('serviceWorker' in navigator) {
  window.addEventListener('load', function() {
    navigator.serviceWorker.register('/service-worker.js');
  });
}
```

as a sensible default, unless there's a specific need in your web app to register earlier on. (I'll cover those edge cases in the article.)

I've run this by a few folks from Chrome who seem to consider it reasonable, but was curious whether there were any concerns specific to Firefox's service worker implementation that might lead to this snippet being less-than-optimal. CC: @wanderview on that question.

Assuming we go ahead and publish the guidance, this issue can also track going through our backlog of service worker examples in our documentation, and updating them to follow that pattern when it makes sense to do so.

jeffposnick:
R: @petele @jpmedley 

Small update to @gauntface's SW primer to call out the pattern of waiting until the page has loaded to perform SW registration.

Fixes #3883

### DIFF
--- a/src/content/en/fundamentals/getting-started/primers/service-workers.md
+++ b/src/content/en/fundamentals/getting-started/primers/service-workers.md
@@ -126,18 +126,20 @@ To install a service worker you need to kick start the process by
 service worker JavaScript file lives.
 
     if ('serviceWorker' in navigator) {
-      navigator.serviceWorker.register('/sw.js').then(function(registration) {
-        // Registration was successful
-        console.log('ServiceWorker registration successful with scope: ', registration.scope);
-      }).catch(function(err) {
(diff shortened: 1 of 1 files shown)


gauntface:
LGTM from my side.

Issue #3920, opened 2016-12-05
Drop-off sentence in PWA codelab


Meggin:
In https://developers.google.com/web/fundamentals/getting-started/codelabs/your-first-pwapp/, this sentence, "We're going to create a Weather app as our first Progressive Web App. The key components will consist of:" stops without any explanation.

Then it's followed by a new paragraph.

It might make sense to actually just delete it, as we are further along in the doc anyway.

petele:
This needs to be updated in the Google Doc version, not the Git version. This is pulled directly from the Google doc version.

PR #3921, opened 2016-12-05
Update index.md

Meggin:
Fixes https://github.com/google/WebFundamentals/issues/3920

Removes drop-off sentence, that seems out of place in the current paragraph.


petele:
This will be reverted next time we push! Changes to codelabs need to be done in the original codelab doc.

PR #3922, opened 2016-12-05
Update promise-flow.svg to be colorblind friendly

c-stone:
The green/red arrow in the flow chart are completely indistinguishable for me, and others with some level of colorblindness. The purposed change is more colorblind friendly.


c-stone:
I signed it!

PR #3930, opened 2016-12-07
Eliminate false posatives on test for titles.

jpmedley:
Specifically, ^#[^#].* was catching CSS id selectors as well as legitimate markdown. We will probably need to revisit this.


jpmedley:
@petele It was either this or revert Sam's PR. We need to revisit this obviously.

gauntface:
Can you make sure you open an issue to track this change and why it was needed.

PR #3594, opened 2016-10-04
[lighthouse] move references to audits/ and add ToC

kaycebasques:
@petele this is good to go unless you or @jpmedley have any requests

PR #3955, opened 2016-12-12
Remove extra double quote from framebox iframe

petele:
Fixes #3594


petele:
Manually tested and confirmed.

PR #3961, opened 2016-12-14
[devtools] delete old images

kaycebasques:
To find unused images I ran:

```sh
#!/bin/bash

# Execute this from /src/content/en/tools/chrome-devtools

function search {
  # Provides just the filename of each image, e.g. `pause.png`
  for img in $(ls $1); do
    # Check if any of the files mention the image filename
    if ! grep -qR $img *; then
      rm $1$img
    fi
  done
}

# Recursively search for all directories named `imgs` or `images`
# All of the image directories use one of these two names
for dir in $(ls -d -R */*im[ag][gs]*/); do
  search $dir
done
```

Executed from `/src/content/en/tools/chrome-devtools/`.

I then ran a local server, and checked each page with DevTools open. DevTools didn't produce any 404s so I'm reasonably confident that the script worked as intended

Produces a 9M reduction in the directory size. I was hoping for more, but hey, every bit counts. Removing all those unused files should help reduce cognitive burden for docs maintainers, too


petele:
Good call - the only think I worry a little about is if someone is using one of those files outside of the `/tools/` directory. But for now, let's go with it and if we find we're in a bad state, we can fix.

kaycebasques:
Thanks @petele 

I thought about that too but I think the chances are slim

PR #3964, opened 2016-12-14
Added bokan to _contributors.yaml

bokand:


src/data/_contributors.yaml
```diff
@@ -1251,3 +1251,16 @@ maciejzasada:
     - author
   google: "104812516714503133849"
   email: maciej@unit9.com
+  
+bokan:
```
petele: Entries should be added to the list alphabetically by last name.

bokand: Done, ptal

PR #3963, opened 2016-12-14
Add 'URL Bar Resizing' article.

jpmedley:

PR #3981, opened 2016-12-19
Add url bar sizing article

bokand:
I can't push to the branch that https://github.com/google/WebFundamentals/pull/3963 is based on and Joe is out so creating a new PR.


bokand:
Bah, sorry, my branch is in a bad state. I'll try again.

PR #3996, opened 2016-12-29
Fixed nit in notifications sample

beaufortfrancois:
Since Chrome 55, `startNotifications` returns `characteristic`. This patch updates notifications sample to reflect this change.

R: @petele


beaufortfrancois:
Thank you @jpmedley!

Issue #4017, opened 2017-01-09
Need test to prevent .js files from being checked into /content directory


petele:

jpmedley:
From discussion w/ Pete: account for is .js files in _code directories, those will require special handling as they are "kind of" allowed.

jpmedley:
Plan: use findBadMDExtensions() as a model. src/content/en/events/embeds/event-map.js is allowed.

PR #4018, opened 2017-01-10
Add test for js files.

jpmedley:
https://github.com/google/WebFundamentals/issues/4017

### DIFF
--- a/gulp-tasks/test.js
+++ b/gulp-tasks/test.js
@@ -242,20 +242,23 @@ function findBadMDExtensions() {
 }
 
 /**
- * Chec to ensure there an o JavaSCript files in the content/ folder.
+ * Checks to ensure there are no JavaSCript files in the content/ folder.
  *
  * Note: the promise will always be fulfilled, even if it fails.
  *
- * @return {promise} An empty promsie
+ * @return {Promise} An empty promsie
  */
- function findJavaSCriptFiles() {
+ function findJavaScriptFiles() {
   return new Promise(function(resolve, reject) {
     var contentPath = GLOBAL.WF.src.content;
+    if (GLOBAL.WF.options.testPath) {
+      contentPath = GLOBAL.WF.options.testPath;
+    }
     var patterns = ['**/*.js', '!**/_code/*.js', '!**/event-map.js'];
     var files = wfHelper.getFileList(contentPath, patterns);
     files.forEach(function(filename) {
       summary.files++;
-      logError(filename.filePath, '\n\tJS files must be in a directory named /_code.')
+      logError(filename.filePath, '\n\tJS files must be in a directory named _code/.')
     });
     resolve();
   });
@@ -541,7 +544,7 @@ gulp.task('test', function(callback) {
         testAllYaml(),
         testAllMarkdown(),
         findBadMDExtensions(),
-        findJavaSCriptFiles()
+        findJavaScriptFiles()
       ]);
     })
     .catch(function(err) {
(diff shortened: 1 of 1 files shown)


gulp-tasks/test.js
```diff
@@ -242,6 +242,26 @@ function findBadMDExtensions() {
 }
 
 /**
+ * Chec to ensure there an o JavaSCript files in the content/ folder.
```
petele: Please fix the description here, missing a few letters ;)

gulp-tasks/test.js
```diff
@@ -242,6 +242,26 @@ function findBadMDExtensions() {
...
+ *
+ * Note: the promise will always be fulfilled, even if it fails.
+ *
+ * @return {promise} An empty promsie
```
petele: Promise in `{}` should be capitalized

gulp-tasks/test.js
```diff
@@ -242,6 +242,26 @@ function findBadMDExtensions() {
...
+ */
+ function findJavaSCriptFiles() {
```
petele: Change to `findJavaScriptFiles()`

gulp-tasks/test.js
```diff
@@ -242,6 +242,26 @@ function findBadMDExtensions() {
...
+  return new Promise(function(resolve, reject) {
+    var contentPath = GLOBAL.WF.src.content;
```
petele: Need to add support for `testPath`, which is used when only doing a single language.

You can probably do something like:
```
var contentPath = GLOBAL.WF.src.content;
if (GLOBAL.WF.options.testPath) {
  contentPath = GLOBAL.WF.options.testPath;
}

File: gulp-tasks/test.js [hunk 5]
```diff
@@ -242,6 +242,26 @@ function findBadMDExtensions() {
...
+    var patterns = ['**/*.js', '!**/_code/*.js', '!**/event-map.js'];
+    var files = wfHelper.getFileList(contentPath, patterns);
+    files.forEach(function(filename) {
+      summary.files++;
+      logError(filename.filePath, '\n\tJS files must be in a directory named /_code.')
```
petele: Be consistent in logging, don't use the `\n\t`, instead I'd recommend: 'JavaScript files are not permitted.'

gulp-tasks/test.js
```diff
@@ -520,7 +540,8 @@ gulp.task('test', function(callback) {
       return Promise.all([
         testAllYaml(),
         testAllMarkdown(),
-        findBadMDExtensions()
+        findBadMDExtensions(),
+        findJavaSCriptFiles()
```
petele: Be sure to update once you've renamed the function.

gulp-tasks/test.js
jpmedley: Am I misunderstanding what `_code`/ is for?

The `\n\t` helps the readability of the output a bit. Can we keep it?

jpmedley:
@petele 
Your request for line 258 isn't finished because I have two questions.

Am I misunderstanding what _code/ is for?

The \n\t helps the readability of the output a bit. Can we keep it?

gulp-tasks/test.js
petele: I'd prefer to drop it for consistency right now as we're not using it anywhere else. 

As for the `_code/` directory, all JS files checked into the Google Source Repo need to be reviewed by someone with JS readability, which makes that CL harder to get approved. I'd rather discourage people from doing it unless it's a special case.

petele:
I'd prefer to drop it for consistency right now as we're not using it anywhere else. 

As for the `_code/` directory, all JS files checked into the Google Source Repo need to be reviewed by someone with JS readability, which makes that CL harder to get approved. I'd rather discourage people from doing it unless it's a special case.

jpmedley:
PTAL

Issue #4041, opened 2017-01-13
local server does not support `setvar`


kaycebasques:
add following to a markdown page:

```
{% setvar name "kayce" %}
hi my name is {{ name }}
```

expected:

```
hi my name is kayce
```

actual:

(Same as markdown)

PR #4042, opened 2017-01-13
[devtools] add reusable feedback widget

kaycebasques:
@petele check this out

My goal is to create a reusable feedback widget that I can inline at the end of DevTools tutorial sections. This will give me much more granular information about where exactly people are failing.

I've put the widget inside of a framebox because I want to disable the buttons after the user clicks one of them, as a means of preventing them from clicking both buttons and distorting the data.

I made it pretty customizable, because I think it'll get boring if the success and response items say the same thing every time. Switching up the button / response texts should help keep the user engaged and participating.

When I run this on the local server, the `{% setvar %}` tags aren't working properly, but I think I'm doing it correctly. See #4041 

Ditto for the framebox. I think I'm doing that one correctly, too, though. I think the local server just isn't set up to support these features

Given this, I'm not sure if the widget is going to display correctly. If this pattern looks OK to you, we can just check it when it's on staging

Note: I also moved some common styles to another file and removed some comments. These aren't related to the feedback widget


nayphyoaung:
[@ * @] reusable feedback widget

src/content/en/tools/chrome-devtools/_shared/_feedback.html
```diff
@@ -0,0 +1,41 @@
+{% framebox width="auto" height="auto" %}
```
petele: In order for widgets to work inside a framebox, the first line needs to be:
`{% framebox width="auto" height="auto" enable_widgets="true" %}`

petele: Only the directory needs to start with `_`, the filename within there doesn't.

kaycebasques: Fixed

kaycebasques: Fixed

kaycebasques:
@petele awesome, thanks for staging and reviewing

Notes:

I updated the styles so that the buttons display next to each other:

screen shot 2017-01-19 at 12 56 16 pm

The UX is quite buggy... When I click **SOMETHING'S WRONG** the response text works as expected, but the **SOMETHING'S WRONG** button disappears. When I click on the link to report the issue, GitHub does not open in a new page as expected. I just stay on the doc page.

The "Feedback" text prefacing the question looks a little weird to me. I wonder if we should style feedback in a different background color than the rest of the docs? We might be able to just wrap it all in a `callout` and style it as a "note"...

src/content/en/resources/widgets.md
```diff
@@ -189,3 +189,48 @@ automatically maintain a 16:9 ratio.
...
+
+## Inline Feedback {: #inline-feedback }
+
+Gain more feedback on your doc by asking your readers yes / no questions.
+
+### Example
+
+See question at bottom of [Get Started Debugging JS][inline feedback example].
+
+[inline feedback example]: https://developers.google.com/web/tools/chrome-devtools/javascript/#step-1
```
petele: Don't use FQDN for d.g.c, use only the absolute URL

kaycebasques:
Thanks for reviews @petele!

The future is 🔜

PR #4049, opened 2017-01-16
Payment request update

agektmr:
Adds various updates to the Payment doc.
Can you please do a final review and approve? @rsolomakhin
One change I might be missing is if I should note that you can't ignore `shippingoptionchange` event. Can you confirm?


rsolomakhin:
lgtm

PR #4062, opened 2017-01-20
Make the podcast audio player bigger

### LINKED ISSUE #4064: Make the podcast audio player bigger
I was listening to the HTTP 203 podcast and I struggled to control the audio because the player was too small.

small

big

murielsilveira:
I was listening to the HTTP 203 podcast and I struggled to control the audio because the player was too small. So I make it bigger :)

small

big

Let me know if I put the css on the wrong file, I couldn't find a better place. :)

Closes #4064

### DIFF
--- a/gae/article.tpl
+++ b/gae/article.tpl
@@ -22,6 +22,7 @@
         box-shadow: 0 1px 4px rgba(0,0,0,.37);
       }
       .devsite-rating-stars { text-align:right; }
+      .devsite-podcast-audio { width: 100%; }
     </style>
     <title>{{ title }}</title>
   </head>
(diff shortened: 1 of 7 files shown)


gae/article.tpl
```diff
@@ -22,6 +22,7 @@
         box-shadow: 0 1px 4px rgba(0,0,0,.37);
       }
       .devsite-rating-stars { text-align:right; }
+      .devsite-podcast-audio { width: 100%; }
```
murielsilveira: I also followed the same `devsite-rating-stars` naming

petele:
Thanks for the commit. The CSS that's used on https://developers.google.com isn't easily editable by us, since it's used across DevSite. The CSS you added will only work on our staging environment. 

Unfortunately, the only way to do this would be to add it to every file, and that's obviously not ideal. I'm going to close this for the time being, but would you file an issue that says the audio should be full width and we'll see if we can get it fixed in DevSite?

Thanks!

murielsilveira:
@petele Got it.
I already opened an issue for that: #4064 :)

petele:
Awesome, thanks!!

Issue #1518, opened 2015-08-12
[EN] Feedback for: /tools/polymer-starter-kit/index.html


mharris7190:
All buttons seem to have a strange double tap behavior (sometimes the tap won't even register) on mobile devices (touch events supposedly). Is there a workaround or bugfix for this in the works?

petele:
@gauntface - can you take a quick peek and determine if there's work necessary here?

gauntface:
@mharris7190 The only button I see on that page is the one for navigation and download polymer starter kit.

Could you give reproducible steps + browser you are seeing it on.

mharris7190:
@gauntface 
After installing the polymerStarterKit and starting it up with 
~$ yo polymer
....
~$ gulp serve
I see this weird double tap thing when visiting the page (ip address and port) from a mobile device.

gauntface:
cc @robdodson 

This is an issue with Polymer Starter Kit not the docs (sorry), can you reopen the bug on this repo: https://github.com/PolymerElements/polymer-starter-kit

Just mention everything you said in the previous report.

It sounds like a Webkit issue.

PR #4074, opened 2017-01-23
[lighthouse] add docs for a11y audits

kaycebasques:
@robdodson here's the rest of the a11y docs... plz review


ebidel:
@kaycebasques you're crushing these!

kaycebasques:
@ebidel 💪

petele:
These are likely false positives on test failures, will investigate today.

src/content/en/tools/lighthouse/audits/_shared/query.md
```diff
@@ -0,0 +1,15 @@
+Below the audit, Lighthouse displays something like `1 element fails
```
petele: This fails because it's an `.md` file, and doesn't have the standard headers. In this case, it's OK, but need to verify a few things and update the test suite to avoid these failures in the future.

src/content/en/tools/lighthouse/audits/contrast-ratio.md
```diff
@@ -0,0 +1,57 @@
...
+         element is being hovered over in order to view its issues.
+       </figcaption>
+     </figure>
+
+1. Adjust the foreground or background color of the failing elements so that
+   they have a contrast ratio of at least 4.5 to 1, which is the minimum
+   recommended ratio. See [contrast ratio][CR] to calculate
+   ratios.
+
+[CL]: https://chrome.google.com/webstore/detail/chromelens/idikgljglpfilbhaboonnpnnincjhjkd?hl=en
```
petele: Please remove the `?hl=en` from the URL, we don't want to force people into a specific language.

src/content/en/tools/lighthouse/audits/_shared/query.md
```diff
@@ -0,0 +1,15 @@
...
+this test`. The number varies depending on how many elements are failing.
+Click this label to expand the list. You can find each element in your
+DOM by running the [`$()`][qs], [`$$()`][qsa], and [`$x()`][xp] functions
+from the Chrome DevTools Console.
+
+You can also use [ChromeLens][ChromeLens] to find the elements. See
```
robdodson: I think this might work for the contrast example but I'm not sure if it works for every test we're doing. Does ChromeLens let you query for a specific element or does it just highlight elements that fail its own tests?

petele:
@kaycebasques - please add `{# wf_md_include #}` to the top of `query.md`. That will prevent many of the tests from failing (#4076).

src/content/en/tools/lighthouse/audits/contrast-ratio.md
```diff
@@ -0,0 +1,57 @@
...
+Some users have low-contrast vision. They don't see many areas that are very
+bright or very dark. Everything looks about the same brightness. If the text
+doesn't have a high contrast ratio, these users literally have trouble seeing
+it at all.
+
+## How to pass the audit {: #how }
+
+To find and fix each of element that does not have a sufficient contrast ratio:
+
+1. Install [ChromeLens][CL].
```
robdodson: @alice just curious if you think it's better to use ChromeLens or would you prefer folks use the Chrome Accessibility DevTools extension. Long term, is this functionality that you think will be brought into Chrome DevTools itself?

src/content/en/tools/lighthouse/audits/form-labels.md
```diff
@@ -0,0 +1,44 @@
...
+
+    <label>First Name <input type="text"/></label>
+
+Explicit labels:
+
+    <label for="first">First Name <input type="text" id="first"/></label>
+
+`aria-label`:
+
+    <label>First Name <input type="text" aria-label="First Name"/></label>
```
robdodson: You wouldn't need to use aria-label here. `<label>` only supports a specific subset of elements like input, select and a few others. If you make a custom form element, like `<custom-dropdown>`, it will not be supported by `<label>`, so you'd need to use aria-label on it.

src/content/en/tools/lighthouse/audits/form-labels.md
```diff
@@ -0,0 +1,44 @@
...
+
+`aria-labelledby`:
+
+    <label id="name">First Name</label>
```
robdodson: Similar to above, you would not need to use aria-labelledby here. aria-labelledby is mainly for custom controls which label doesn't already support.

```
<span id="foo">Select seat:</span>
<custom-dropdown aria-labelledby="foo">...</custom-dropdown>
```

src/content/en/tools/lighthouse/audits/required-aria-attributes.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. Check the `aria-*` attributes of the element against the **Required
+   States and Properties** list.
+
+1. Add any attributes that are missing.
+
+[qs]: /web/tools/chrome-devtools/console/command-line-reference#queryselector
+[qsa]: /web/tools/chrome-devtools/console/command-line-reference#queryselectorall
+[xp]: /web/tools/chrome-devtools/console/command-line-reference#xpath
+[roles]: https://www.w3.org/TR/wai-aria/roles#role_definitions
```
robdodson: I think we may want to update all ARIA docs to point to the new 1.1 version
https://www.w3.org/TR/wai-aria-1.1/#roles

src/content/en/tools/lighthouse/audits/required-aria-attributes.md
```diff
@@ -0,0 +1,43 @@
...
+
+{# wf_updated_on: 2017-01-17 #}
+{# wf_published_on: 2017-01-17 #}
+
+# Elements With ARIA Roles Have Required Attributes  {: .page-title }
+
+## Why the audit is important {: #why }
+
+The `role` attribute helps screen readers understand the purpose of an ARIA
+widget. Each `role` has a set of required, additional attributes that
```
robdodson: I think many roles do not have any required attributes so suggest rephrasing to "A `role` may also have a set of required, additional attributes".

src/content/en/tools/lighthouse/audits/tabindex.md
```diff
@@ -0,0 +1,32 @@
...
+
+See [Using tabindex](/web/fundamentals/accessibility/focus/using-tabindex)
+for more information.
+
+## How to pass the audit {: #how }
+
+<<_shared/query.md>>
+
+Set the `tabindex` of each of these elements to either `-1`, for elements
+that should not be keyboard navigable, or `0`, for elements that should.
```
robdodson: I'd suggest also adding, "if you need an element to appear earlier in the tab order, consider moving it earlier in the DOM."

src/content/en/tools/lighthouse/audits/_shared/query.md
kaycebasques: It just highlights elements that fail the tests. But you're right, ChromeLens only covers vision tests. It doesn't make sense to include this in all the a11y docs

src/content/en/tools/lighthouse/audits/form-labels.md
kaycebasques: Can you give me an example to use instead?

src/content/en/tools/lighthouse/audits/form-labels.md
kaycebasques: Can you give me an example to use instead?

src/content/en/tools/lighthouse/audits/required-aria-attributes.md
kaycebasques: Does the new version have a simple list of roles like the one I provided, [Definitions of Roles](https://www.w3.org/TR/wai-aria/roles#role_definitions)?

The whole idea here is to find the role in question from that simple list, and then 

I think the bigger win here would be for Lighthouse to provide this info to the user, see [#1518](https://github.com/GoogleChrome/lighthouse/issues/1518)

src/content/en/tools/lighthouse/audits/contrast-ratio.md
```diff
@@ -0,0 +1,57 @@
...
+description: Reference documentation for the "Background and Foreground Colors Have Sufficient Contrast Ratio" Lighthouse audit.
+
+{# wf_updated_on: 2017-01-20 #}
+{# wf_published_on: 2017-01-20 #}
+
+# Background and Foreground Colors Have Sufficient Contrast Ratio  {: .page-title }
+
+## Why the audit is important {: #why }
+
+Some users have low-contrast vision. They don't see many areas that are very
```
kaycebasques: Fixed

src/content/en/tools/lighthouse/audits/required-aria-attributes.md
kaycebasques: Fixed

src/content/en/tools/lighthouse/audits/tabindex.md
kaycebasques: Fixed

kaycebasques:
Thanks for review, @robdodson. Plz see my replies to your comments, I have a couple questions.

I'll wait on @alice's response too, as that may have a biggish impact on the docs

petele:
@kaycebasques - whenever this is ready, remove the DO NOT MERGE tag, then merge and let me know it's merged so I can push it live.

src/content/en/tools/lighthouse/audits/form-labels.md
robdodson:
```
<button class="hamburger-menu" aria-label="menu">...</button>
```

src/content/en/tools/lighthouse/audits/required-aria-attributes.md
robdodson: yep https://www.w3.org/TR/wai-aria-1.1/#h-role_definitions. Also there are new roles in aria 1.1 that aren't in aria 1.0.

src/content/en/tools/lighthouse/audits/contrast-ratio.md
alice: Could they use the lighthouse extension instead?

kaycebasques: @alice what I'm trying to solve here is to give users a relatively easy path for finding the elements that are failing the test. The Lighthouse extension just gives you a list of selectors like `.nav > .button`. I already mention that you can use the DevTools Console to query for the elements, but that can get kind of tedious. So I was just providing another path for finding the elements, hence ChromeLens, which highlights the failing elements in your viewport...

PR #4077, opened 2017-01-24
Lighthouse: fix stray attempt-right

ebidel:
R: all

Was not rendering properly by the server.


petele:
Fixed in #4075 already - just haven't published it yet.

PR #4080, opened 2017-01-25
[devtools] polish the UI reference

kaycebasques:
This addresses some feedback I got about this doc, which is really just a reference on some common UI elements. The updated title and intro should clarify the purpose of the doc. I also polished up the sections and moved some sections to more appropriate places in other docs.


src/content/en/tools/chrome-devtools/ui.md
```diff
@@ -0,0 +1,166 @@
+project_path: /web/_project.yaml
+book_path: /web/tools/_book.yaml
+description: A reference on how to access and use common parts of the DevTools UI, and change the UI's appearance.
+
+{# wf_updated_on: 2017-01-24 #}
+{# wf_published_on: 2016-03-28 #}
```
petele: Don't forget to update the published on, looks like this is a new file.

PR #4081, opened 2017-01-25
Dev Feedback on Home Page

petele:
Here's what I've got for the landing page.

screen shot 2017-01-25 at 1 17 41 pm

### DIFF
--- a/src/content/en/_index.yaml
+++ b/src/content/en/_index.yaml
@@ -99,6 +99,20 @@ landing_page:
       path: https://www.ampproject.org/docs/get_started/create
 
 
+  - items:
+    - heading: Developer Feedback
+      description: >
+        <p>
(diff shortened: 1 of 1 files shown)


robnyman:
Looks good to me!
Out of interest, do you plan to have something to the right of it? Since I guess it will be 2 two-col layouts, one on top of the other.

petele:
No, was going to leave the right empty. When it goes full width, it doesn't look right. And we may add something there in the future.

robnyman:
Ah, got it. Thanks!

vikaskyadav:
Hey,

Not a big issue I guess but why so much emphasis on the word "downasaur"?

petele:
@drufball - any comments before I merge?

PR #4088, opened 2017-01-27
update with original update  

alphayang:
@petele any advice on image resource location change sync between EN and CN? The problem should be the same in other translations.


petele:
@alphayang - If the English ones work, I'd suggest leaving them so updating is easier, but you can also submit updates to images if necessary.

PR #4093, opened 2017-01-27
Extend the tags system for /updates/ to work for /showcase/

wibblymat:
We would like the ability to link to WebVR specific showcases from the WebVR section of fundamentals.

To that end I have extended the `tags` feature of `/updates/` so that it will work for `/showcase/` too.

Actual showcases with the `webvr` tag are still incoming, but for testing you can add `{# wf_tags: webvr #}` (or whatever tags you like) to random case studies and rebuild.

R: @petele @PaulKinlan 

@meganlindsay: FYI. The resulting pages will be just like https://developers.google.com/web/updates/tags/chrome55 only they will say "Showcases" instead of "Updates" at the top.


wibblymat:
@petele: Done, PTAL.

src/templates/tags/tag-index.md
```diff
@@ -7,5 +7,5 @@ book_path: /web/{{section}}/_book.yaml
 # {{ title }} {: .page-title }
 
 {{#each tags}}
-* [{{this}}](/web/updates/tags/{{this}})
+* [{{this}}](/web/{{section}}/tags/{{this}})
```
wibblymat: @petele actually, this doesn't work because it seems like `section` isn't defined inside the loop, even though it is defined at the top of the template. Any idea what syntax to use here?

wibblymat: Never mind, figured it out!

petele:
@wibblymat - looks great, I added tags to all of the files. Just waiting for it to finish the build then I'll merge.

PR #4094, opened 2017-01-27
Move deprecation policy to canned widget

petele:


petele:
@jpmedley - here's that canned deprecation policy file that can be included in all future deprecation updates. Once this is merged, we can start using it for all future posts.

PR #4108, opened 2017-01-31
Translated the overview of Engage-and-Retain in Korean

joshua1988:
It's been a while Pete.


joshua1988:
@cwdoh, thanks for the review! I will update the header :)

joshua1988:
@petele - pete, could you check this build failure? I am not sure what is happening here.

petele:
Not sure what's going on with the build failure, but it didn't fail on the tests, so we're good.

PR #4114, opened 2017-01-31
Push codelab updates

gauntface:
Couple of users were running into devtools truncating the subscription in the console so I've removed the subscription printing to force developers to get the subscription from the web page.


petele:
I'm closing this, but will replace it with the autogenerated one using `tools/update-codelabs.sh`

petele:
Fixed with #4131

PR #4134, opened 2017-02-03
Make resource loading deprecated

petele:


kaycebasques:
@petele is this how you want me to handle deprecated devtools pages?

petele:
I added this back and added `status: deprecated` to it because otherwise it looks icky without the index on the left.

I'd suggest when something is deprecated, add `status: deprecated` to the toc, then once views drop on the page to near zero, we delete the page.

PR #4166, opened 2017-02-09
Fix left nav encoding

beaufortfrancois:
R:@petele
BEFORE: before
AFTER: after2


petele:
I'm going to close as this would make our rendering behave differently than production DevSite. 

From [DevSite](https://developers.google.com/web/updates/2011/11/Stream-video-using-the-MediaSource-API):
screen shot 2017-02-09 at 9 54 04 am

PR #4147, opened 2017-02-07
Navigation preload article

jakearchibald:
Ready for review!

PR #4185, opened 2017-02-14
Omit self in SW

beaufortfrancois:
Following https://github.com/google/WebFundamentals/pull/4147, I've omitted `self` in Service Worker code snippets.

### DIFF
--- a/src/content/en/updates/2017/02/media-session.md
+++ b/src/content/en/updates/2017/02/media-session.md
@@ -2,7 +2,7 @@ project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: Customize web media notifications and respond to media related events with the new Media Session API.
 
-{# wf_updated_on: 2017-02-06 #}
+{# wf_updated_on: 2017-02-14 #}
 {# wf_published_on: 2017-02-06 #}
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2017/02/media-session.md
```diff
@@ -307,7 +307,7 @@ browser can't fetch them. Here's how you could implement this:
 
     const FALLBACK_ARTWORK_URL = 'fallbackArtwork.png';
     
-    self.addEventListener('install', event => {
+    addEventListener('install', event => {
       self.skipWaiting();
```
jpmedley: Should 'self' go away here too?

beaufortfrancois: I think it's best to omit only for top-level calls.

jpmedley: As an editor, I'll take your word for it. As a student of web development, I'd like to know why?

beaufortfrancois: > I guess it makes sense to use "self" when you want to be explicit about using the global, but that doesn't seem necessary if you're writing code at the top level anyway

PR #4186, opened 2017-02-14
Add a missed deprecation.

jpmedley:


jpmedley:
@petele The extra moving of content in this fixes the mistake in alphabetizing of the heading.

PR #4193, opened 2017-02-16
Fix IntersectionObserver sample code

RockoDev:


petele:
@Surma, would you take a look at this please? it's not clear to me if these values are supposed to be part of an object or a function.

surma:
The fix is correct, LGTM! Thanks @RockoDev :)

PR #4204, opened 2017-02-18
Update location of Lighthouse "Open in Viewer"

dandv:


petele:
@dandv - would you mind fixing the build errors please?
https://travis-ci.org/google/WebFundamentals/builds/202843542

dandv:
["always leave the code behind in a better state than you found it"](https://martinfowler.com/bliki/OpportunisticRefactoring.html)

PR #4220, opened 2017-02-22
[devtools] update inline feedback

kaycebasques:
* Removes hardcoding of `url` from the main feedback logic.
* Makes `question` optional. If omitted, then it's assumed that all of needed context will be provided in the button text. See `web/tools/chrome-devtools/javascript` on the staging site for an example.
* Makes the fail response more flexible.
* Prefixes the labels with numbers, so that I can keep track of the ordering of questions in Google Analytics
* Updates doc

### DIFF
--- a/src/content/en/_shared/feedback.html
+++ b/src/content/en/_shared/feedback.html
@@ -3,13 +3,14 @@
   usage.
 {% endcomment %}
 {% framebox width="auto" height="auto" enable_widgets="true" %}
-{% setvar url %}https://github.com/google/webfundamentals/issues/new?title=[{{ category }} / {{ label }}] {% endsetvar %}
 <style>
   .wf-feedback--hidden {
     display: none;
   }
 </style>
+{% if question %}
 <p>{{ question }}<p>
+{% endif %}
 <button type="button" class="button button-primary gc-analytics-event"
     data-category="{{ category }}" data-label="{{ label }}" data-value="1">
   {{ success_button }}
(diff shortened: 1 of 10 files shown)


src/content/en/_shared/feedback.html
```diff
@@ -3,13 +3,14 @@
   usage.
 {% endcomment %}
 {% framebox width="auto" height="auto" enable_widgets="true" %}
-{% setvar url %}https://github.com/google/webfundamentals/issues/new?title=[{{ category }} / {{ label }}] {% endsetvar %}
```
petele: Instead of removing this, and having to define in in all use cases, why not use something like this:
```
{% if url %}
{% setvar feedbackURL %}{{url}}{% endsetvar %}
{% else %}
{% setvar feedbackURL %}https://github.com/google/webfundamentals/issues/new?title=[{{ category }} / {{ label }}]{% endsetvar %}
{% endif %}

{{feedbackURL}}
```

If the URL is defined, it uses that, but otherwise it uses the fallback URL.

kaycebasques: I like the direction, but I think it'll cause "compile" errors. A feedback file like `/web/tools/chrome-devtools/javascript/_feedback/1.html` uses `feedbackURL` like so:

    {% setvar fail_response %}... please <a href="{{feedbackURL}}">file an issue</a>{% endsetvar %}

And then `1.html` includes `/web/_shared/feedback.html`. Since `feedbackURL` is used before it's declared, Django is going to throw an error.

kaycebasques:
@petele thanks for the review, I think the design I've laid down is the only feasible option right now (see my reply to your comment), so I'm happy with this design for now

Issue #4251, opened 2017-02-24
Your first PWA: Terms need defining


garanj:
https://developers.google.com/web/fundamentals/getting-started/codelabs/your-first-pwapp/

Based on review from Aleks:

App-like - Feels like an app to the user with app-style interactions and navigation because it's built on the app shell model.

==> Explain "app shell model" before using it. (This is the first page I am reading, and I don't know anything about progressive web apps; that's why I am here. Should I know what app shell model is? Is that a prerequisite?)

Discoverable - Is identifiable as an "application" thanks to W3C manifest and service worker registration scope, allowing search engines to find it.

Am I supposed to know what "W3C manifest" and "service worker registration" are? Is there a prerequisite that's not stated? otherwise, please define or addd link to definition.

Why is "keep" in quotes? 

Easily share via URL...
Share what? the page? the app? 

the key principles ==> the above key principles? or are there other ones not yet stated? 

considerations, as well as implementation details to ensure 
=> remove comma or add another one after details

What are we going to be building? ==> What you will build

Gray box after this title is garbled and looks like it's interpreted as a pre element perhaps? 
Content repeats previous definitions. 

How to design and construct an app using the "app shell" method
==> Please define "app shell method" 

How to store data for use offline later ==> later offline use

Chrome 52 or above => Browser? Please link to download. 

The sample code ==> link to code

Basic knowledge of HTML, CSS, JavaScript, and Chrome DevTools => State prereqs at beginning of codelab; link to Chrome DevTools or docs about them. 

This codelab is focused on Progressive Web Apps. Non-relevant concepts and code blocks are glossed over and are provided for you to simply copy and paste. ==> delete

Click the following button => not button; it's a link

More help: [Add and open Chrome apps](https://support.google.com/chrome_webstore/answer/3060053?hl=en) << raw Markup

Delete "obviously" 

What is the app shell? -- YAY!

Define service worker and add link: https://developers.google.com/web/fundamentals/getting-started/primers/service-workers
A service worker is a script that your browser runs in the background, separate from a web page, opening the door to features that don't need a web page or user interaction.

kosamari:
I believe the issue is adressed & fixed. closing.

PR #4252, opened 2017-02-24
Fix issues with Your First PWA identified in #4251

garanj:
Tried to address the issues identified in #4251.

The one item that I have not changed is the prerequisites:

"State prereqs at beginning of codelab" - The prerequisites are fairly near the beginning and consistent with the other codelabs.

Wondered whether to change all codelabs, or none, but either way remain consistent?


petele:
Hey @garanj - codelabs are exported from a Google Doc source, so changes here will be blown away on our next sync. Instead of closing this, I'll keep it open and make the changes in the source doc ASAP.

PR #4271, opened 2017-02-27
Grammar fixes

mgiuffrida:
Fix some grammatical errors and a few minor style/clarity issues.

### DIFF
--- a/src/content/en/_index.yaml
+++ b/src/content/en/_index.yaml
@@ -60,7 +60,7 @@ landing_page:
         </p>
         <ul class="checkmarks">
           <li>
-            <b>Reliable</b> - Load instantly and never show the downasaur,
+            <b>Reliable</b> - Load instantly and never show the dinosaur,
             even in uncertain network conditions.
           </li>
           <li>
--- a/src/content/en/_index.yaml
+++ b/src/content/en/_index.yaml
@@ -60,7 +60,7 @@ landing_page:
         </p>
         <ul class="checkmarks">
           <li>
-            <b>Reliable</b> - Load instantly and never show the dinosaur,
+            <b>Reliable</b> - Load instantly and never show the downasaur,
             even in uncertain network conditions.
           </li>
           <li>
--- a/src/content/en/progressive-web-apps/_index.yaml
+++ b/src/content/en/progressive-web-apps/_index.yaml
@@ -25,7 +25,7 @@ landing_page:
         </p>
         <ul>
           <li>
-            <b>Reliable</b> - Load instantly and never show the dinosaur,
+            <b>Reliable</b> - Load instantly and never show the downasaur,
             even in uncertain network conditions.
           </li>
           <li>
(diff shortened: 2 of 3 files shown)


src/content/en/_index.yaml
```diff
@@ -60,7 +60,7 @@ landing_page:
         </p>
         <ul class="checkmarks">
           <li>
-            <b>Reliable</b> - Load instantly and never show the downasaur,
+            <b>Reliable</b> - Load instantly and never show the dinosaur,
```
jpmedley: 'Downasaur' isn't a typo. It's what we call the 8-bit t-rex on Chrome's 'Unable to connect to the Internet' message.

src/content/en/progressive-web-apps/_index.yaml
```diff
@@ -25,7 +25,7 @@ landing_page:
         </p>
         <ul>
           <li>
-            <b>Reliable</b> - Load instantly and never show the downasaur,
+            <b>Reliable</b> - Load instantly and never show the dinosaur,
```
jpmedley: Keep 'downasaur'.

src/content/en/progressive-web-apps/_index.yaml
```diff
@@ -93,9 +93,9 @@ landing_page:
         <p>
           The Web App Manifest allows you to control how your app appears and
           how it's launched. You can specify icons for the home screen and
-          splash screen which is shown while the app is loading. Which page
-          is loaded when the app is launched, screen orientation, even
-          whether to show the browser chrome or not.
+          the splash screen shown while the app loads, which page to load when
+          the app launches, screen orientation, even whether or not to show the
+          browser chrome.
         </p>
```
jpmedley: That's a step in the right direction, but it could still be better. How about:

You can specify icons for the home screen, the splash screen (shown while the app is loading), the page loaded when the app is launched, screen orientation, and even, whether or not to show the browser chrome.

petele:
@mgiuffrida - would you address @jpmedley's comments. thanks!

src/content/en/_index.yaml
mgiuffrida: So, I've been on Chromium for 3 years and never heard it called this. [A Google search](https://www.google.com/search?q=chrome+%22downasaur%22) turns up few results, and the only results for site:google.com are from this copy. It's not in the Chromium code base either. Is this something web developers would be familiar with?

I'll leave this change out of my next push and let y'all decide whether to change this later.

src/content/en/progressive-web-apps/_index.yaml
mgiuffrida: I interpreted the original text differently: It seems like [the `icons` property is used for the home screen and the loading screen](https://developers.google.com/web/updates/2015/10/splashscreen), as opposed to being a separate properties allowing specific tweaking of the "splash screen". TBH it seems like a minor and confusing point (especially if we're advertising "load instantly") , so here's how it flows without mentioning the splash screen.

mgiuffrida:
@petele @jpmedley Updated to address comments. LMK if you'd like the changes should be squashed into one commit (that sometimes seems to make the pull request disappear, so I avoid it...)

PR #4275, opened 2017-02-27
Warn end users about enabling unknown sources

sbirch:


jpmedley:
@PaulKinlan (cc @petele)

Paul,
Since the whole site is geared toward developers, I think a warning is overkill. But your parenthetical in step 2b might do with a little context and possibly a link to android developer documentation.

PaulKinlan:
@jpmedley we were asked to put it in, and I agree with it. It should only be enabled for testing *your* site and not left on at all.

PR #4283, opened 2017-03-02
Fixes 404s

petele:


petele:
@kaycebasques - fyi, we had about 2k 404's on `/web/tools/chrome-devtools/javascript/add-breakpoints` in Feb, this redirects them to the main javascript node.

kaycebasques:
@petele did you say we have a tool for catching 404s?

PR #4301, opened 2017-03-09
Redo use of revokeObjectURL().

jpmedley:


src/content/en/fundamentals/getting-started/primers/media-source-extensions.md
```diff
@@ -159,10 +159,16 @@ if (window.MediaSource) {
...
 <strong>function sourceOpen(e) {
+  URL.revokeObjectURL(vidElement.src);
   // Create a SourceBuffer and get the media file.
 }</strong>
 </pre>  
 
+Notice that I've also called `revokeObjectURL()`. I can do this any time after
+the media element's `src` attribute is connected to a `MediaSource` instance.
+Calling this method doesn't destroy any objects. It _does_ allow the platform to
+handle garbage collecction at an appropriate time.
```
wolenetz: nit: s/collecction/collection/

jpmedley: https://github.com/google/WebFundamentals/pull/4310

src/content/en/fundamentals/getting-started/primers/media-source-extensions.md
```diff
@@ -159,10 +159,16 @@ if (window.MediaSource) {
 }
 
 <strong>function sourceOpen(e) {
+  URL.revokeObjectURL(vidElement.src);
   // Create a SourceBuffer and get the media file.
 }</strong>
 </pre>  
 
+Notice that I've also called `revokeObjectURL()`. I can do this any time after
```
wolenetz: nit: If revocation is done immediately after setting the vidElement's src attr to the object URL, HTML5.1 and MSE specs do not indicate any need to dereference the object URL until the "resource fetch algorithm", which occurs after a stable state after setting the src attribute. Therefore, implementations (like Chrome) may not be able to figure out that the object URL is associated with a media source object during the initial phase of the resource fetch algorithm, and load can fail. This is why I recommend putting revocation in the 'sourceopen' handler, by which time the implementation has vetted the object URL, found the mediasource object successfully, and "attached" it to the media element to the URL can then be revoked.  Revocation might also be good in a catch-all error handler, to prevent URL storage bloat, but that's a separate topic.

tl;dr: change to something "Revocation is best done after successful attachment of a MediaSource object to the media element, indicated by 'sourceopen' event" (and perhaps in an error handler, too).

PR #4334, opened 2017-03-20
Removed outdated note on background sync support

TalAter:


src/content/en/fundamentals/instant-and-offline/offline-cookbook/index.md
```diff
@@ -308,8 +308,6 @@ This code updates caches before showing a notification:
 
 <img src="images/cm-on-bg-sync.png">
 
-Dogfood: Background sync hasn't yet landed in Chrome stable.
```
petele: @TalAter - would you please also update the `{# wf_last_updated: xxx #}` field as well. We use that to ensure that translations are kept up to date.

petele:
@TalAter - That was FAST! Thanks! Will wait for build to complete then merge and push live in the next hour or so.

PR #4340, opened 2017-03-22
Reworded ControlsList API paragraph

beaufortfrancois:
Following https://github.com/google/WebFundamentals/pull/4337, what do you think @jpmedley?


beaufortfrancois:
@jpmedley Can you push this live?

PR #4353, opened 2017-03-24
Create Media Section

jpmedley:


app.yaml
```diff
@@ -2,7 +2,8 @@ runtime: python27
 api_version: 1
 threadsafe: true
 # application: petele-scratch
-# version: wf-devsite
+application: google.com:milkbone
```
petele: this is breaking the auto-deploy.

jpmedley:
@petele I've looked high and low for the source of the first three warnings without success. Any ideas?

petele:
If a file was moved/deleted, it fires that warning. It's OK to ignore, as long as you set up the appropriate redirects.

jpmedley:
The new toc isn't showing up in staging. Did I miss something?

beaufortfrancois:
For info, the WIP Mobile Web Video Playback article can be previewed now at https://pr-4353-dot-web-central.appspot.com/web/fundamentals/media/mobile-web-video-playback

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+      } else {
+        video.pause();
+      }
+    });
+
+Note: I call `event.stopPropagation()` to prevent parent handlers (e.g. video
+controls) from being notified of the click event.
+
+Rather than adjusting our video controls in the `click` event listener, we use
+the `play` and `pause` video events. When video starts playing, we change the
```
mounirlamouri: Maybe explain why this is important: making the controls events based helps with flexibility (using Media Session API) but will keep the controls working properly if the browser intervene in the playback.

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+    video.addEventListener('pause', function() {
+      playPauseButton.classList.remove('paused');
+    });
+
+When time indicated by video `currentTime` attribute changed via the
+`timeupdate` video event, we also update our custom controls if they're
+visible.
+
+    video.addEventListener('timeupdate', function() {
+      if (videoControls.classList.contains('visible')) {
```
mounirlamouri: Where is this `visible` coming from?

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+    <strong>&lt;button id="seekForwardButton">&lt;/button>
+    &lt;button id="seekBackwardButton">&lt;/button></strong>
+    &lt;div id="videoCurrentTime">&lt;/div>
+    &lt;div id="videoDuration">&lt;/div>
+    &lt;div id="videoProgressBar">&lt;/div>
+  &lt;/div>
+&lt;/div>
+</pre>
+
+    let skipTime = 10; // Time to skip in seconds
```
mounirlamouri: nit: maybe use `var` if you want to guarantee a larger compat pool?

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+<video controls controlsList="nodownload" muted playsinline>
+  <source src="/web/fundamentals/media/videos/perfect-fullscreen.webm"
+          type="video/webm">
+  <source src="/web/fundamentals/media/videos/perfect-fullscreen.mp4"
+          type="video/mp4">
+</video>
+
+### Prevent automatic fullscreen
+
+On iOS, `video` elements automagically enter fullscreen mode when media
```
mounirlamouri: **Safari** iOS maybe? I don't think Chrome iOS does this.

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+playback begins. As we're trying to tailor and control as much as possible our
+media experience across mobile browsers, I recommend you set the `playsinline`
+attribute of the `video` element to force it to play inline on iPhone and not
+enter fullscreen mode when playback begins.
```
mounirlamouri: Maybe insist that there is no side effects on other browsers as they would either follow the attribute or by default play inline?

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+
+Now that we prevent automatic fullscreen, we need to handle ourselves the
+fullscreen mode for the video with the [Fullscreen API]{: .external}. When user
+clicks the "fullscreen button", let's exit fullscreen mode with
+`document.exitFullscreen()` if fullscreen mode is currently in use by the
+document. Otherwise, request fullscreen on the video container with the method
+`requestFullscreen()` if available or fallback to `webkitEnterFullscreen()` on
+the video element only on iOS.
+
+Note: I'm going to use a [tiny shim] for the Fullscreen API in code snippets below that
```
mounirlamouri: Maybe you want @foolip to look at your shim? I'm not sure you want to use `let` in there.

mounirlamouri: In general, I don't know the fullscreen api and JS enough to have useful comments on the shim. It looks to advance for my skills :)

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+### Lock screen in landscape on button click
+
+As video may be better viewed in landscape mode, we may want to lock screen in
+landscape when user clicks the "fullscreen button". We're going to combine the
+previously used [Screen Orientation API]{: .external } and some [media
+queries]{: .external } to make sure this experience is the best.
+
+Locking screen in landscape is as easy as calling
+`screen.orientation.lock('landscape')`. However, we should do this only when
+device is in portrait mode with `matchMedia('(orientation: portrait)')` and can
```
mounirlamouri: Why matchMedia() and not `screen.orientation.type.startsWith('portrait')`?

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+As user rotates device in landscape mode, let's be smart about this and
+automatically request fullscreen to create an immersive experience. For this,
+we'll need the [Screen Orientation API]{: .external} which is not yet supported
+everywhere.  Thus, this will be our first progressive enhancement.
+
+How doest this work? As soon as we detect the screen orientation changes, let's
+request fullscreen if the browser window is in landscape mode (that is, its
+width is greater than its height). If not, let's exit fullscreen. That's all.
+
+    if ('orientation' in screen) {
```
mounirlamouri: Microsoft implements an old version of the API with a prefix, see: https://msdn.microsoft.com/en-us/library/dn433241(v=vs.85).aspx I'm not sure what set of devices have this available. I don't know if it is in Edge either.

mounirlamouri: Mozilla is into a similar situation: old version of the API with a prefix, see MDN: https://developer.mozilla.org/en/docs/Web/API/Screen/orientation

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+instance.
+
+    document.addEventListener('visibilitychange', function() {
+      // Pause video when page is hidden.
+      if (document.hidden) {
+        video.pause();
+      }
+    });
+
+Note: Chrome for Android already pauses videos when page is hidden.
```
mounirlamouri: Maybe also mention that most mobile browsers offer controls outside of the browser that allow to resume a paused video so this technique should not be used if the user is not allowed to play in the background?

src/content/en/fundamentals/media/mobile-web-video-playback.md
```diff
@@ -0,0 +1,611 @@
...
+[Media Session API]{: .external }.
+
+<pre class="prettyprint">
+playPauseButton.addEventListener('click', function() {
+  event.stopPropagation();
+  if (video.paused) {
+    video.play()
+    <strong>.then(function() {
+      setMediaSession();
+    });</strong>
```
mounirlamouri: nit: this could be written as `video.play().then(setMediaSession)`, right? I bet I must be missing some subtleties :)

src/content/en/fundamentals/media/mobile-web-video-playback.md
foolip: Nobody has yet shipped the unprefixed API, but Edge implements the webkit-prefixed one, so to work roughly everywhere it's still necessary to fall back to both webkit- and moz-prefixed APIs. And there are funny differences in capitalization, many things to get wrong. I recommend https://github.com/sindresorhus/screenfull.js

To roll your own, fall back to `element.webkitRequestFullscreen()`, not `video.webkitEnterFullscreen()`.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Done.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: It is a plain class I'm using to indicate that video controls are visible or not.
See explanation text above.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Done.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Chrome iOS does this, not only Safari.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Done.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Noted.
Done.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: What do you suggest there?

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Only because I've used one `matchMedia` call in the code. See below.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Done.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: It could of course but since `setMediaSession` is a custom function that doesn't return a promise I've decided to use that syntax to make it clear.

beaufortfrancois:
@mounirlamouri Thanks for the review. I've addressed all your feedback and asked for one clarification at https://github.com/google/WebFundamentals/pull/4353#discussion_r110166803

src/content/en/fundamentals/media/mobile-web-video-playback.md
mounirlamouri: Either try to support all the implementations but that can be hard. Firefox is easy to test but IE/Edge might be tough. Otherwise, at least mention that there are prefixed APIs if the reader care.

src/content/en/fundamentals/media/mobile-web-video-playback.md
mounirlamouri: Oups :)

src/content/en/fundamentals/media/mobile-web-video-playback.md
mounirlamouri: Oh :(

jpmedley:
@petele We are ready to go.

jpmedley:
@petele Sorry for the confusion. Folks told me they were done reviewing. Apparently not.

jpmedley:
It's ready now. @petele

src/content/en/fundamentals/media/_code/mobile-web-video-playback.js
```diff
@@ -0,0 +1,353 @@
...
+    video.classList.remove('seeking');
+  }, 200);
+});
+
+
+/* Fullscreen */
+
+toggleFullscreenButton.addEventListener('click', function(event) {
+  event.stopPropagation();
+  if (document.fullscreenElement) {
```
foolip: This will be undefined in all current browsers, so will always take the else branch. I don't have time to do a review of this section this week, but before publishing it would be good to run the code in desktop and mobile browsers based on the 4 different engines.

src/content/en/fundamentals/media/mobile-web-video-playback.md
beaufortfrancois: Done.

src/content/en/fundamentals/media/_code/mobile-web-video-playback.js
beaufortfrancois: My shim defines this as:

```js

if (!('fullscreenElement' in document)) {
  Object.defineProperty(document, 'fullscreenElement', {
    get: function() {
      return document.mozFullScreenElement || document.msFullscreenElement || document.webkitFullscreenElement;
    }
  });
}
```

See https://github.com/beaufortfrancois/sandbox/blob/gh-pages/media/tiny-fullscreen-shim.js#L4

foolip: OK, I see. I'll comment some on fullscreen-shim.js

src/content/en/fundamentals/media/_code/tiny-fullscreen-shim.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+for (var prefixedFullscreenChangeEvent of ['webkitfullscreenchange']) {
+  document.addEventListener(prefixedFullscreenChangeEvent, function(event) {
+    event.stopPropagation();
+    event.stopImmediatePropagation();
+
+    var fullscreenChange = document.createEvent('Event');
+    fullscreenChange.initEvent('fullscreenchange', true /*bubbles */, false /* cancelable */);
+    event.target.dispatchEvent(fullscreenChange);
```
foolip: The event target of the webkitfullscreenchange event is an element, but for the unprefixed fullscreenchange event it is the document. However, it's not certain that this will remain so, in order to actually get the unprefixed API shipped we may have to change it.

The best bet for now is to listen for all the events on document, since the prefixed events bubble.

src/content/en/fundamentals/media/_code/tiny-fullscreen-shim.js
```diff
@@ -0,0 +1,21 @@
+HTMLElement.prototype.requestFullscreen = HTMLElement.prototype.requestFullscreen || HTMLElement.prototype.webkitRequestFullscreen;
```
foolip: This is missing the moz variants. I would strongly encourage you to not attempt to shim the unprefixed API, and instead write helper methods that tries to use each API in order, with fallbacks. https://github.com/sindresorhus/screenfull.js is the best such library I have found, and is already used in the wild, so using that does not risk placing any additional compat constraints on the unprefixed API before it has shipped.

Note that requestFullscreen should be on Element and exitFullscreen on Document, not HTMLElement/HTMLDocument.

PR #4360, opened 2017-03-24
Fixes quotes in TOCs

petele:


jpmedley:
Is this the fix for what I sent you yesterday?

petele:
Yep

PR #4372, opened 2017-03-29
[devtools] add load performance tutorial for performance panel

kaycebasques:


src/content/en/tools/chrome-devtools/evaluate-performance/_shared/set-up.md
```diff
@@ -0,0 +1,69 @@
+## Set up {: #set-up }
```
kaycebasques: This section is in a separate file because another tutorial is going to re-use it.

kaycebasques:
@petele I've moved some content to a separate Markdown file (to re-use it in another tutorial I'm writing now). This separate MD file is causing errors, because it doesn't have all of the metadata or required tags (like page title). Isn't there some `{# wf ... #}` tag to ignore shared files like these?

kaycebasques:
Some parts of the demo aren't as compelling as I want them to be. May need to go back to the drawing board and re-visit when it's more PR-worthy. Sigh

PR #4377, opened 2017-03-30
DevTools: m57 isn't canary any longer. removing note.

paulirish:


jpmedley:
I always replace the words stable, canary, and dev with a specific version whenever I find them for exactly this reason.

kaycebasques:
I have an open PR that fixes this... #4372 

But ya, I'm going to stop mentioning stable / canary / etc. and just use the version of Chrome

kaycebasques:
That other PR fell through so using this one...

PR #4387, opened 2017-04-06
Fix Santa showcase meta tag

wibblymat:
Article metadata tags don't work if you split them across multiple lines

@samthor @petele


wibblymat:
Splitting across multiple lines means that the tag doesn't get recognised, and the content just gets put into the page

imebghomnjh

PR #4397, opened 2017-04-07
Fix variable in code sample.

NeilFraser:


jpmedley:
@kaycebasques Can you double-check this or get someone to?

kaycebasques:
Thanks for the fix, @NeilFraser

PR #4402, opened 2017-04-10
[devtools] add "coming soon" placeholder for release notes

kaycebasques:
The "What's New" panel in DevTools is going to reference a doc on /web/updates before that doc exists. This ensures that the people who try to navigate to that doc don't hit a 404.

It's a clunky solution I know, but the amount of people it'll affect is very low, so it's not worth doing any more work than this


kaycebasques:
nevermind

PR #4410, opened 2017-04-13
Clarify promise usage in service worker example

timruffles:
This change makes the use of promises in the code example clear, and safe when people copy-paste.

A lot of readers will be unclear about the important differences between `registration.then(mapSuccess, mapError)` and `registration.then(mapSuccess).catch(mapError)`, namely that in the second example `mapError` handles both registration errors *and* errors from the `mapSuccess` handler.

This change makes the comment `// registration failed :(` true regardless of changes made to the currently safe `mapSuccess` function. The error handler will now indeed run if-and-only-if the registration itself fails.

### DIFF
--- a/src/content/en/fundamentals/getting-started/primers/service-workers.md
+++ b/src/content/en/fundamentals/getting-started/primers/service-workers.md
@@ -129,7 +129,7 @@ service worker JavaScript file lives.
         navigator.serviceWorker.register('/sw.js').then(function(registration) {
           // Registration was successful
           console.log('ServiceWorker registration successful with scope: ', registration.scope);
-        }).catch(function(err) {
+        }, function(err) {
           // registration failed :(
           console.log('ServiceWorker registration failed: ', err);
         });


PaulKinlan:
I don't disagree, but want to get +1 from @jakearchibald because I am not sure if we are consistent with our other docs.

jakearchibald:
Hmm, I tend to avoid the 2nd `.then` callback, as I think it's pretty confusing to read vs an explicit `.catch`, but @timruffles is right about the difference here.

However, would we be equally concerned about code like this?

```js
try {
  doThing();
  console.log('doThing worked!');
}
catch (err) {
  console.log('doThing failed!');
}
```

timruffles:
We should if people end up doing:

```
try {
  riskyThingA();
  console.log('doThing worked!');
}
catch (err) {
  console.log('doThing failed!');
}
```

timruffles:
@jakearchibald in case you got a notification when I fat-fingered my unfinished reply: I have now finished it above!

jakearchibald:
@timruffles so, how would you demo try/catching a synchronous function?

PR #4419, opened 2017-04-14
New in Chrome 58

petele:


src/content/en/updates/2017/04/nic58.md
```diff
@@ -0,0 +1,119 @@
...
+
+## IndexedDB 2.0 {: #idb2 }
+
+The structure of your site’s database has large performance impacts, and can
+be difficult to change.
+[IndexedDB 2.0](https://hacks.mozilla.org/2016/10/whats-new-in-indexeddb-2-0/) 
+changes that.
+
+* `object` stores and `indexes` can now be renamed in-place after a
+  refactoring.
```
petele: Not quite, it means that the object stores and indexes in an indexeddb created by a developer can be renamed in place. In the past, you couldn't rename an index, but now you can.

PR #4435, opened 2017-04-18
add note about scroll event not needing to be passive

samthor:
This was confusing - I want to make it clear that `onscroll` doesn't really need a passive event listener. I think our documentation makes it sound like even the basic scroll event is something that users should worry about setting passive.

### DIFF
--- a/src/content/en/updates/2016/06/passive-event-listeners.md
+++ b/src/content/en/updates/2016/06/passive-event-listeners.md
@@ -11,15 +11,21 @@ description: New to Chrome 51, passive event listeners provide a major potential
 
 {% include "web/_shared/contributors/kaycebasques.html" %}
 
-
-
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2016/06/passive-event-listeners.md
```diff
@@ -11,16 +11,22 @@ description: New to Chrome 51, passive event listeners provide a major potential
...
 
+<aside class="note">
+  <p><strong>Note:</strong> <span>The basic `scroll` event cannot be canceled, so it does not need
+  to be set passive. However, you should still [prevent expensive work][mdn-scroll] from being
+  completed in the handler.</span></p>
+</aside>
+
+[mdn-scroll]: https://developer.mozilla.org/en-US/docs/Web/Events/scroll#Example
+
 ## How it works
```
jpmedley: @samthor 

I'm approving this, but FYI. for notes this is sufficient:

Note: blah blah blah.

It 's enough for DevSite to apply the correct formatting.

PR #4445, opened 2017-04-19
[devtools] update "optimize js execution"

kaycebasques:
Updates the section that explains how to use DevTools to analyze JS execution by fixing the screenshots and descriptions.


kaycebasques:
Done... I wonder if I could write a pre-submit hook that checks / fixes that

PR #4450, opened 2017-04-21
Adds Deep Dive for Payment request API

gauntface:
cc @zkoch

This adds the updated PaymentRequest API docs to Web Fundamentals.


kosamari:
Change set looks all 👍  and test passing. 
Holding merge until go sign on the content.

PR #4454, opened 2017-04-21
Update guide for including videos

surma:


surma:
Good point. Fixed and thanks :)

PR #4469, opened 2017-04-25
PWA ILT: changed titles to Progressive Web Apps Training

samdutton:
I don't think many readers will know what ILT means.


kosamari:
Test is warning `wf_updated_on` error but merging this since it is just tile tweak, not content update.

PR #4474, opened 2017-04-26
DevTools m58 changelog post: credit sroussey

paulirish:


kaycebasques:
@kosamari two reasons... 1) I didn't know it was fixed 2) I personally don't like the format, b/c it clutters the page and b/c I have no way of getting an alert when somebody leaves a comment... but any feedback is good feedback (and the community might help each other, so it doesn't matter if I miss the comments) so I added it back in

PR #4475, opened 2017-04-26
Update index.md

badersur:
Seems like `cache-polyfill` is not needed any more and should be removed: https://github.com/dominiccooney/cache-polyfill

Style guide suggests using 4 spaces / 1 tab for code blocks instead of <code>```</code> but both are working on DevSite! Maybe the guide should be updated.


kosamari:
I'll make separate request to update the `cache-polyfill` matter.

badersur:
@kosamari you're welcome

PR #4486, opened 2017-04-26
Add touch-action: pinch-zoom (Ch56+) and note on combining actions

patrickhlauke:


kosamari:
Hi @patrickhlauke! Thanks for the PR.
Update posts(https://developers.google.com/web/updates) are for news type articles, and looks like this post was written before Chrome 56 release. 

I've [requested to move the content out of update](https://github.com/google/WebFundamentals/issues/4488) so we can have a reference content elsewhere to keep updating instead of updating old news post. 

Thanks !

patrickhlauke:
Ah, yeah that makes sense. Thanks.

PR #4519, opened 2017-05-02
Replace “(Mac) OS X” with “macOS”

mathiasbynens:
OS X is the old name. macOS is the official name nowadays.


kosamari:
@jpmedley qucick check for approval ??

PR #4522, opened 2017-05-03
Mention ability to audit logged-in pages

dandv:
Context: https://webmasters.stackexchange.com/questions/42960/how-to-test-the-speed-of-a-page-that-is-guarded-behind-login-page/105954#105954


kosamari:
@kaycebasques quick check for approval ??

PR #4537, opened 2017-05-05
Fixed issue with notes in numbered list

nasearle:
Separated note containing bulleted list into two notes. Added spaces in front of notes in numbered list so the notes are indented and the numbers increment correctly. Removed double underscores. Fixed in source doc as well.


petele:
Nick - this'll get squashed when we run the importer again.

PR #4540, opened 2017-05-07
Add markups to support RTL languages (fixes #4539)

### LINKED ISSUE #4539: Arabic text is not displayed properly
When I switch to Arabic, I'm expecting to see almost everything (the logo, menus and the content) in Arabic and floating from right to left but it is not the case!

For example, the following pic shows text that is difficult to read:
03-hard-text

I should start reading the yellow portions, then the blue ones, then the greens and the rest either from RTL/LTR depending on the language.

I have been working on a solution, and I get this which affects the right toc and the contents:
03-expected

and would love to have this but some styling issues prevented me from proceeding:
01-html-dir-auto

badersur:
Related issue: #4539 

This solution is based on W3C articles:

1. [Creating HTML Pages in Arabic, Hebrew and Other Right-to-left Scripts (tutorial)](https://www.w3.org/International/tutorials/bidi-xhtml)
2. [Structural markup and right-to-left text in HTML](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics)

There're other articles from W3C to read about this, but I'm tired & busy to read them now and wanted to share what I have with you. Maybe someone could see this fix as an appropriate fix and add to it. Or, it could be inapplicable for the project and I should stop!

Known issues:
1. no margins in the right toc.
2. In RTL languages (tested with Arabic), lines with English text only flow from RTL! (Could be fixed by wrapping it with `<bdi>` or `<span>`, didn't test but should work. Or add `dir=auto` to every paragraph!
3. English-only headings are from RTL in RTL languages. Either it should be translated or wrapped with `<bdi>` / span

I just test Arabic and English text which look fine to me. Maybe other languages need extra treatments...

Suggested reviewers:
@PaulKinlan, @samdutton, @petele, @ianbarber (because you were involved with localisation & RTL issues)

### DIFF
--- a/gae/article.tpl
+++ b/gae/article.tpl
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="{{ lang }}">
-  <head>
+  <head dir="auto">
     <meta charset="utf-8" />
     <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Roboto:300,400,400italic,500,500italic,700,700italic|Roboto+Mono:400,500,700|Material+Icons">
     <link rel="stylesheet" href="/wf-local/styles/devsite-google-blue.css">
@@ -95,8 +95,8 @@
             {% endautoescape %}
           </nav>
           <nav class="devsite-page-nav devsite-nav">
-            <ul class="devsite-page-nav-list">
-              <li class="devsite-nav-item devsite-nav-item-heading">
+            <ul class="devsite-page-nav-list" dir="auto">
+              <li class="devsite-nav-item devsite-nav-item-heading" dir="auto">
                 <a href="#top_of_page" class="devsite-nav-title">
                   <span>Contents</span>
                 </a>
@@ -123,7 +123,7 @@
                   <span class="devsite-rating-stats"></span>
                 </div>
               </div>
-              <div class="devsite-article-body clearfix">
+              <div class="devsite-article-body clearfix" dir="auto">
                 {% autoescape off %}{{ content }}{% endautoescape %}
               </div>
               <div class="devsite-content-footer nocontent">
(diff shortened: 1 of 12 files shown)


PaulKinlan:
Thanks for the suggestions and the changes. We need to do some proper investigation around this because iirc this will only fix our development version of the site and not what is deployed via devsite.

gae/article.tpl
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="{{ lang }}">
-  <head>
+  <head dir="auto">
```
badersur: W3C suggested adding dir to the head (I think they mean ltr/rtl because it was before introducing dir=auto, or it's for older versions of HTML) but it doesn't seem to do any thing on latest versions of chrome & firefox!

src/content/en/_shared/udacity/ud811.html
```diff
@@ -1,4 +1,4 @@
-<div itemscope itemtype="http://schema.org/Course">
+<div itemscope itemtype="http://schema.org/Course" dir="ltr">
```
badersur: Some RTL translated docs doesn't have translated _shared/udacity files, so I added dir here to avoid floating from RTL in these langs!

badersur:
@PaulKinlan I did expect this to affect the dev version because I was wondering how it's deploying to DevSite while reading travis file! I hope you'll find this useful :)

badersur:
I forgot to say: you're welcome!

petele:
@PaulKinlan is right - this'll need a bit of investigation to ensure we're doing the same thing as production devsite. I'll take a look after I/O

petele:
Closing PRs that haven't been touched in 12+ months.

PR #4560, opened 2017-05-15
Updated /web/billions video to link to Developer Portraits trailer

samdutton:
Updated /web/billions to link to  [Developer Portraits trailer](https://youtu.be/9Bz5TikmJ-A)  in time for I/O NBU session.

I'll update the page to link to the video of the I/O 2017 NBU session once that's happened.


petele:
Since this is a single line change, can you resubmit with only that change, instead of ALL of the other commits?

PR #4568, opened 2017-05-17
Fix broken link and clarify how to open Cache Storage

jonerez:
Two changes:
1. Fixes link target for "Architect the App Shell" link
2. Add instructions to "Refresh Caches" in the "Cache Storage" pane on the Application panel. It took me 2 minutes to find it and I initially thought I did something wrong.


jonerez:
I signed it!

Actually I'm a Googler (johnerez@) so it told me I don't need to sign, but I thought I'd better do what googlebot told me.

petele:
Hey @jonerez - thanks for the PR. Our codelabs are actually stored as Google Docs, so changes here will be blown away when we update them from the Google Doc source.  But, I ported your changes to our Google Doc source, and then re-exported in #4569.  Thanks for your PR!!

PR #4589, opened 2017-05-23
Add a new WebVR page about how it is different from regular web

wibblymat:
I wrote a new page for the VR fundamentals section. I already got eng feedback but could do with a tech writer review.

### DIFF
--- a/src/content/en/fundamentals/vr/_toc.yaml
+++ b/src/content/en/fundamentals/vr/_toc.yaml
@@ -6,7 +6,7 @@ toc:
       path: /web/fundamentals/vr/
     - title: WebVR Status and Considerations
       path: /web/fundamentals/vr/status/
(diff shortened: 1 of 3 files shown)


src/content/en/fundamentals/vr/_toc.yaml
```diff
@@ -6,6 +6,8 @@ toc:
       path: /web/fundamentals/vr/
     - title: WebVR Status and Considerations
       path: /web/fundamentals/vr/status/
+    - title: Seeing the web from a VR perspective
```
petele: Titles should be Capital Case

wibblymat: Done

wibblymat:
Eng have added no further comments on the original doc, so assuming I got the case rules correct this is now good to go.

Thanks!

PR #4611, opened 2017-05-23
Update slide id

nasearle:
Slides source folder:

https://drive.google.com/corp/drive/folders/0B1-BUeTLlQTOQ0JyZE1JTWlLeE0


petele:
@nasearle - I've closed all these. Would you please submit a single PR with all of these changes, and verify that the new slide decks are available to the world, not just to Googlers.

PR #4613, opened 2017-05-24
Add section for HowTo: Components

surma:
This PR adds a section for the HowTo: Components to `/web/fundamentals/architecture`.

Currently, there is no components, because off of them are still under review. However, the project is being referenced in an upcoming “Designers vs Developers” video, so I want to the section to be live and have something to link to.

@petele @jpmedley PTAL (this time, this actually needs a tech writer review 😄)

@devnook @robdodson FYI


surma:
Ping @petele @jpmedley Is there any way we can get this out before/on Thursday?

PR #4625, opened 2017-05-30
make inline feedback widget more flexible

kaycebasques:
Hey @petele check it out:

* by storing the question / response data in JSON and populating the framebox with a little JS, we can support questions with any number of choices. see [evaluate-performance/index.md](https://github.com/google/WebFundamentals/compare/master...kaycebasques:wf-multichoicefeedback?expand=1#diff-dd6e42295cd623684916dbe974b717b7) for an example of a question with three choices (v1 of the widget is hard-coded to only support two choices)
* this refactor can also replace the v1 widget. it handles all of v1's features 100%. see [javascript/index.md](https://github.com/google/WebFundamentals/compare/master...kaycebasques:wf-multichoicefeedback?expand=1#diff-f755648971e7f7ec554c20e5ff5454cd) for example
* I like that I can include the JSON directly in the page. In v1 there's too much indirection and it's tedious to author
* I like the flexibility of a JS solution

In general I'm very happy with this new workflow. Let me know if you're cool with it

### DIFF
--- a/src/content/en/_shared/multichoice.html
+++ b/src/content/en/_shared/multichoice.html
@@ -1,0 +1,40 @@
+<script>
+function disableButtons() {
+  var buttons = document.querySelectorAll('button');
+  for (var i = 0; i < buttons.length; i++) {
+    buttons[i].disabled = true;
+  }
+}
+function makeButtonAndResponse(choice, responseContainer) {
+  var b = document.createElement('button'),
+      response = document.createElement('aside'),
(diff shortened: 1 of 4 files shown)


kaycebasques:
Asking that question is totally do-able... part of why I refactored

Text field is not do-able right now. I don't have a way to capture / store arbitrary data within Google Analytics. There's other ways to do it but we'd have to make sure that it's safe for us to view

PR #4638, opened 2017-06-01
Add first 2 articles to howto components

surma:
Not as urgent as the last one :)

@petele @jpmedley PTAL


surma:
Deleted the bot comment. As I discussed with @petele some time ago, the side-by-side code is generated and can’t easily be trimmed down to <100 chars per line. Hope we can work around this.

PR #4658, opened 2017-06-05
Site wide Udacity link fix

22nds:


petele:
I've reached out to Udacity to see if we can get them to fix/replace the redirector they used to have.

PR #4688, opened 2017-06-13
Add play-request-was-interrupted update article

beaufortfrancois:
R: @mounirlamouri
Preview at https://pr-4688-dot-web-central.appspot.com/web/updates/2017/06/play-request-was-interrupted
Background: https://bugs.chromium.org/p/chromium/issues/detail?id=728294#c25

### DIFF
--- a/src/content/en/updates/2017/06/play-request-was-interrupted.md
+++ b/src/content/en/updates/2017/06/play-request-was-interrupted.md
@@ -1,0 +1,146 @@
+project_path: /web/_project.yaml
+book_path: /web/updates/_book.yaml
+description: What is really happening with "DOMException: The play() request was interrupted"?
+
+{# wf_updated_on: 2017-06-13 #}
+{# wf_published_on: 2017-06-13 #}
+{# wf_tags: media,devtools #}
+{# wf_featured_image: /web/updates/images/generic/play-outline.png #}
+{# wf_featured_snippet: What is really happening with "DOMException: The play() request was interrupted"? #}
+
+# DOMException: The play() request was interrupted {: .page-title }
+
+{% include "web/_shared/contributors/beaufortfrancois.html" %}
+
+Did you just stumble upon this unexpected media error in the Chrome DevTools
+JavaScript Console?
+
+<blockquote class="warning">
+Uncaught (in promise) DOMException: The play() request was interrupted by a new
+load request.
+</blockquote>
+
+<blockquote class="warning">
+Uncaught (in promise) DOMException: The play() request was interrupted by a
+call to pause().
+</blockquote>
+
+You're in the right place then. Have no fear. I'll explain [what is causing
+this](#error) and [how to fix it](#fix).
+
+## What is causing this {: #error }
+
+Here's some JavaScript code below that reproduces the "uncaught (in promise)"
+error you're seeing:
+
+<span class="compare-worse">DON'T</span>
+```html
+<video id="video" preload="none" src="https://example.com/file.mp4"></video>
+
+<script>
+  video.play(); // <-- WRONG! This is NOT synchronous.
+  video.pause();
+</script>
+```
+
+The code above results in this error message in Chrome DevTools: 
+
+> _Uncaught (in promise) DOMException: The play() request was interrupted by a
+> call to pause()._
+
+This is due to the fact that the `play()` call is NOT synchronous, like it used
+to be. Since Chrome 50 - [March 2016], a `play()` call on an a `<video>` or
+`<audio>` element returns a [Promise], a function that returns a single result
+**asynchronously**. If playback succeeds, the Promise is fulfilled, and if
+playback fails, the Promise is rejected along with an error message explaining
+the failure.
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2017/06/play-request-was-interrupted.md
```diff
@@ -0,0 +1,146 @@
...
+</script>
+```
+
+The code above results in this error message in Chrome DevTools: 
+
+> _Uncaught (in promise) DOMException: The play() request was interrupted by a
+> call to pause()._
+
+This is due to the fact that the `play()` call is NOT synchronous, like it used
+to be. Since Chrome 50 - [March 2016], a `play()` call on an a `<video>` or
```
mounirlamouri: It did not used to be synchronous. It is just more obvious it is not, now. Websites should have listened to the `playing` event before.

src/content/en/updates/2017/06/play-request-was-interrupted.md
```diff
@@ -0,0 +1,146 @@
...
+`<audio>` element returns a [Promise], a function that returns a single result
+**asynchronously**. If playback succeeds, the Promise is fulfilled, and if
+playback fails, the Promise is rejected along with an error message explaining
+the failure.
+
+As `video.play()` is asynchronous, `video.pause()` is executed before
+video playback actually starts. This causes `video.play()` to return a rejected
+Promise as user just asked to pause video. Think about it this way:
+`video.play()` fails eventually.
```
mounirlamouri: The reason is actually that before the video did not preload, the playback will start only when the video is loaded which will be blocked by the call to `pause()`.

src/content/en/updates/2017/06/play-request-was-interrupted.md
```diff
@@ -0,0 +1,146 @@
...
+
+Note: Calling `video.pause()` isn't the only way to reset video playback state.
+It also occurs with `video.src = ''`, `video.removeAttribute('src')`, and
+`video.load()` for instance.
```
mounirlamouri: FWIW, those do more than pausing, they reset everything, including the buffer. (maybe what you wrote is enough, sharing this for background)

mounirlamouri: Also, `remoteAttribute('src')` is a no-op I believe. I don't think it loads anything.

src/content/en/updates/2017/06/play-request-was-interrupted.md
```diff
@@ -0,0 +1,146 @@
...
+    // This can be due to an autoplay policy on mobile for instance...
+  });
+</script>
+```
+
+That's great for this simple example but what if you use `video.play()` to be
+able to play a video later when user interacts with the website you may think?
+
+I'll tell you a secret... you don't have to use `video.play()`, you should use
+`video.load()`. And here's how:
```
mounirlamouri: FWIW, that kind of hacks will no longer be needed with the new autoplay policy.

src/content/en/updates/2017/06/play-request-was-interrupted.md
```diff
@@ -0,0 +1,146 @@
...
+  })
+  .catch(error => {
+    // Automatic playback failed.
+    // This can be due to an autoplay policy on mobile for instance...
+  });
+</script>
+```
+
+That's great for this simple example but what if you use `video.play()` to be
+able to play a video later when user interacts with the website you may think?
```
mounirlamouri: You don't mention autoplay. Is this on purpose?

src/content/en/updates/2017/06/play-request-was-interrupted.md
beaufortfrancois: I'll update this article when new autoplay policy is out.
It will still be useful for Safari though.

src/content/en/updates/2017/06/play-request-was-interrupted.md
beaufortfrancois: You're right for `removeAttribute('src')`. Any reason why it is not the same as `src=''`?

beaufortfrancois: @mounirlamouri What is "everything" there? I'd like to tell as much as possible

src/content/en/updates/2017/06/play-request-was-interrupted.md
beaufortfrancois: Noted.

beaufortfrancois:
No no! Mounir didn't approve yet. I'll ping you when needed ;)

petele:
sorry - already live. it didn't have the DO NOT MERGE tag, so I assumed it was good to go. Does it need to be pulled?

jpmedley:
It's not horrible. As long as Francois addresses the issues, soon, I ok
with keeping this in.

 816-678-7195

beaufortfrancois:
@jpmedley @petele See https://github.com/google/WebFundamentals/pull/4697

PR #4700, opened 2017-06-14
Fix syntax error in JS examples

fstanis:
Added commas to fix syntax error in JS code, removed '' from publicKey to be consistent with other fields.


fstanis:
CLA problem should be resolved.

PR #4749, opened 2017-06-22
Adding further tweaks to the article

gauntface:


src/content/en/updates/2017/06/budget-api.md
```diff
@@ -124,8 +123,7 @@ navigator.budget.getBudget()
 });
 ```
 
-The first entry will be your current budget and additional values will be future
-changes to your budget.
+The first entry will be your current budget and additional values will show what your budget will be at various points in the future.
```
jpmedley: How about 'is estimated to be' rather than 'will be'.

gauntface: It's not an estimate - it will be.

Issue #4767, opened 2017-06-27
Add Blink components meta data to 


ebidel:
The content metadata, add a `blink_components` array with the list of relevant components:

```yaml
blink_components:
  - Blink>Layout>Flexbox
  - Blink>CSS
```
Canonical list of components is currently:
https://blinkcomponents-b48b5.firebaseapp.com/blinkcomponents

We should look through analytics to add this to the most popular existing content. For new content, I'll add a presubmit [task](https://github.com/google/WebFundamentals/tree/master/gulp-tasks) that makes sure authors include this in new material.

jpmedley:
+1 to this for categorizing. 

Additionally, can we add the Chrome status feature ID as metadata, then we could automate adding article links to Chrome status features. https://github.com/google/WebFundamentals/issues/4769

jpmedley:
The PR I submitted only tests the validity of blink components _when they are present_. Throwing a warning for every article that doesn't contain components will create noise in the output since we have many articles, like [Surma's recent supercharged article](https://developers.google.com/web/updates/2017/06/supercharged-codesplit) that very appropriately should _not_ list any blink components.

PR #4779, opened 2017-06-28
Validate components listed in metadata.

jpmedley:
@ebidel @petele #4767 

* budget_api.md contains an incorrect component so that you can see the failing case. We'll need to remove that before merge.
* The Issue ticket says, "a presubmit task that **makes sure** authors include this in new material" (emphasis mine). This will create noise in the output since we have many articles, like [Surma's recent supercharged article](https://developers.google.com/web/updates/2017/06/supercharged-codesplit) that very appropriately should _not_ list any blink components.


gulp-tasks/test.js
```diff
@@ -502,6 +503,19 @@ function testMarkdown(filename, contents, options) {
       });
     }
 
+    // Validate listed components
+    do {
```
petele: Instead of do-while, I have a preference for `.forEach` - see `// Check for uncommon tags` further down for the method I used. Not a hard requirement, but would like to keep consistency.

src/content/en/updates/2017/06/budget-api.md
```diff
@@ -1,6 +1,9 @@
 project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: The Budget API allows developers to perform background actions without notifying users, enabling use cases like silent push.
+blink_components:
```
petele: This may break DevSite, we should use the same format we use for other wf meta tags. Specifically `{# wf_blink_components: asdf,qwer,zxcv #}`

src/data/blinkComponents.json
```diff
@@ -0,0 +1 @@
...Browser>FirstRun","UI>Browser>FullScreen","UI>Browser>History","UI>Browser>Import","UI>Browser>Infobars","UI>Browser>Interstitials","UI>Browser>Mobile","UI>Browser>Mobile(Android)","UI>Browser>Mobile>CompositedUI","UI>Browser>Mobile>SearchWidget","UI>Browser>Mobile>Share","UI>Browser>NewTabPage","UI>Browser>Offline","UI>Browser>Omnibox","UI>Browser>Passwords","UI>Browser>Passwords>Generation","UI>Browser>Payments","UI>Browser>Permissions>Prompts","UI>Browser>Preferences","UI>Browser>Preferences>Protector","UI>Browser>Previews","UI>Browser>PrintPreview","UI>Browser>Profiles","UI>Browser>Search","UI>Browser>Search>ContextualSearch","UI>Browser>Spellcheck","UI>Browser>TabCapture","UI>Browser>TabContents","UI>Browser>TabStrip","UI>Browser>Themes","UI>Browser>Toolbar","UI>Browser>VR","UI>Browser>WebUI","UI>Browser>Zoom","UI>GFX","UI>Input","UI>Input>Text>IME","UI>Input>VirtualKeyboard","UI>Localization","UI>Notifications","UI>OSIntegration","UI>ProximityAuth","UI>Security>UrlFormatting","UI>Settings","UI>Shell","UI>Shell>GestureNav","UI>Shell>LockScreen","UI>Shell>Shelf","UI>Shell>StatusArea","UI>Shell>TouchView","UI>Shell>Wallpaper","UI>Shell>WindowManager","UI>TaskManager","Webstore"]
```
petele: This should be nicely formatted, single item per line, sorted alphabetically.

gulp-tasks/test.js
```diff
@@ -502,6 +503,19 @@ function testMarkdown(filename, contents, options) {
...
+      matched = wfRegEx.RE_BLINK_COMPONENTS.exec(contents);
+      if (matched && options.blinkComponents) {
+        position = {line: getLineNumber(contents, matched.index)};
+        let component = matched[1].trim();
+        if (options.blinkComponents.indexOf(component) === -1) {
+          msg = `The component (\`${component}\`) is non-standard or misspelled.`;
+          logWarning(filename, position, msg);
```
petele: This should be an error if it doesn't match. We don't want to accidentally let warnings through.

petele:
Re:  "a presubmit task that makes sure authors include this in new material" 

We should throw a warning if there is no attribute. Most stuff should have a tag, but it's OK if it doesn't.

gulp-tasks/test.js
```diff
@@ -502,6 +503,19 @@ function testMarkdown(filename, contents, options) {
       });
     }
 
+    // Validate listed components
+    do {
+      matched = wfRegEx.RE_BLINK_COMPONENTS.exec(contents);
+      if (matched && options.blinkComponents) {
+        position = {line: getLineNumber(contents, matched.index)};
+        let component = matched[1].trim();
```
ebidel: let -> const

src/data/blinkComponents.json
ebidel: We should also have a way to keep this file updated. @jpmedley what about adding a task to gulp to update it an overwrite the file. It comes sorted and formatted from the endpoint :)

gulp-tasks/test.js
```diff
@@ -502,6 +503,19 @@ function testMarkdown(filename, contents, options) {
...
+        if (options.blinkComponents.indexOf(component) === -1) {
```
ebidel: `if (!options.blinkComponents.includes(component))`

gulp-tasks/wfRegEx.js
```diff
@@ -8,6 +8,7 @@
 var RE_BOOK_PATH = /^book_path: (.*)\n/m;
 var RE_PROJECT_PATH = /^project_path: (.*)\n/m;
 var RE_DESCRIPTION = /^description:\s?(.*)\n/m;
+var RE_BLINK_COMPONENTS = /^  - (Blink>[\w>]+)/gm;
```
ebidel: there are some components that don't start with `Blink>`. Most that we use will however. Instead of regexs, is there a way to get the yaml list and then compare that to the `blinkComponents` array?

```
const usesAnUnknownComponent = blinkComponentsInYaml.filter(c => !blinkComponents.includes(c)).length > 0;
```

gulp-tasks/test.js
jpmedley: I tried to do that, but the proposed syntax for the blink components is multi-line. That forces me into doing multiple calls to regex.exec().

So, leave as is or change the syntax.

src/content/en/updates/2017/06/budget-api.md
jpmedley: That would solve my do-loop problem.

gulp-tasks/test.js
petele: change the syntax - use the `{# wf_blink_component_id: a,b,c #}`. The YAML at the top of the page is used by DevSite and this may break the way DevSite parses a page.

gulp-tasks/wfRegEx.js
petele: Yes - use the way same style/syntax we use for tags.

src/data/blinkComponents.json
petele: I'd like to see that go into `tools/update-resources.sh` - right now it updates the devsite resources, but we should expand it so it also updates the list of components as well.

gulp-tasks/test.js
jpmedley: I'm following Pete's style.

petele: Use Eric's here, he's right. I should have done it in my previous one.

jpmedley:
PTAL

jpmedley:
"I'd like to see that go into tools/update-resources.sh - right now it
updates the devsite resources, but we should expand it so it also updates
the list of components as well."

Do you reall think that's necessary? How often do Blink components change?

"We should throw a warning if there is no attribute. Most stuff should have
a tag, but it's OK if it doesn't."

I did it, but frankly I'm not crazy about the extra noise.

 816-678-7195

src/data/blinkComponents.json
```diff
@@ -0,0 +1,1277 @@
+{
+  "AAA-README": [
+    "", 
+    "This file is generated by src/tools/checkteamtags/extract_components.py",
```
petele: is this actually the case?

gulp-tasks/test.js
```diff
@@ -497,11 +498,27 @@ function testMarkdown(filename, contents, options) {
         tag = tag.trim();
         if (options.commonTags.indexOf(tag) === -1) {
           msg = `Uncommon tag (\`${tag}\`) found.`;
-          logWarning(filename, position, msg);
+          logError(filename, position, msg);
```
petele: commonTags shouldn't change and should remain a warning.

src/data/blinkComponents.json
jpmedley: @ebidel Eric?

gulp-tasks/test.js
jpmedley: Cut and paste error. Sorry about that.

src/data/blinkComponents.json
ebidel: I would pull from https://blinkcomponents-399b3.firebaseapp.com/blinkcomponents for now. You can format it using https://jsonformatter.curiousconcept.com/

jpmedley:
@ebidel Re: "We should also have a way to keep this file updated. @jpmedley what about adding a task to gulp to update it an overwrite the file. It comes sorted and formatted from the endpoint :)"

There's a couple ways I can do this. I'll implement it in another PR. In the meantime, I've opened a ticket so that we can discuss the approach.

https://github.com/google/WebFundamentals/issues/4786

src/data/blinkComponents.json
jpmedley: Done. @petele I think this is your only outstanding issue. Can I merge now?

gulp-tasks/test.js
```diff
@@ -502,6 +503,22 @@ function testMarkdown(filename, contents, options) {
       });
     }
 
+    // Check for valid Blink components
+    matched = wfRegEx.RE_BLINK_COMPONENTS.exec(contents);
+    if (matched && options.blinkComponents) {
+      position = {line: getLineNumber(contents, matched.index)};
+      matched[1].split(',').forEach(function(component) {
+        component = component.trim();
+        if (options.blinkComponents.indexOf(component) === -1) {
```
ebidel: `if (!options.blinkComponents.includes(component)) {` yea?

jpmedley: No brain fart this time. This really is consistent with Pete's other tests.

ebidel:
Let's land this today so I can start testing against.

ebidel:
Pete is on vacation. Looks like all of his feedback was addressed. Merging!

PR #4836, opened 2017-07-13
fixed translation error for zh-cn

molee1905:
the original text is **there's data**,  so **如果缓存中有数据** is correct
> If there's **data** in the cache, it'll be returned and rendered extremely quickly (tens of milliseconds) and update the card only if the XHR is still outstanding


molee1905:
I signed it!

PR #4845, opened 2017-07-14
Make `wf_blink_components` tag name mono space

kosamari:
Tiny fix to make log message nicer like other tag. 

update_old_sw-helpers_link_by_kosamari_ _pull_request__4843_ _google_webfundamentals


jpmedley:
Mistook this for an issue instead of PR, which is why I duplicated it. Monday morning brain fog.

PR #4856, opened 2017-07-19
Update a11y language

robdodson:
Tiny update to improve this article so it doesn't seem like we're giving legal advice.


robdodson:
I'm not sure what a `wf_blink_components` field would be for?

PR #4862, opened 2017-07-22
Fix typo from Guideance to Guidance

sonyarianto:


sonyarianto:
will fix the CLA first

PR #4865, opened 2017-07-24
Fix code sample syntax error

sonyarianto:
Fix code sample error on this article https://developers.google.com/web/updates/2017/04/headless-chrome


sonyarianto:
@jpmedley found another syntax error

PR #4871, opened 2017-07-25
Misc. copy edits.

jpmedley:


jpmedley:
@petele These errors are on files that are NOT in this PR.

PR #4970, opened 2017-08-29
Dvd podcast s203

mustafa-x:
Ready to merge :)


src/content/en/shows/designer-vs-developer/_toc.yaml
```diff
@@ -3,6 +3,8 @@ toc:
     section:
     - title: 'Podcast'
       section:
+      - title: 'Improving your creative confidence'
```
petele: Just caught this now, but titles should be Capital Case. I'll merge for now, but we should fix in the future.

PR #4977, opened 2017-08-30
fix typo in cm-api update article

agektmr:
Fixes typos in CM API update article.


agektmr:
I think you meant `wf_updated_on`. Fixed.

PR #5020, opened 2017-09-10
Add missing words to be understandable in spanish

007lva:
Another option could be: 

> Eso nos libera de **muchas** implementaciones


007lva:
I signed it!

PR #5026, opened 2017-09-12
Fixed nits in Web Share article

beaufortfrancois:
FYI @petele @mgiuca


src/content/en/updates/2016/09/navigator-share.md
```diff
@@ -75,7 +75,7 @@ experience to the user. For example, you might do:
 ```js
 let url = document.location.href;
 const canonicalElement = document.querySelector('link[rel=canonical]');
-if (canonicalElement !== undefined) {
+if (canonicalElement !== null) {
```
petele: would it make more sense to use:
```
if (canonicalElement && canonicalElement.href) {
  url = canonicalElement.href;
}
```

PR #5040, opened 2017-09-19
Updates to web storage articles

wibblymat:
@Meggin this article fell out of the Snapshot work


src/content/en/fundamentals/instant-and-offline/web-storage/storing-user-generated-content.md
```diff
@@ -0,0 +1,109 @@
+project_path: /web/_project.yaml
+book_path: /web/fundamentals/_book.yaml
+description: Learn how to store user generated content like images, media, and data files in a progressively enhanceable way.
+
+{# wf_updated_on: 2017-09-25 #}
+{# wf_published_on: 2017-09-25 #}
+{# wf_tags: indexeddb #}
```
ebidel: needs `wf_blink_components`

petele:
@addyosmani - where does this fit in the new perf/loading IA?

petele:
@jpmedley - would you please do a review on this?

src/content/en/fundamentals/instant-and-offline/web-storage/storing-user-generated-content.md
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+The other direction is slightly more involved, and is an asynchronous process. You can use a
+`FileReader` object to read the blob as an `ArrayBuffer`. When the reading is finished a `loadend`
+event is triggered on the reader. You can wrap this process in a `Promise` like so:
+
+    function blobToArrayBuffer(blob) {
+      return new Promise((resolve, reject) => {
+        const reader = new FileReader();
+        reader.addEventListener('loadend', async (e) => {
```
beaufortfrancois: I believe you don't need `async`

beaufortfrancois: As you don't use `e`, you may want to omit it as well.

```js
    reader.addEventListener('loadend', () => {
      resolve(reader.result);
    });
```

Meggin:
Proposed new approach for the doc:

**Purpose**
This doc shows developers how to cache data when offline and then sync that data with the server once online again

In the intro, you would explain that the recommended way to cache data offline is to use service workers with Cache API. However, not all browsers support this, and the doc will provide work-around strategies with indexedDB to acommondate this.

**First section**
The first section should cover determining the file types that need to be cached and coming up with a caching strategy for those file types. There’s quite a bit of docs on caching strategies that you could summarize briefly and point to, but at the very least, you are making the developer aware that they need to have made this decision before they jump in.

The doc does mention that this doc covers storage of larger file types, but I’m kind of wondering what the other types of common storage might be and the solutions for those too.  If it really is just for larger files, then I think we need to make sure this is clear in the purpose.

**Second section**
You should explain how the developer needs to determine if service workers are supported in the browser, and target storage to Cache API if it is, or else target indexedDB if service workers aren’t supported. I recommend showing the check for this.

And this is a chance to explain the two different cache storage options, why one is better than the other, briefly, with the main focus on checking browser compatibility.

**Third section**
This section will cover how to cache data in either Cache API or indexedDB depending on the browswer support. I recommend replacing the very basic code for Cache API with code from a viewable working app, such as snapshot.

And I’d include the code for storing in Cache API, and equivalent code for storing the exact same data in indexedDB. That parallel sample is quite useful to developers.

**Fourth section**
Show how to test that the caching is working. This would cover going into DevTools and showing that a file has been cached either in Cache API if service worker supported or in indexedDB if not. A screenshot is great for this-- and you can show the either or easy enough.

**Fifth section**
Explain how to sync cached data to the server from either Cache API or indexedDB and back it up with two samples. This is a good section to cover version control in as well. And you should also show how to delete the data item from the cache once it’s been stored. 

This might be an instance where Workbox could greatly simplify the work developers have to do, as I think does a lot for syncing and version control.

Meggin:
Some questions:

- The doc says this: “This article focuses on storing larger items”. Is there a totally different way to cache and sync smaller items? If so, why not mention that here and link to a doc on it. If this doc is only about storing certain file types, you should explicitly call that out and it would be helpful to provide resources for other types of content developers might want to store and how to handle it differently.

- The doc says ‘Since the content is in a URL’. Maybe this has something to do with the above question-- you are talking about content that’s been uploaded, like a file or something, and so it has a URL. As mentioned above, you need to be more explicit about what types of content you are covering in the doc, as well as types of content that are outside the scope.

- What’s a good way to connect a local URL to a real global URL on the server? I’d like to see a specific code snippet on how to do this elegantly.

- When covering poor performance for IndexedDB, the doc mentions that the Cache API response body doesn’t include the body, the body is streamed in later. Is there a way to access response.body? Does the developer presumably need to write the code to stream the body later? If so, this is very important to cover in the doc with a code snippet. 

- Is it necessary to convert an ArrayBugger to a blob? Can you just store the array? Is there a reason why it’s better to do this conversion?
- How do I go about storing the object in indexedDB once I’ve converted it to a blob? Can you also point me to lots of good docs on indexedDB.

PaulKinlan:
Just to add on top of Meggin's comments. 

I need to be able to take away something actionable. I am cool with this article being about storing large objects in the client. But... Cache API does not work in anything but Blink and Firefox at the moment, so indexeddb is the only solution and we don't show any real code about what to persist and how to persist it given that we can't store a blob in iOS, how should the user recreate that in an objectStore? Cache API doesn't allow querying outside of get and set of an object, in snapshot you display a list of images that have been stored, how should we structure that in the indexeddb?

wibblymat:
@PaulKinlan @Meggin 
After consulting with you both, I have changed direction. Instead of introducing a new article here, I will put the content that I wanted to add (i.e. devs should use IDB because it is cross-platform, but need to convert Blobs to ArrayBuffers) into the existing articles.

I will do another PR for a new article about syncing from client to server using background sync.

I want to try to address the rest of the comments about actionable content and adding more code samples. I think that this might best be served by adding two new (missing) articles to the section - one showing how to use the CacheStorage API, and another showing how to use IDB. They can both be practical, code oriented articles.

We can do more work on this section over time to make the rest of the articles more practical, too.

WDYT?

wibblymat:
Rebased onto the new IA. Added the Cache API doc.

The warning from the bot is about a line that I don't intend to fix. The line is a single opening `<a>` tag, and wrapping it any further will make the doc harder to read, not easier.

src/content/en/fundamentals/instant-and-offline/web-storage/cache-api.md
```diff
@@ -0,0 +1,248 @@
...
+The API can be accessed from a window, iframe, worker, or service worker.
+
+## What can be stored
+
+The caches only store pairs of `Request` and `Response` objects, representing HTTP requests and
+responses, respectively. However, the requests and responses can contain any kind of data that can
+be transferred over HTTP.
+
+Generally speaking, if you are using the API for generic storage, your `Request` objects will not
+need more data than just the URL to be used as a key.
```
Meggin: Sentence reads a bit funny. How about something like:

If you are using the Cache API for generic storage, store the URL for the `Request` object:

src/content/en/fundamentals/instant-and-offline/web-storage/cache-api.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Searching
+
+The Cache API does not provide a way to search for requests or responses except for matching entries
+against a `Response` object. However, you can implement your own search using the features that are
+available.
+
+### Filtering
+
+One way is to iterate over all entries and filter down to the ones that you want. Let's say that you
```
Meggin: Link your first sentences back up to the intro, something like:

One way to implement your own search is to iterate over all the entries...

src/content/en/fundamentals/instant-and-offline/web-storage/cache-api.md
```diff
@@ -0,0 +1,248 @@
...
+        }
+      }
+
+      return result;
+    }
+
+This way you can use any property of the `Request` and `Response` objects to filter the entries.
+Note that this will, of course, be slow if you search over large sets of data.
+
+### Creating an index
```
Meggin: Add a connector to the intro:

Another way to implement your own search is to maintain a separate index of...

src/content/en/fundamentals/instant-and-offline/web-storage/cache-api.md
```diff
@@ -0,0 +1,248 @@
...
+
+The other way is to maintain a separate index of entries that can be searched, stored in IndexedDB.
+Since this is the kind of operation that IndexedDB was designed for it will have much better
+performance with large numbers of entries.
+
+If you store the URL of the `Request` alongside the searchable properties then you can easily
+retrieve the correct cache entry after doing the search.
```
Meggin: Add mentioned before, it is good to show the IndexedDB code.

src/content/en/fundamentals/instant-and-offline/web-storage/cache-api.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Adding to a cache
+
+There are three ways to add an item to a cache. All three methods return a `Promise`.
+
```
Meggin: Introduce the three methods in the intro, and then create three subsections that cover each of them. It's a bit easier to follow when you've put the three ways into their own subsections with titles that call them out.

src/content/en/fundamentals/instant-and-offline/web-storage/cache-api.md
```diff
@@ -0,0 +1,248 @@
...
+The first is to use `cache.put(request, response)`. `request` is either a `Request` object or a
+string - if it is a string, then `new Request(request)` will be used instead. `response` must be a
```
Meggin: General comment to avoid future tense and replace it with active voice.

So something like:

If it is a string, use `new Request(request)` instead.

src/content/en/fundamentals/instant-and-offline/web-storage/cache-api.md
```diff
@@ -0,0 +1,248 @@
...
+`Response`. This pair will be stored in the cache.
```
Meggin: Another example of 'will' future tense. Use something like: 'This pair is stored in the cache.

src/content/en/fundamentals/instant-and-offline/web-storage/index.md
```diff
@@ -179,15 +192,16 @@ many browsers as possible and which offer asynchronous call models, to maximize
 interoperability with the UI. These criteria lead naturally to the following
 technology choices:
 
-* For device local key/value storage use the Cache API.
+* For making your application itself available offline, use the Cache API. This API is available in
```
Meggin: This is a run-on sentence and hard to parse. Try something like:

For offline storage, use the Cache API. This API is available in any browser that supports Service Worker technology necessary for creating offline apps. The Cache API is ideal for storing resources associated with a known URL.

src/content/en/fundamentals/instant-and-offline/web-storage/index.md
```diff
@@ -179,15 +192,16 @@ many browsers as possible and which offer asynchronous call models, to maximize
...
+  any browser that supports the core Service Worker technology that is necessary for offline apps,
+  and is ideal for storing resources that are associated with a known URL
 
-* For device local structured storage: use IndexedDB.
+* For storing application state and user-generated content use IndexedDB. This enables users to work
```
Meggin: Small nit-- add missing comma before 'use IndexedDB'.

wibblymat:
@Meggin resolved almost everything you mentioned, including going through and removing every instance of 'will'. I get confused about active versus passive voice, so it's possible that this doesn't completely fix the problem - let me know.

The one thing I haven't done yet is add some code for storing a search index in IndexedDB. I'm still working on that alongside the IndexedDB counterpart of the Cache API article.

src/content/en/fundamentals/instant-and-offline/web-storage/index.md
```diff
@@ -179,15 +192,16 @@ many browsers as possible and which offer asynchronous call models, to maximize
 interoperability with the UI. These criteria lead naturally to the following
 technology choices:
 
-* For device local key/value storage use the Cache API.
+* For offline storage, use the Cache API. This API is available in any browser that supports Service
```
Meggin: Small nit-- make Service Worker technology link to https://jakearchibald.github.io/isserviceworkerready/.

src/content/en/fundamentals/instant-and-offline/web-storage/index.md
```diff
@@ -223,6 +237,6 @@ of interest:
 
 * Deep Dive: IndexedDB (coming soon)
 
-* Deep Dive: Cache API (coming soon)
+* [Deep Dive: Cache API](cache-api)
```
Meggin: So I only just now saw that we have a list of docs that are coming soon-- this is weird to me. Just how long does coming soon imply? I'd be inclined to delete the whole list, and just link to Cache API doc, as mentioned in other comment, above.

src/content/en/fundamentals/instant-and-offline/web-storage/index.md
```diff
@@ -179,15 +192,16 @@ many browsers as possible and which offer asynchronous call models, to maximize
 interoperability with the UI. These criteria lead naturally to the following
 technology choices:
 
-* For device local key/value storage use the Cache API.
+* For offline storage, use the Cache API. This API is available in any browser that supports Service
+Worker technology necessary for creating offline apps. The Cache API is ideal for storing resources
```
Meggin: Small nit: make 'Cache API' a link to new cache API doc.

wibblymat:
Done, thanks @Meggin!

wibblymat:
All reviews dealt with. The last remaining warning is about a line that is a single `<a>` tag. I don't believe that wrapping it any further will make the document easier to read for anyone.

Dismissing the reviews was apparently a mistake, since it leaves the PR still unable to be merged by me. @petele @kosamari, since you will be the next committers awake :)

wibblymat:
@petele @Meggin @kosamari ping.

I believe that everything is good here but that I messed up when I mistakenly dismissed the reviews. I think this just needs merging by someone with commit access.

PR #5041, opened 2017-09-20
New update: CSS position:sticky custom event

ebidel:


ebidel:
Is there a reason for the file size limits? They're not much over 2.5MB.

petele:
GIFs that are that huge often work better as autoplay muted videos. In almost every experiment I've done, serving a webm+mp4 works much better, is higher quality, etc. I'm also concerned about page weight for users, this page would end up being almost 10megs!

PR #5044, opened 2017-09-20
headless: update chromedriver note

ebidel:


JKereliuk:
I think the example is still a bit misleading, for example here:

`const chromeCapabilities = webdriver.Capabilities.chrome();
chromeCapabilities.set('chromeOptions', {
  binary: PATH_TO_CANARY // Screenshots require Chrome 60. Force Canary.
  'args': [
    '--headless',
  ]
});`

The comment implies that Canary is required, but this will work with the current stable release.
What do you think?

ebidel:
@JKereliuk good call, missed those. How about now?

src/content/en/updates/2017/04/headless-chrome.md
```diff
@@ -391,9 +391,9 @@ examples below to get you started.
 #### Using ChromeDriver
 
 [ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/) 2.3.0
```
JKereliuk: nit: we refer to versions in the format 2.30 and not 2.3.0

ebidel: done

src/content/en/updates/2017/04/headless-chrome.md
```diff
@@ -391,9 +391,9 @@ examples below to get you started.
...
-supports Chrome 59 and later and works with headless Chrome. In some cases, you may need
-Chrome 60 to work around bugs. For example, there are known issues with taking
-screenshots in Chrome 59.
+was the first version to support Chrome 59 and work with headless Chrome,
+but you'll want Chrome 61 to work around screenshot bugs. The good news
+is that more recent versions of ChromeDriver (2.3.2) work with Chrome 61!
```
JKereliuk: nit: 2.3.2 should be 2.32

ebidel: done :)

JKereliuk:
Was just about to comment on the paragraph, but just saw that you changed it to something like what I was going to suggest!

LGTM, thanks for being quick to respond :)

src/content/en/updates/2017/04/headless-chrome.md
Hexcles: Thanks for the change! It's a lot less intimidating to use ChromeDriver with headless Chrome now.

I'm a bit confused by this paragraph. It says ChromeDriver 2.3.0 "supports" Chrome 59, but needs Chrome 61 because of some bugs, and then says there's this new ChromeDriver 2.3.2 which also works with Chrome 61.

IMHO it'd be much clearer if we only mention the latest and working versions, especially since 61 is stable now. Something like:

> [ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/) supports headless Chrome. ChromeDriver 2.3.2 and Chrome 61 are the minimum versions required because there were screenshot bugs in older versions.

(Of course the wording could use some improvement!)

ebidel: Yea, I already changed it to this:

`[ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/) 2.32 uses Chrome 61 and works well with headless Chrome.`

Hexcles:
Ditto. I didn't see the new change while writing the comment. That was fast! Thanks!

PR #5076, opened 2017-09-28
Adding jsdoc reference template

gauntface:
@petele this is a devsite friendly template that anyone can use when publishing to WebFundamentals.

Can you give it a look and see what you think. Wasn't sure if `tools/` was the best place for it, but seemed like best place to start with.


jpmedley:
@gauntface  What is this for? Can you provide some context?

gauntface:
@jpmedley Workbox needs reference docs which is built with jsdoc. To work with devsite, JSDocs needs a custom template to use. This is that template.

PR #5084, opened 2017-10-02
Fixed IntersectionObserver links

samdutton:
@surma fyi in case the updated W3C links are not correct.

Not sure quite why there are a few whitespace changes :/.


surma:
LGTM, thanks Sam

PR #5086, opened 2017-10-02
Add usage warning to howto-checkbox

robdodson:
oops, got some `test` warning. hold plzzzzz


robdodson:
ok so the warnings are all from the generated code that goes inside the framebox. I didn't actually touch any of that, I just added a 'warning' to the docs. @petele I think this is ready for review.

PR #5091, opened 2017-10-04
Add `Intl.PluralRules` article

mathiasbynens:


mathiasbynens:
Added a paragraph at the end.

PR #5092, opened 2017-10-04
Fix typo

lucab85:
Fix typo


PaulKinlan:
This is a most preferable change. thanks.

PR #5102, opened 2017-10-04
Updated some glossary terms

nasearle:
- Removed unnecessary quotes in the app shell definition 
- Made descriptions consistently repeat the term being defined in the first sentence of the definition
- Fixed grammar in the service worker definition
- Removed init caps from some terms (e.g., changed "Add to Home Screen" to "add to home screen")

### DIFF
--- a/src/data/glossary.yaml
+++ b/src/data/glossary.yaml
@@ -18,28 +18,28 @@
 #   - title: Your First PWA Code Lab (REQUIRED)
 #     link: /web/fundamentals/something (REQUIRED)
 
-- term: Accelerated Mobile pages
+- term: Accelerated Mobile Pages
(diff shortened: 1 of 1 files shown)


src/data/glossary.yaml
```diff
@@ -111,20 +111,20 @@
     title: Detect DOM changes with Mutation Observers
     link: /web/updates/2012/02/Detect-DOM-changes-with-Mutation-Observers
 
-- term: Progressive Web App
+- term: progressive web app
```
petele: Progressive Web Apps is a proper term and we should be capitalizing it.

petele:
@jpmedley - can you confirm the capitalization on these?

PR #5103, opened 2017-10-04
Update 'Improved A2HS' article with details about rollout.

yfriedman:
Also remove mentions of unknown sources since it's unnecessary.


PaulKinlan:
Thank you!

PR #5109, opened 2017-10-05
fix typo

lucab85:
Fix typos.
Prefer American spelling as PR #5009.


petele:
@gauntface - before I merge this, can you make sure to fix the sources as well please.

gauntface:
I thought these files shouldn't be touched but only ever updated via the drive doc.

This actually requires me to update the source code to the american spelling to be correct.

petele:
@lucab85 - you're doing the right thing, thank you! The source for `src/content/en/fundamentals/codelabs/push-notifications/index.md` actually comes from a Google Doc that we have internally, that's why your changes keep getting blown away. @gauntface owns the Google Doc, so we need to update in there. 

I'm going to merge now so that we can at least get the fix for add to home screen, but hold off on updating the code lab one again. I've filed a bug against @gauntface to fix in the source.

PR #5116, opened 2017-10-06
Fix 2 runtime errors, fix styling

bryanjenningz:
Running `var player = document.getElementById('player')'` caused an error because it's a syntax error.
Running `source.connect(processor);` caused an error because `source` wasn't defined.
I fixed these 2 errors.
I also added some small styling fixes.


bryanjenningz:
I signed it!

PR #5129, opened 2017-10-11
Fix save form page

deeyehgo:
## Issue
`/` is not wrapped inside the single quote which is causing the code example to not look formatted correctly.

## Affected page
https://developers.google.com/web/fundamentals/security/credential-management/save-forms

## Screenshot
image


deeyehgo:
I signed it!

agektmr:
LGTM!

PR #5143, opened 2017-10-13
Removing article that uses pepe in examples

jakearchibald:


jakearchibald:
Just removing the videos and references to the videos

Issue #5142, opened 2017-10-13
Urgent: Hate Symbol of alt-right used in dev tools examples


jakearchibald:
https://developers.google.com/web/updates/2017/10/devtools-release-notes

Pepe appears in multiple examples here. This needs removed, now.

https://en.m.wikipedia.org/wiki/Pepe_the_Frog

jakearchibald:
It's in both the first and second video

jakearchibald:
https://github.com/google/WebFundamentals/pull/5143

jakearchibald:
Fixed.

clobber:
LMAO this is hilarious - it's a cartoon frog.

PR #5145, opened 2017-10-13
Devtools article without first two vids

jakearchibald:
Followup to https://github.com/google/WebFundamentals/issues/5142


jakearchibald:
The comment widget was missing from the original, but I'm happy to add it.

PR #5163, opened 2017-10-17
Fix controlsList DOMTokenList behaviour

beaufortfrancois:
FYI @mounirlamouri


petele:
@beaufortfrancois - would you fix the conflict please?

PR #5165, opened 2017-10-17
Update WebVR spec status in prep for CDS.

jpmedley:
There are no substantive content changes in this, just meta-content related to the spec status.


toji:
LGTM too, FWIW. Thanks!

PR #5178, opened 2017-10-18
Add formatting fixes. Add numbers to progressive enhancement paragraph.

jpmedley:


jpmedley:
@petele Since I'm doing these for next week, can I get a pass on the 100 char limit?

PR #5180, opened 2017-10-18
[lighthouse] add v2.5 update post

kaycebasques:
@vinamratasingal I can't add you as a reviewer but you should be able to start one anyways

Here's how the rendered page looks. The error about V's contributor file should be fixed when the site gets properly built

ex


vinamratasingal:
Looks good to me! LGTM.

PR #5197, opened 2017-10-20
Fixed spacing error

naruthk:
I fixed the spacing error on the Web Fundamental phrase.


naruthk:
I signed it!

petele:
Thanks for the PR, we actually did that intentionally, sort of using WebFundamentals as some of our branding across different places.

PR #5202, opened 2017-10-21
Fix translation in accessible-styles

yuheiy:


yuheiy:
let the bot run again

PR #5217, opened 2017-10-24
Use the correct document to get template for Custom Element. Fix #5032

### LINKED ISSUE #5032: Code example for "Creating elements from a <template>" does not always work
Hi there,

The following code snippet (found [here](https://developers.google.com/web/fundamentals/architecture/building-components/shadowdom)) does not work when extracted in its own .html file:

```
<template id="x-foo-from-template">
  <style>
    p { color: orange; }
  </style>
  <p>I'm in Shadow DOM. My markup was stamped from a &lt;template&gt;.</p>
</template>

<script>
  customElements.define('x-foo-from-template', class extends HTMLElement {
    constructor() {
      super(); // always call super() first in the ctor.
      let shadowRoot = this.attachShadow({mode: 'open'});
      const t = document.querySelector('#x-foo-from-template');
      const instance = t.content.cloneNode(true);
      shadowRoot.appendChild(instance);
    }
    ...
  });
</script>
```
I have the following error in the console Chrome `TypeError: Cannot read property 'content' of null`
because `document.querySelector('#x-foo-from-template');` returns `null`. 

When I put directly the code snippet in my main HTML file (inside off putting it in it's own HTML file and importing it), I works fine. Am I missing something here?

I tried with `document.querySelector('template');` too but it doesn't work.

imsingh:
It fixes the #5032. In the example, it uses document.querySelector to get the template for custom element. But correct method will be document.currentScript.ownerDocument.querySelector. Because template resides in different document.

### DIFF
--- a/src/content/en/fundamentals/web-components/customelements.md
+++ b/src/content/en/fundamentals/web-components/customelements.md
@@ -732,11 +732,11 @@ structure of a custom element**.
     </template>
 
     <script>
+      const t = document.currentScript.ownerDocument.querySelector('#x-foo-from-template');
       customElements.define('x-foo-from-template', class extends HTMLElement {
         constructor() {
           super(); // always call super() first in the ctor.
           let shadowRoot = this.attachShadow({mode: 'open'});
-          const t = document.querySelector('#x-foo-from-template');
           const instance = t.content.cloneNode(true);
           shadowRoot.appendChild(instance);
         }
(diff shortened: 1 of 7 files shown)


petele:
@robdodson - can you confirm and approve?

robdodson:
The ownerDocument trick should only be required if you're using an HTML Import. If you're loading the element with a script tag or ES module, document.querySelector should work.

robdodson:
Maybe we should instead add a note to the doc that if you're using an HTML Import you should use that ownerDocument syntax.

imsingh:
@robdodson also, this will fix the demo for the example. Right now it is also broken and looks like this image. Take a look at box below <ul> list. It is empty.
bug

robdodson:
Planning to update and fix this either tomorrow or over the weekend.

robdodson:

petele:
@robdodson fixed in #5229

PR #5229, opened 2017-10-30
Update cusom elements primer. Fix demos. Fixes #5217, #5032

robdodson:
I took this as an opportunity to organize the primer so it's a bit easier to read.
I added [a comment](https://github.com/google/WebFundamentals/compare/ce-primer-update?expand=1#diff-cd83fbd6d9fe3e6e33a0fadc04bb42dcR612) to address the issue raised by @imsingh and fixed all of the demos.

@petele what's the best path to get this translated to other languages? Most of the changes I made were just reorganizing things and adding notes, so even though it looks like a big PR it's mostly just house cleaning.

I did choose to delete the `adoptedCallback` example because it's extremely esoteric and I think throws folks off. I've literally never encountered a use for `adoptedCallback` nor has anyone on the Polymer team. I don't even think Polymer bothers to wrap it.

### DIFF
--- a/src/content/en/fundamentals/web-components/customelements.md
+++ b/src/content/en/fundamentals/web-components/customelements.md
@@ -167,191 +167,66 @@ element's properties, inspect its children (`this.children`), query nodes
    to be self-closing. Always write a closing tag
    (<code>&lt;app-drawer&gt;&lt;/app-drawer&gt;</code>).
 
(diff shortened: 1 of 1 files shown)


petele:
CLA is good - came from #5217

petele:
Let me see what I can find about localization this week.

PR #5238, opened 2017-10-31
[workbox] add new "get started with webpack" tutorial

kaycebasques:
@addyosmani or @gauntface or @jeffposnick plz review!

You can preview at https://pr-5238-dot-web-central.appspot.com/web/tools/workbox/get-started/webpack


jeffposnick:
Thanks, @kaycebasques! I'm assuming the content is identical to the draft that was previously reviewed via Google Docs?

I noticed an issue with the left-hand nav bar in your staging instance:

screen shot 2017-10-31 at 3 00 05 pm

If that's just an artifact of staging and that won't be used as the left-hand nav in prod, then feel free to ignore—can you confirm?

kaycebasques:
Thanks @jeffposnick 

I'm just modifying an existing file, and the ToC appears to display fine on prod, so it seems to just be a shortcoming in staging.

The content is essentially the same. I fleshed out some bits and added screenshots and re-arranged some info, but yes essentially the same as the draft you've already reviewed.

petele:
@kaycebasques - Should I wait for an LGTM from @addyosmani and/or @gauntface before merging?

addyosmani:
Sorry for only seeing this late. Post-merge LGTM

PR #5240, opened 2017-10-31
Tweak to Spanish translation

kant:


petele:
@kant - Thanks for the PR, can you tell me why your suggestion is better than what we have? Comparing in Google Translate seems to be pretty similar.

JoseAlcerreca:
I would write "las" 

"the following main areas"

Currently it says
"some following main areas" which I agree is wrong.

and the PR suggests:
"these following main areas"

petele:
@kant - any updates on this.

PR #5248, opened 2017-11-03
Fixes #5126 - Updates image CRP analysis-dom-css-js-async.png

### LINKED ISSUE #5126: Flow diagram needs to be corrected
Hi,

Flow diagram on this page:
https://developers.google.com/web/fundamentals/performance/critical-rendering-path/analyzing-crp

needs to be corrected in below image:
https://developers.google.com/web/fundamentals/performance/critical-rendering-path/images/analysis-dom-css-js-async.png

to swap:
RUN JS <-> RENDER PAGE

ChathuminaVimukthi:
Corrected the flow diagram


ChathuminaVimukthi:
I signed it!

ChathuminaVimukthi:
Name of the above file is a bit different from the original file. I'm adding the file again with the correct name. 
analysis-dom-css-js-async

ChathuminaVimukthi:
Deleted the specific file.

petele:
Thanks for the fast turn around! Sorry to ask one more, the file went from 14K to 935K! Any way to get it back closer to its original size?

ChathuminaVimukthi:
No problem :) I will make the file size smaller.
Happy to contribute!

PaulKinlan:
Wait. Can we hold off on merging. I'd like to get @igrigorik and @addyosmani to chime in.

The async attribute still blocks onload, but not the creation of the CSSOM or DOM construction. I'd just like either Addy or Ilya to validate the intent of the original image vs this.

ChathuminaVimukthi:
No problem at all :)

igrigorik:
Is there a summary for what the intent is here? `<script async>` executes as soon as the fetch completes, so the existing diagram is not wrong. As Paul noted, the benefit of async is that it doesn't block DOM construction (which current diagram captures also), and rendering.. could happen anytime after we have both DOM + CSSOM.

petele:
@ChathuminaVimukthi - can you address @igrigorik's comments.

petele:
@ChathuminaVimukthi - any updates on this?

ChathuminaVimukthi:
I understood the whole async thing by referring to igrigorik's comment and I agree with @igrigorik. It seems that both diagrams are correct. I think it is not necessary to merge my updated diagram :)

And sorry for the late reply.

PR #5252, opened 2017-11-05
Update webpack.md to clarify `urlPattern` operation

johnnyreilly:
See https://github.com/GoogleChrome/workbox/issues/977#issuecomment-341982446 for details

### DIFF
--- a/src/content/en/tools/workbox/get-started/webpack.md
+++ b/src/content/en/tools/workbox/get-started/webpack.md
@@ -313,10 +313,10 @@ that they had an internet connection.
 1. Open `webpack.config.js` again.
 1. Add a `runtimeCaching` property to your Workbox configuration.
    `urlPattern` is a regular expression pattern telling Workbox which
-   URLs to store locally. *(Please note your regex will be matched
-   against full URLs such as 'https://hacker-news.firebaseio.com' -
-   not against partial URLs. This means you can cache both content from
-   your site and content from external sites as well, as this example
+   URLs to store locally. *(When your app makes a network request at 
+   runtime, Workbox caches any request that matches the regular 
+   expression in `handler`, regardless of its origin. This means that 
+   you can cache content from external sites as well, as this example
    demonstrates.)* `handler` defines the caching strategy that Workbox
    uses for any matching URL. See [The Offline Cookbook][cookbook] for more
    on caching strategies.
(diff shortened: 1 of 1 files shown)


petele:
@kaycebasques - can you please confirm?

johnnyreilly:
Sorry @petele I don't know what the `wf_updated_on` field is?

petele:
No worries - on line 5 you'll see something like `{# wf_updated_on: 2017-01-01 #}` can you update that with the current date please.

johnnyreilly:
Oh sure - will do when I get to a keyboard!

johnnyreilly:
Done!

src/content/en/tools/workbox/get-started/webpack.md
```diff
@@ -313,7 +313,11 @@ that they had an internet connection.
 1. Open `webpack.config.js` again.
 1. Add a `runtimeCaching` property to your Workbox configuration.
    `urlPattern` is a regular expression pattern telling Workbox which
-   URLs to store locally. `handler` defines the caching strategy that Workbox
+   URLs to store locally. *(Please note your regex will be matched
```
kaycebasques: Our style guide instructs us to avoid the use of "please". See [Politeness](https://developers.google.com/style/tone#politeness-and-use-of-please).

src/content/en/tools/workbox/get-started/webpack.md
```diff
@@ -313,7 +313,11 @@ that they had an internet connection.
...
+   against full URLs such as 'https://hacker-news.firebaseio.com' -
+   not against partial URLs. This means you can cache both content from
```
kaycebasques: The use of the word "not" here is a little problematic to me. I don't think I want to use "full URLs" or "partial URLs" as the main distinction. The idea here is that if your app makes a network request at runtime, and you have a runtime caching handler that matches the regular expression, then that resource gets cached, regardless of its origin.

So how about we rephrase your section to:

```When your app makes a network request at runtime, Workbox caches any request that matches the regular expression in `handler`, regardless of its origin. This means that you can cache content from external sites as well.```

src/content/en/tools/workbox/get-started/webpack.md
johnnyreilly: Such a shame - the world's so harsh nowadays. C'est la vie

src/content/en/tools/workbox/get-started/webpack.md
johnnyreilly: Done - though I think it's a loss not being explicit that the URLs are "full" rather than partial.  I burned 2 days thinking workbox was broken, going on the assumption that the regex was being tested against partial URLs, i.e. local URLs.  I reckon others may make the same false assumption; as it's common to only think about local resources when developing a website.

The redraft loses that clarity and I think that may lead others down the same rabbithole.  Essentially I wanted to save them that.

PR #5275, opened 2017-11-13
[Translation] Translated ud891.html into Simplified Chinese

chikathreesix:
@limhenry translated this file on GitLocalize. Create a pull request on behalf of him.

[Check out a review request at GitLocalize](https://gitlocalize.com/repo/107/zh-CN/review/284)

### DIFF
--- a/src/content/zh-cn/_shared/udacity/ud891.html
+++ b/src/content/zh-cn/_shared/udacity/ud891.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
+<html><body>
+<div itemscope itemtype="http://schema.org/Course">
+  <h2 itemprop="name">网站易用性</h2>
+  <a href="https://www.udacity.com/course/web-accessibility--ud891">
+    <img src="/web/shows/udacity/img/ud891.jpg" class="attempt-right" itemprop="image">
+  </a>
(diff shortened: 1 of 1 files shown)


chikathreesix:
I signed it!

chikathreesix:
Hi @limhenry, I have created a pull request from GitLocalize since your review request hasn't been reviewed for a while. But the problem is, it seems that I need your confirmation. Could you check this, please?

limhenry:
@chikathreesix yes, the PR is OK for me 😄 

cc @agektmr

agektmr:
@chikathreesix Does this mean there needs to be an admin to send a Pull Request on GitLocalize on behalf of a translator?

chikathreesix:
@agektmr Not only admins but also language moderators can send a pull request on GitLocalize.

petele:
I'll merge for now, but we should figure out a better way to handle this since the CLA chain becomes painful otherwise.

petele:
CLA - chain OK.

sotayamashita:
@petele 

> I'll merge for now, but we should figure out a better way to handle this since the CLA chain becomes painful otherwise.

Do you have any idea?

agektmr:
@chikathreesix That sounds like we need at least one person admin/moderator taking care of a language, right? That's ideal in a long term, but not a short term.

agektmr:
@sotayamashita I'm happy to take care of this at least until we figure out a better solution.

chikathreesix:
@agektmr Yes, correct. Actually, that's what I was considering as well. If it's difficult to assign someone as a moderator for now, in this case, it might be better to send a pull request directly from GitLocalize (not through review request)

PR #5285, opened 2017-11-15
Workbox v3

gauntface:
cc @petele @PaulKinlan @addyosmani @jeffposnick @philipwalton 

Next version of Workbox documentation. Doesn't need to be publicly linked from the current /web/tools/workbox/ but I will add a link in the `_book.yaml` in a seperate PR when we are ready. This PR just introduces the site into devsite (everything can be public) allowing us to iterate on the docs in smaller chunks and see the final result on DevSite (Which is currently different to the appengine instance in this Repo).


petele:
Hey Matt - would you please split this into two PRs? One that includes the changes to the staging server code and one that has your content? The current size makes it super hard to do any kind of review on.

PR #5290, opened 2017-11-16
Changes to DevSite Changes for Workbox

gauntface:
Workbox Appengine changes


gulp-tasks/build.js
```diff
@@ -20,6 +20,10 @@ gulp.task('build:announcement', function() {
   const globOpts = {
     srcBase: 'src/content/en/',
     prefixBase: true,
+    ignore: [
```
petele: Why are you disabling this? It won't be disabled on the real site.

gulp-tasks/claat-export.js
```diff
@@ -4,7 +4,6 @@ const gulp = require('gulp');
 const path = require('path');
 const glob = require('globule');
 const wfHelper = require('./wfHelper');
-const runSequence = require('run-sequence');
```
petele: Please rebase and add this stuff back

gulp-tasks/workbox:generate-contributors.js
```diff
@@ -0,0 +1,94 @@
...
+const fs = require('fs-extra');
+const path = require('path');
+const gulp = require('gulp');
+
+const MAX_COLLABORATORS = 20;
+
+function getContributorHTML(contributorStats) {
+  const contributor = contributorStats.author;
+  var html = `
+<div class="wb-contrib">
```
petele: can we do this using templates so we can keep code to build and template code separate?

gulp-tasks/workbox:generate-contributors.js
```diff
@@ -0,0 +1,94 @@
+const GitHubApi = require('github');
```
petele: some filesystems may puke on a filename with a colon in it, can we change the colon?

gulp-tasks/workbox:generate-contributors.js
```diff
@@ -0,0 +1,94 @@
...
+  <a class="wb-contrib__link" href="${contributor.url}" target="_blank">
+    ${contributor.login}
+  </a>
+</div>
+`;
+  return html;
+}
+
+gulp.task(`workbox:generate-contributors`, function() {
+  if (!process.env.GITHUB_TOKEN) {
```
petele: Let's change this to the same way that the youtube token works, that way not everyone needs to have this setup.

gulp-tasks/workbox:generate-contributors.js
```diff
@@ -0,0 +1,94 @@
...
+  })
+  .then((contributorStats) => {
+    contributorStats.sort((a, b) => b.total - a.total);
+    if (contributorStats.length > MAX_COLLABORATORS) {
+      contributorStats = contributorStats.splice(0, MAX_COLLABORATORS);
+    }
+    return contributorStats.map(getContributorHTML).join('');
+  })
+  .then((html) => {
+    const finalHtml = `{% framebox height="auto" %}
```
petele: Same comment about templating.

gulp-tasks/workbox:reference-docs-v3.js
```diff
@@ -0,0 +1,194 @@
+const gulp = require('gulp');
```
petele: same comment about filename

gulp-tasks/workbox:generate-contributors.js
```diff
@@ -0,0 +1,94 @@
...
+    ${contributor.login}
+  </a>
+</div>
+`;
+  return html;
+}
+
+gulp.task(`workbox:generate-contributors`, function() {
+  if (!process.env.GITHUB_TOKEN) {
+    console.warn(`Unable to build Workbox contributors due to no ` +
```
petele: to keep things consistent, can you please use `gutil.log` (and throughout)

gulp-tasks/workbox:reference-docs-v3.js
```diff
@@ -0,0 +1,194 @@
...
+const remoteGitTags = require('remote-git-tags');
+const semver = require('semver');
+const fse = require('fs-extra');
+const path = require('path');
+const os = require('os');
+const spawn = require('child_process').spawn;
```
petele: in wfHelper.js there's a `function promisedExec(cmd, cwd)` that will does all this for you.

gulp-tasks/build.js
gauntface: Because I don't want the announcements on workbox.

gulp-tasks/workbox:generate-contributors.js
gauntface: Can you give me an example of doing this please - I struggle with figuring out the devsite stuff at the best of times.

gulp-tasks/build.js
petele: Understood, but we want to be able to put them across all pages, so for now, let's leave them on and we can talk about removing them later.

gulp-tasks/workbox-generate-contributors.js
```diff
@@ -0,0 +1,60 @@
...
+const MAX_COLLABORATORS = 20;
+
+/**
+ * Generates a list of contributors with links and images into
+ * tools/workbox/templates using the Github API to get the top 20
+ * contributors.
+ */
+
+gulp.task(`workbox-generate-contributors`, function() {
+  var token = process.env.GITHUB_TOKEN;
```
petele: var, var? not let? what are we, in 1997? :P j/k

PR #5302, opened 2017-11-17
Webfont loading/optimization updates

jeffposnick:
CC: @notwaldorf @robdodson @addyosmani @igrigorik for feedback.

This may still need some work, but it's my initial cut at at update to https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization

All feedback is welcome!

The primary changes are:
- Removing references to inlined base64-encoded fonts (see https://www.bramstein.com/writing/web-font-anti-patterns-inlining.html)
- Pointing people towards `<link rel="preload">` as the easiest way to download a font earlier.
- Moving most of the `font-display` material from [Rob's updates article](https://developers.google.com/web/updates/2016/02/font-display) into the relevant section of this article.
- Somewhat steering folks toward those two options vs. the Font Loading API, though I'm not very aggressive about that.
- Putting in a few bits about service workers (of course!).

One area that I didn't cover, but was thinking about adding in, was a detailed explanation of how to use a fallback font with similar dimensions to your primary font (chosen using https://meowni.ca/font-style-matcher/) to minimize layout churn. What are thoughts about putting that into this post vs. pointing people to standalone resources?

### DIFF
--- a/src/content/en/fundamentals/performance/optimizing-content-efficiency/webfont-optimization.md
+++ b/src/content/en/fundamentals/performance/optimizing-content-efficiency/webfont-optimization.md
@@ -280,7 +280,7 @@ The next section describes a number of options for customizing this default beha
 
 If there's a high probability that your page will need a specific Webfont hosted
 at a URL you know in advance, you can take advantage of a new web platform
-feature: [`<link rel="preload">`](https://developers.google.com/web/fundamentals/performance/resource-prioritization).
+feature: [`<link rel="preload">`](/web/fundamentals/performance/resource-prioritization).
 
 It allows you to include an element in your HTML, usually as part of the
 `<head>`, that will trigger a request for the Webfont early in the critical
@@ -522,9 +522,10 @@ to do or it may do it as a low priority depending on what it thinks would be
 best for the user. This can be beneficial in situations where the user is on a
 weak connection and pulling down a font may not be the best use of resources.
 
-`font-display` is not [widely enabled](https://caniuse.com/#feat=css-font-rendering-controls)
-in browsers other than Chrome as of this writing, but it should bring
-consistency to browser behaviors as it gets wider support.
+`font-display` is [gaining adoption](https://caniuse.com/#feat=css-font-rendering-controls)
+in many modern browsers. You can look forward to consistency in browser
+behavior as it becomes widely implemented.
+
 
 ### The Font Loading API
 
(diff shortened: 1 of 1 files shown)


robdodson:
wow this looks good to me. super comprehensive!

src/content/en/fundamentals/performance/optimizing-content-efficiency/webfont-optimization.md
```diff
@@ -271,16 +272,266 @@ Lazy loading of fonts carries an important hidden implication that may delay tex
...
-* IE immediately renders with the fallback font if the request font is not yet available, and re-renders the text after the font download is complete.
+omits any text.
+
+The next section describes a number of options for customizing this default behavior.
+
+### Preload your Webfont resources
+
+If there's a high probability that your page will need a specific Webfont hosted
+at a URL you know in advance, you can take advantage of a new web platform
+feature: [`<link rel="preload">`](https://developers.google.com/web/fundamentals/performance/resource-prioritization).
```
addyosmani: The linter will warn that these references include the fully qualified developers.google.com path. You might want to go back and drop this back to `/web/fundamentals/..`.

src/content/en/fundamentals/performance/optimizing-content-efficiency/webfont-optimization.md
```diff
@@ -271,16 +272,266 @@ Lazy loading of fonts carries an important hidden implication that may delay tex
...
+- **`optional`** gives the font face an extremely small block period (100ms or less is
+recommended in most cases) and a zero second swap period. Similar to `fallback`,
+this is a good choice for when the downloading font is more of a "nice to have"
+but not critical to the experience. The `optional` value leaves it up to the
+browser to decide whether to initiate the font download, which it may choose not
+to do or it may do it as a low priority depending on what it thinks would be
+best for the user. This can be beneficial in situations where the user is on a
+weak connection and pulling down a font may not be the best use of resources.
+
+`font-display` is not [widely enabled](https://caniuse.com/#feat=css-font-rendering-controls)
```
addyosmani: Given FF and the Safari TP will both be gaining font-display in their next release (and it often takes us time to update these guides), I wonder if a statement that's a little more future-proof would be useful here. e.g font-display support is improving in modern browsers. Be sure to evaluate if browsers you're targeting support it by checking out CanIUse.

jeffposnick: SGTM.

src/content/en/fundamentals/performance/optimizing-content-efficiency/webfont-optimization.md
jeffposnick: Yup, will do.

jeffposnick:
Thanks! I've avoided fixing the line length linting warnings in this PR, as it would make the diffs way too noisy. I believe the other feedback and linting issues are addressed.

PaulKinlan:
Can we resolve the line length warnings

PaulKinlan:
To jeff's point, I am not sure what diff issue would be if we sort the line lengths out.... also this is going to appear every time there is a change to this content.

jeffposnick:
@PaulKinlan There's a decent number of inline changes within the text of existing paragraphs, and those changes would be lost in the noise in GitHub's diff viewer if I explicitly wrapped all of the lines to 100 characters.

But since we're now at the point where the content changes have been reviewed by Rob and Addy, I'll add in the line rewrapping in another commit.

jeffposnick:
Linting fixes are in, along with a corresponding increase in the noisiness of the diff.

PR #5305, opened 2017-11-20
forms: type=number requires [pattern] on iOS

paulirish:


PaulKinlan:
I'm just gonna double check this because the suggestion in this article is different to fillament group's suggestion.

PaulKinlan:
https://jsbin.com/fijurozela < tests... works

PR #5306, opened 2017-11-21
Change JavaScript MIME type

tomayac:
The IETF considers text/javascript [obsolete](https://tools.ietf.org/html/rfc4329#section-7) and recommends going for application/javascript.


tomayac:
From [W3C](https://www.w3.org/TR/html5/infrastructure.html): *"Similarly, the MIME type used to refer to JavaScript in this specification is ```text/javascript```, since that is the most commonly used type, despite it being an officially obsoleted type according to RFC 4329".* 

Closing :-)

PR #5307, opened 2017-11-21
Correct comment about static import

mathiasbynens:
Thanks to @robpalme for pointing this out. https://twitter.com/robpalmer2/status/932920137107288064


mathiasbynens:
The `script` warnings are false positives. They occur within code blocks.

PR #5313, opened 2017-11-22
[Translation zh-tw] _project.yaml Open

agektmr:
[Check out a review request at GitLocalize](https://gitlocalize.com/repo/107/zh-TW/review/346)

### DIFF
--- a/src/content/zh-tw/_project.yaml
+++ b/src/content/zh-tw/_project.yaml
@@ -3,3 +3,19 @@ announcement:
   description: </#cdata-section></style><style><#cdata-section> 錯過了今年的 <b>Chrome 开发峰会</b>?
     您可以在我們的YouTube播放列表觀看全部的峰会視頻。 <a href="\"https://www.youtube.com/playlist?list=PLNYkxOF6rcICUD5nBfRdAR6Fveosnqa5m\""
     class="\"gc-analytics-event\"" data-category="\"webFu\"" data-label="\"cds-watch-playlist\"">观看</a>
+buganizer_id: '180451'
+color: google-blue
+content_license: cc3-apache2
+footer_path: "/web/_footer.yaml"
+google_analytics_ids:
+- UA-52746336-1
+home_url: "/web/"
+icon:
+  path: "/web/images/web-fundamentals-icon192x192.png"
+is_family_root: t
+name: Web
+tags:
+- html5
+- "\t\nmobile"
+- 
+- web
--- a/src/content/zh-tw/_project.yaml
+++ b/src/content/zh-tw/_project.yaml
@@ -1,0 +1,5 @@
+---
+announcement:
+  description: </#cdata-section></style><style><#cdata-section> 錯過了今年的 <b>Chrome 开发峰会</b>?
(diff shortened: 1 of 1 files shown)


agektmr:
Reviewed by @limhenry translated by @kennith

src/content/zh-tw/_project.yaml
```diff
@@ -0,0 +1,22 @@
...
+buganizer_id: '180451'
+color: google-blue
+content_license: cc3-apache2
+footer_path: "/web/_footer.yaml"
+google_analytics_ids:
+- UA-52746336-1
+home_url: "/web/"
+icon:
+  path: "/web/images/web-fundamentals-icon192x192.png"
+is_family_root: t
```
agektmr: `t` instead of `true`
@sotayamashita

src/content/zh-tw/_project.yaml
```diff
@@ -0,0 +1,22 @@
+---
+announcement:
+  description: </#cdata-section></style><style><#cdata-section> 錯過了今年的 <b>Chrome 开发峰会</b>?
```
agektmr: Contents inside `<style>` tag is missing
@sotayamashita

src/content/zh-tw/_project.yaml
```diff
@@ -0,0 +1,22 @@
...
+google_analytics_ids:
+- UA-52746336-1
+home_url: "/web/"
+icon:
+  path: "/web/images/web-fundamentals-icon192x192.png"
+is_family_root: t
+name: Web
+tags:
+- html5
+- "\t\nmobile"
```
agektmr: Odd `\t\n` prepended
@sotayamashita

sotayamashita: @agektmr the reason is that some user add the HTML special characters.

src/content/zh-tw/_project.yaml
sotayamashita: @agektmr Sorry for the inconvenience, I’ll keep you posted when we figure out how to handle the Boolean type. https://github.com/gitlocalize/feedback/issues/27

src/content/zh-tw/_project.yaml
sotayamashita: @agektmr I’ll keep you posted when we figure out how to handle the Boolean type. I will overwrite the value directly until the issue will be solved. gitlocalize/feedback#28

src/content/zh-tw/_project.yaml
agektmr: @sotayamashita do you mean in translation or in original document? If this happens in GitLocalize then you should probably trim them automatically in GitLocalize?

petele:
Due to the way DevSite works, localizing some files will cause some pretty serious headaches:
* `_book.yaml`, `_toc.yaml` or `_upper_tabs-*.yaml`
* `_project.yaml`

But, we can do `_footer.yaml`!

Sorry! :(

src/content/zh-tw/_project.yaml
chikathreesix: In this case, it seems that a tab and a line feed were added by a user from the GitLocalize editor. In YAML file, we should consider that a string might have those characters intentionally, so if removing them it would raise another issue, I think.

agektmr: @chikathreesix Got it. Let's leave this for now as this PR is deprecated. I'll ping you next time this comes up as an issue.

PR #5320, opened 2017-11-23
Review of automating-image-optimization page

lucab85:
Improved "automating-image-optimization" page


PaulKinlan:
cc @addyosmani as fyi.

It appears we can't run travis against remotely hosted PR's but this looks reasonable.

PR #5324, opened 2017-11-24
Added missing await operator

kyarik:
Since caches.match() returns a promise, with the code as is, the async function
will always return that promise. Thus, if what we want is in cache, everything is OK,
but if it's not, the promise passed to e.respondWith() will be resolved with undefined, which
is not what we want. To fix this, we await the promise returned by caches.match().
If it resolves with a response object, we return it, otherwise, the || will make sure that
`fetchResponseP' is returned instead of undefined.


petele:
@kyarik - please resync with master to fix the build break

@jeffposnick - can you review for technical accuracy?

kyarik:
Updated

PR #5352, opened 2017-11-29
Fix BigQuery link

rviscomi:


rviscomi:
Fixed `wf_updated_on`. I think the other warnings are bogus.

PR #5378, opened 2017-11-30
fixes indentation on index.yaml include

petele:
`contributors.html` is included `_index.yaml`, which means indentation must match the yaml file, otherwise it's not processed as valid yaml.


petele:
cc @gauntface

PR #5379, opened 2017-11-30
Adds ESLint support for gulp-tasks

petele:
cc: @gauntface 

Most "errors" are only warnings for now, but we'll change that as we get files cleaned up.


petele:
I finished up gulpfile.js, and moved all warnings to full on errors now.

PR #5400, opened 2017-12-01
Adding some guidance on when to use inject vs generate

gauntface:


src/content/en/tools/workbox/_shared/when-to-use-generate-sw.md
```diff
@@ -0,0 +1,12 @@
+{# wf_md_include #}
+
+<p><span class="compare-better">When to use generateSW</span></p>
+
+- You want to precache files.
+- You have simple runtime configuration needs (i.e. the configuration allows
```
philipwalton: I think you mean to use "e.g." here and below.

PR #5424, opened 2017-12-04
Update zh-tw translation

kennith:
[Check out a review request at GitLocalize](https://gitlocalize.com/repo/107/zh-TW/review/424)


petele:
I don't believe that we need local translated versions of redirects. The english versions should cover all of this. If not, lmk and I'll take a look.

PR #5430, opened 2017-12-05
[Translation zh-tw] Fundamentals

kennith:
[Check out a review request at GitLocalize](https://gitlocalize.com/repo/107/zh-TW/review/429)


petele:
please do not translate _book.yaml files

PR #5473, opened 2017-12-11
[Translation zh-tw] Tools > Chrome Devtools > Evaluate Performance

limhenry:
Chinese (Traditional) translation of:

- src/content/zh-tw/tools/chrome-devtools/evaluate-performance/performance-reference.md
- src/content/zh-tw/tools/chrome-devtools/evaluate-performance/timeline-tool.md

Note: All the articles are directly translated from zh-CN.


VivianWhitman:
hi

PR #5489, opened 2017-12-12
Adding plugins guide

gauntface:
Adds Plugins Guide for Workbox


gauntface:
I don't think it's mentioned anywhere in V3 JSDocs (probably because the RequestWrapper was removed).

PR #5491, opened 2017-12-12
Adding generate SW docs and updating sw checklist with alpha banner

gauntface:
This adds a generate service worker guide.


gauntface:
Totally agree with your comments. I'm imagining your stuff will end up in the relevant modules/workbox-(cli, build, webpack-plugin) and this will be a higher level guide that will just cover getting from a -> b as quickly as possible and then call out for modules for extra info and more options so :+1:

PR #5493, opened 2017-12-13
Workbox advanced recipes

gauntface:
Advanced Recipes for docs

### DIFF
--- a/src/content/en/tools/workbox/guides/advanced-recipes.md
+++ b/src/content/en/tools/workbox/guides/advanced-recipes.md
@@ -3,7 +3,8 @@ book_path: /web/tools/workbox/_book.yaml
 description: Advanced recipes to use with Workbox.
 
 {# wf_updated_on: 2017-12-17 #}
-{# wf_published_on: 2017-11-15 #}
+{# wf_published_on: 2017-12-17 #}
+{# wf_blink_components: N/A #}
 
 # Advanced Recipes {: .page-title }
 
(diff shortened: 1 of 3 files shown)


src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -0,0 +1,137 @@
...
+
+{# wf_updated_on: 2017-12-17 #}
+{# wf_published_on: 2017-12-17 #}
+{# wf_blink_components: N/A #}
+
+# Advanced Recipes {: .page-title }
+
+{% include "web/tools/workbox/_shared/alpha.html" %}
+
+## Offer a page reload for users
```
jeffposnick: My gut is that we should link to https://redfin.engineering/how-to-fix-the-refresh-button-when-using-service-workers-a8e27af6df68 instead of including a recipe inline, because there are a number of different considerations and tradeoffs that that post does a good job of explaining.

If you'd prefer to include the same code inline rather than linking out then I could offer some line-by-line feedback, but I'll hold off until we decide.

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Provide a fallback response to a route
+
+There are scenarios where returning a fallback response is better than failing
+to return a response at all. An example is returning a placeholder image when
+the original image can't be retrieved.
+
+To do this in Workbox you can use the `handle()` method on strategy to make
+a custom handler function.
+
```
jeffposnick: Can you include a note here that the developers should ensure `FALLBACK_IMAGE_URL` is already cached, presumable via the precache manifest?

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -0,0 +1,173 @@
...
+A common UX pattern for progressive web apps is to show a banner when a service
+worker has updated and waiting to install.
+
+To do this you'll need to add some code to your page and to your service worker.
+
+**Add to your page**
+
+```javascript
+const showRefreshUI = (registration) => {
```
jeffposnick: I tend to avoid arrow functions and `const` in code that's meant to run inside of a `window` context, as it breaks parsing the JS in older browsers, and not everyone transpiles.

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -0,0 +1,173 @@
...
+      // NOOP
+      break;
+  }
+});
+```
+
+This will receive a the 'force-activate' message and call `skipWaiting()` and
+`clients.claim()` so that the service worker activates immediately and controls
+all of the currently open windows, which in turn results in the `
+controllerchange` event firing causing the windows to reload.
```
jeffposnick: It's the `message` event, not the `controllerchange` event, that causes the `window`s to reload themselves.

PR #5502, opened 2017-12-13
Fix PR template date format

philipwalton:
The `YYYY-DD-MM` format is non-standard and has probably confused a few folks; this PR updates it to the standard.


petele:
thanks @philipwalton - that was a total goof on my part!

Issue #5485, opened 2017-12-12
framebox escaping issue


beaufortfrancois:
I'm having some issue with escaping inside framebox in /web post
In https://github.com/google/WebFundamentals/commit/89dbe0d0d66dbb59a5194a1288afdae2e5590d52#diff-83e6426eb86c98e522a73a7757ce10dbR222, the JS string `>=` is replaced with `&gt;= ` in prod while it stays unchanged in staging.

Prod: https://developers.google.com/web/updates/2017/12/chrome-63-64-media-updates#hdr-video-playback-windows-10
Staging: https://pr-5427-dot-web-central.appspot.com/web/updates/2017/12/chrome-63-64-media-updates#hdr-video-playback-windows-10

I'm not sure how to fix this.

beaufortfrancois:
Please merge and push live as soon as possible.

petele:
this appears to be a DevSite issue, not an issue with our devserver. :( will send you instructions on how to file with them.

PR #5510, opened 2017-12-14
Add workaround for framebox escaping

beaufortfrancois:
FIX https://github.com/google/WebFundamentals/issues/5485

R: @petele


beaufortfrancois:
@petele Do you mind pushing it live?

PR #5516, opened 2017-12-14
[lighthouse] add "aspect ratio" reference

kaycebasques:
@patrickhulce @ThisIzKp I took a stab at "recommendations" for the aspect ratio lighthouse audit, but it's probably incomplete. Suggestions?

- adds "aspect ratio" reference

**Fixes:** N/A

**Target Live Date:** 2017-12-21

- [ ] This has been reviewed and approved by @patrickhulce  

**R:** @petele

### DIFF
--- a/src/content/en/tools/lighthouse/audits/aspect-ratio.md
+++ b/src/content/en/tools/lighthouse/audits/aspect-ratio.md
@@ -2,8 +2,8 @@ project_path: /web/tools/_project.yaml
 book_path: /web/tools/_book.yaml
 description: Reference documentation for the "Displays Images With Incorrect Aspect Ratio" Lighthouse audit.
 
-{# wf_updated_on: 2017-12-14 #}
-{# wf_published_on: 2017-12-14 #}
+{# wf_updated_on: 2018-01-03 #}
+{# wf_published_on: 2018-01-03 #}
 {# wf_blink_components: N/A #}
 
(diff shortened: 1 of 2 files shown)


src/content/en/tools/lighthouse/audits/aspect-ratio.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Overview {: #overview }
+
+If a rendered image has a significantly different aspect ratio from the aspect ratio in its
+source file (the "natural" aspect ratio), then the rendered image may look distorted, possibly
+creating an unpleasant user experience.
+
+## Recommendations {: #recommendations }
+
+Use `vw` for *both* the width and height of images.
```
patrickhulce: Hm not sure about the focus on `vw` here since we'd be perfectly happy here if people used no explicit size at all, any fixed sizes that were the same ratio`img.half-size { width: var(--real-width) / 2; height: var(--real-height) / 2; }`, or [object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit).

Maybe it would be better to just say to avoid setting width and height as % of a variably-sized container?

kaycebasques: SGTM! Yeah I was just winging it and couldn't find a good general rule...

kaycebasques:
@patrickhulce check the updates

PR #5498, opened 2017-12-13
Standalone docs for Workbox build tool configuration

jeffposnick:
This takes much of the content at https://developers.google.com/web/tools/workbox/reference-docs/prerelease/module-workbox-build#.Configuration, which is currently generated from JSDocs, and puts it into a format that more accurately reflects the configuration that each build option currently supports.

Assuming this PR is merged, we'll need follow-up changes to the JSDocs in the Workbox project to point to the new URL.

**Target Live Date:** 2018-12-XX

- [ ] This has been reviewed and approved by @gauntface
- [X] I have run `gulp test` locally and all tests pass.
- [X] I have added the appropriate `type-something` label.
- [X] I've staged the site and manually verified that my content displays correctly.

**R:** @petele

PR #5521, opened 2017-12-15
Modules workbox webpack plugin

jeffposnick:
This adds a module guide for the `workbox-webpack-plugin` to the overall Workbox docs set. This is a follow-up from #5498.

CC: @johnnyreilly for any feedback he might have, as some of the content attempts to address questions he's raised while testing out `workbox-webpack-plugin`. 

**Target Live Date:** 2017-12-XX

- [x] This has been reviewed and approved by (@gauntface)

**R:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
+++ b/src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
@@ -13,7 +13,9 @@
     <p>
       This only needs to be set if you want to precache local files matching <code>glob</code>
       patterns. It does not need to be set if you have <code>self.__precacheManifest</code> defined
-      via a file included in <code>importScripts</code>.
+      via a file included in <code>importScripts</code>. In other words, developers using the
+      <code>workbox-webpack-plugin</code> do not need to set this option unless there are additional
+      files not included in your webpack bundles that you need to precache.
     </p>
     <p>
       If you do set this, make sure to also configure <code>globPatterns</code>.
@@ -44,7 +46,9 @@ globDirectory: '.'</pre>
     <p>
       This only needs to be set if you want to precache local files matching <code>glob</code>
       patterns. It does not need to be set if you have <code>self.__precacheManifest</code> defined
-      via a file included in <code>importScripts</code>.
+      via a file included in <code>importScripts</code>. In other words, developers using the
+      <code>workbox-webpack-plugin</code> do not need to set this option unless there are additional
+      files not included in your webpack bundles that you need to precache.
     </p>
     <p>
       If you do set this, make sure to also configure <code>globDirectory</code>.
@@ -76,12 +80,14 @@ globDirectory: '.'</pre>
       those entries will be automatically precached in the generated service worker.
     </p>
     <p>
-      This is also useful when you want to let Workbox create your top-level service worker file,
-      but want to include some additional code, such as a <code>push</code> event listener.
+      If you are using the <code>workbox-webpack-plugin</code> then this option will be
+      automatically set on your behalf, configured to import both the precache manifest and the
+      Workbox runtime code. You can list additional scripts, like code to add a
+      <code>push</code> event listener, that should be imported as well.
     </p>
     <p>
       <strong>Example:</strong>
     </p>
-    <pre class="prettyprint">importScripts: ['precache-manifest.abcd1234.js']</pre>
+    <pre class="prettyprint">importScripts: ['push-listener.js']</pre>
   </td>
 </tr>
--- a/src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
+++ b/src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
@@ -11,13 +11,6 @@
       current working directory.
     </p>
     <p>
-      This only needs to be set if you want to precache local files matching <code>glob</code>
-      patterns. It does not need to be set if you have <code>self.__precacheManifest</code> defined
-      via a file included in <code>importScripts</code>. In other words, developers using the
-      <code>workbox-webpack-plugin</code> do not need to set this option unless there are additional
-      files not included in your webpack bundles that you need to precache.
-    </p>
-    <p>
       If you do set this, make sure to also configure <code>globPatterns</code>.
     </p>
     <p>
@@ -42,13 +35,6 @@ globDirectory: '.'</pre>
     <p>
       For more information, see the
       <a href="https://github.com/isaacs/node-glob#glob-primer" class="external">glob primer</a>.
-    </p>
-    <p>
-      This only needs to be set if you want to precache local files matching <code>glob</code>
-      patterns. It does not need to be set if you have <code>self.__precacheManifest</code> defined
-      via a file included in <code>importScripts</code>. In other words, developers using the
-      <code>workbox-webpack-plugin</code> do not need to set this option unless there are additional
-      files not included in your webpack bundles that you need to precache.
     </p>
     <p>
       If you do set this, make sure to also configure <code>globDirectory</code>.
@@ -80,14 +66,12 @@ globDirectory: '.'</pre>
       those entries will be automatically precached in the generated service worker.
     </p>
     <p>
-      If you are using the <code>workbox-webpack-plugin</code> then this option will be
-      automatically set on your behalf, configured to import both the precache manifest and the
-      Workbox runtime code. You can list additional scripts, like code to add a
-      <code>push</code> event listener, that should be imported as well.
+      This is also useful when you want to let Workbox create your top-level service worker file,
+      but want to include some additional code, such as a <code>push</code> event listener.
     </p>
     <p>
       <strong>Example:</strong>
     </p>
-    <pre class="prettyprint">importScripts: ['push-listener.js']</pre>
+    <pre class="prettyprint">importScripts: ['push-notifications.abcd1234.js']</pre>
   </td>
 </tr>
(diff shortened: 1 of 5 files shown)


src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
```diff
@@ -0,0 +1,82 @@
...
+to your file. It will otherwise leave the file as-is. This is referred to as `injectManifest` mode.
+
+{% include "web/tools/workbox/_shared/when-to-use-inject-manifest.md" %}
+
+### Do you need to cache additional, non-webpack assets?
+
+By default, the plugin will generate a precache manifest that contains URLs for all of the assets
+that come from the current webpack compilation. Any assets that webpack doesn't "know" about will
+not be picked up.
+
```
johnnyreilly: So if you'd copied something into the same output location as that which webpack uses then this would **not** be included.  i.e. You would need to use `globDirectory` and `globPatterns` in this case.  Not otherwise.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
```diff
@@ -0,0 +1,82 @@
...
+- The [`glob` pattern matching](https://github.com/isaacs/node-glob#glob-primer) will be performed
+against the local file system, and the directory that `globDirectory` is set to must exist at the
+time the plugin runs. This might run afoul of
+[`webpack-dev-server` configurations](https://github.com/webpack/webpack-dev-server), where an
+in-memory file system is used.
+- The options that work with `glob`-based precache manifests, like `manifestTranforms` and
+`modifyUrlPrefix`, can also be used, but they'll apply **only** to the entries that are matched via
+`glob` patterns, and not to any assets that are picked up via the webpack compilation.
+
```
johnnyreilly: It's great that this information is here.   However it would be good if the workbox docs for `globDirectory` and `globPatterns` explicitly state that they are an "escape hatch" rather than something that you should need to configure.  Left to my own devices I had assumed there was a default `globPattern` of `'**/*.*` or something.  Not so - but I didn't realise.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
```diff
@@ -0,0 +1,82 @@
...
+This is a plugin for the [webpack](https://webpack.js.org/) which generates a list of URLs to
+[precache](workbox-precaching) (known as a "precache manifest") based on the assets in a webpack
+compilation.
+
+## Considerations
+
+There are a few decisions to make which will determine the right configuration for your use case.
+
+### Should the plugin create your service worker file?
```
jeffposnick: This documentation describes what's currently implemented as of v3-alpha.3, but it feels clunky. See https://github.com/GoogleChrome/workbox/issues/1136 to track decisions about whether we should change this interface.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
jeffposnick: Right.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
jeffposnick: There's currently the following disclaimer:

> This only needs to be set if you want to precache local files matching <code>glob</code> patterns. It does not need to be set if you have <code>self.__precacheManifest</code> defined via a file included in <code>importScripts</code>.

But I'll make it clearer that developers who are using the webpack plugin are the ones who wouldn't need to set them.

jeffposnick:
Thanks for the feedback, @johnnyreilly! @gauntface, how do you feel about this PR?

src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
```diff
@@ -13,7 +13,9 @@
     <p>
       This only needs to be set if you want to precache local files matching <code>glob</code>
       patterns. It does not need to be set if you have <code>self.__precacheManifest</code> defined
-      via a file included in <code>importScripts</code>.
+      via a file included in <code>importScripts</code>. In other words, developers using the
```
gauntface: I find this section confusing and I feel it could be more direct.

"This option sets the base you wish to match `globPatterns` against, relative to the work directory. If you don't want to precache files then this configuraiton isn't needed."

Regarding this new piece, I would assume this would feed into self.__precacheManifest (i.e. the files would be added to that list, if thats the case then why isn't being mentioned here as a reason not to use it? and if it's not added to that list, why not?

src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
```diff
@@ -44,7 +46,9 @@
...
```
gauntface: Are all of these going to be added to webpack plugin *and* workbox build and CLI? If so, could we find some way to moving this info into a condition template that is only displayed for the webpack plugin?

Also feeling uneasy that we need a "In other words" rather than just stating what is easiest to understand.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
```diff
@@ -0,0 +1,86 @@
...
+
+{# wf_updated_on: 2017-12-15 #}
+{# wf_published_on: 2017-12-15 #}
+{# wf_blink_components: Blink>ServiceWorker #}
+
+# Workbox webpack Plugin  {: .page-title }
+
+## What's the Workbox webpack Plugin?
+
+This is a plugin for the [webpack](https://webpack.js.org/) which generates a list of URLs to
```
gauntface: Remove 'the' from "plugin for the [webpack]"

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
```diff
@@ -0,0 +1,86 @@
...
+[precache](workbox-precaching) (known as a "precache manifest") based on the assets in a webpack
```
gauntface: I wouldn't link this up to workbox-precaching. I would expect this to go to a page that is about what the term "precache means" rather than a guide on another module.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
```diff
@@ -0,0 +1,86 @@
...
+
+If you need to precache additional assets that are managed outside of webpack, then you can
+use the `globDirectory` and `globPatterns` options to specify how to find those additional assets.
+
+If you decide to use `globDirectory` and `globPatterns`, the following applies:
+
+- The [`glob` pattern matching](https://github.com/isaacs/node-glob#glob-primer) will be performed
+against the local file system, and the directory that `globDirectory` is set to must exist at the
+time the plugin runs. This might run afoul of
+[`webpack-dev-server` configurations](https://github.com/webpack/webpack-dev-server), where an
```
gauntface: I might move this negative out into a "Known Issues" section at the bottom of the page.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
```diff
@@ -0,0 +1,86 @@
...
+in-memory file system is used.
+- The options that work with `glob`-based precache manifests, like `manifestTranforms` and
+`modifyUrlPrefix`, can also be used, but they'll apply **only** to the entries that are matched via
+`glob` patterns, and not to any assets that are picked up via the webpack compilation.
+
+## Configuration
+
+Once you've determined the answers to those considerations, you can pass the appropriate
```
gauntface: I was expecting the two modes description to then be followed by two distinct sections, one for generate and one for manifest injection and then both have their own sets of code examples. But this may be something that is easier and more obvious with the changes mentioned in https://github.com/GoogleChrome/workbox/issues/1136

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
jeffposnick: I've changed the link to point to `/web/tools/workbox/guides/precache-files`.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
jeffposnick: Yes, assuming GoogleChrome/workbox#1136 is resolved this would end up being two sections for the different modes.

src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
jeffposnick: We don't expose `generateSWString` via the CLI.

We do expose it via `workbox-build` and yes, it would show up in those docs.

I can change the conditional language by creating a standalone `_shared/webpack-options.html` instead of including it in this file.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
jeffposnick: Done.

src/content/en/tools/workbox/modules/workbox-webpack-plugin.md
jeffposnick: I'd prefer keeping it in this section alongside the other guidance about when developers might want to use it.

src/content/en/tools/workbox/guides/_shared/generate-sw-string-schema.html
jeffposnick: I've moved this to its own webpack-specific include file, so the language can be more precise.

It now reads "This only needs to be set if you want to precache additional local files that are not created by the webpack build process."

petele:
@jeffposnick - if you're happy, feel free to merge!

Issue #5535, opened 2017-12-18
Refresh recipe can cause the page to refresh immediately on first load


dfabulich:
**Page Affected:** https://developers.google.com/web/tools/workbox/guides/advanced-recipes#offer_a_page_reload_for_users

To reproduce, look at https://github.com/dfabulich/service-worker-refresh-sample which copies and pastes from the advanced recipes guide. Check out the `clients-claim` branch.

1. `npm install && npm run start`
2. In an incognito window, open dev tools, enable Preserve Log, and load `http://localhost:8000`

Actual: The page will load, then immediately log "Controller loaded" and refresh the page, forcing the page to load twice.
Expected: We should only refresh the page when the user clicked a button to activate a waiting worker.

**What needs to be done?**

We should only listen for `controllerchange` if the page already has a controller. (I'll file a PR.)

PR #5536, opened 2017-12-18
Prevent refreshing on initial install

dfabulich:
**Fixes:** #5535

**R:** @petele @gauntface

### DIFF
--- a/src/content/en/tools/workbox/guides/advanced-recipes.md
+++ b/src/content/en/tools/workbox/guides/advanced-recipes.md
@@ -73,7 +73,12 @@ function onNewServiceWorker(registration, callback) {
 
 window.addEventListener('load', function() {
   // When the user asks to refresh the UI, we'll need to reload the window
+  var refreshing;
   navigator.serviceWorker.addEventListener('controllerchange', function(event) {
+    // Ensure refresh is only called once.
+    // This works around a bug in "force update on reload".
+    if (refreshing) return;
+    refreshing = true;
     console.log('Controller loaded');
     window.location.reload();
   });
--- a/src/content/en/tools/workbox/guides/advanced-recipes.md
+++ b/src/content/en/tools/workbox/guides/advanced-recipes.md
@@ -2,7 +2,7 @@ project_path: /web/tools/workbox/_project.yaml
 book_path: /web/tools/workbox/_book.yaml
 description: Advanced recipes to use with Workbox.
 
-{# wf_updated_on: 2017-12-17 #}
+{# wf_updated_on: 2017-12-19 #}
 {# wf_published_on: 2017-12-17 #}
 {# wf_blink_components: N/A #}
 
@@ -72,17 +72,6 @@ function onNewServiceWorker(registration, callback) {
 }
 
 window.addEventListener('load', function() {
-  // When the user asks to refresh the UI, we'll need to reload the window
-  var refreshing;
-  navigator.serviceWorker.addEventListener('controllerchange', function(event) {
-    // Ensure refresh is only called once.
-    // This works around a bug in "force update on reload".
-    if (refreshing) return;
-    refreshing = true;
-    console.log('Controller loaded');
-    window.location.reload();
-  });
-
   navigator.serviceWorker.register('/sw.js')
   .then(function (registration) {
       // Track updates to the Service Worker.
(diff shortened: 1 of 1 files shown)


dfabulich:
This PR is built on top of #5535; if there's a problem with that, I can rebase it.

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -87,6 +81,17 @@ window.addEventListener('load', function() {
       return;
     }
 
+    // When the user asks to refresh the UI, we'll need to reload the window
+    var refreshing;
+    navigator.serviceWorker.addEventListener('controllerchange', function(event) {
+      // Ensure refresh is only called once.
+      // This works around a bug in "force update on reload".
+      if (refreshing) return;
```
gauntface: Can you rename `refreshing` to `window.__preventDevToolsReloadLoop`, this just makes it clear what the boolean is doing and the double underscore just makes it clear that it's a custom property.

dfabulich: Can you say a little more about why? This local `refreshing` variable doesn't need to be shared globally; generally speaking, it's better to use local variables instead of putting stuff on the global `window` object.

petele:
@dfabulich - please also update the `wf_updated_on` tag at the top of the doc.

@gauntface - when you're happy, please merge.

src/content/en/tools/workbox/guides/advanced-recipes.md
gauntface: At the moment it's ambiguous what you are trying to assign this to. I don't mind if assigning it to the function works, just assign it to something.

gauntface:
From the other PR - your comments:

I think you might be thinking of a different PR. This PR just adds a local refreshing variable to prevent each tab from double-reloading. It should not have any different effect on background tabs.

I have a sample repository https://github.com/dfabulich/service-worker-refresh-sample that copies and pastes from the Advanced recipe guide; I just pushed a dev-tools-workaround branch incorporating the proposed fix in this PR.

Here's what I did:

npm install && npm run start
Opened http://localhost:8000 in an incognito window
Refreshed to make sure the client is claimed
Opened a second tab to http://localhost:8000
Edited index.html to replace the "0" in the body tag on line 2 with a "1", and edited index.html to change the first line v variable from "0" to "1".
Refreshed one browser tab. The reload prompt button appeared.
I clicked the button. The page refreshed and I saw a "1" on the screen.
I switched back to the other tab. That tab had also refreshed, and was now also showing a "1", as expected.
Are you able to find a multiple-tab bug in my sample repro? If so, can you tell me how to reproduce that bug?

I should point out that step 3 in that list is subtle and important, because my sample doesn't call clients.claim() during activation.

So if you skip step 3 in the first tab, the only controlled client will be the second tab; in that case, if you click the button in step 7, and then switch back to the first tab in step 8, well, it never had a controller and still doesn't, so it doesn't refresh.

That's true regardless in both branches of my sample, before and after the change proposed in this PR.

EDIT: It turns out that if the SW does clients.claim() during activation, that causes issue #5535; I've put together a PR #5536 to fix that.

src/content/en/tools/workbox/guides/advanced-recipes.md
dfabulich: > it's ambiguous what you are trying to assign this to

I think we may be miscommunicating. I'm not "trying to assign" `refreshing` at all at line 85.

Specifically, I think you may have incorrectly assumed that `var refreshing` on line 85 is assigning refreshing to something, either as a property of the `window` or as a property of a function.

That's not what `var` does here. `var refreshing` defines a local variable, scoped to current function; it's not an assigned property of the window, the function, or anything else. As a local variable, it's only visible inside this function, which is much safer/better than assigning it somewhere visible (where it could clobber someone's existing variable).

For example, even if there _were_ a `window.refreshing` variable, `var refreshing` here in this function would shadow the `refreshing` variable on the global scope, automatically preventing a naming conflict.

In an effort to meet you halfway, I've renamed the variable, so now it says `var preventDevToolsReloadLoop`, but with no underscores and no explicit assignment, ensuring that it can't conflict with global window properties.

I hope this explanation makes sense and the updated name will be agreeable. LMK if this didn't make sense.

PR #5554, opened 2017-12-21
`oldVersion` is not a method of UpgradeDB object

kyarik:

`oldVersion` is simply a property with a number value.

**R:** @petele

### DIFF
--- a/src/content/en/ilt/pwa/working-with-indexeddb.md
+++ b/src/content/en/ilt/pwa/working-with-indexeddb.md
@@ -547,7 +547,7 @@ Note: The browser throws an error if we try to create object stores or indexes t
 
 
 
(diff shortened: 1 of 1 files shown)


kyarik:
Looks like on this article `wf_updated_on` has format `yyyy-dd-mm` instead of `yyyy-mm-dd` as with other articles? Since right now it's `2017-07-12`, and in the articles it says `Last updated December 7, 2017`.

petele:
@DavidScales & @NaSearle - this is in the PWA ILT content, can you take a peek and port into your Google doc versions.

petele:
@kyarik - it should be YYYY-MM-DD. These files are auto-generated from Google Doc source files, so there may be a problem there. I'll investigate after the holiday.

DavidScales:
@kyarik & @petele - apologies for the slow response (out for the holidays). I have incorporated these changes into the source doc. Thanks for making these fixes! 👍

petele:

PR #5557, opened 2017-12-21
`transaction` method only takes object store names

kyarik:
In fact, if `IDBDatabase.transaction` would accept names of object stores and indexes, how would it distinguish a store name from an index name? Even if it could distinguish them, since different object stores can have indexes with the same name, how would it know which index you mean?

**R:** @petele


petele:
@DavidScales & @NaSearle - this is in the PWA ILT content, can you take a peek and port into your Google doc versions.

DavidScales:
@kyarik - yeah that makes sense. I think we must have been confused by the [MDN documentation for transaction()](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/transaction), which says that Indexes can be used in the `storeNames` argument. I'll try to update that documentation or file an issue. 

Thanks again for these changes! :)

@petele - I have incorporated these changes into the source doc, thanks!

petele:

PR #5559, opened 2017-12-21
Fix imprecision about IDBKeyRange API

kyarik:
To specify exclusion we need to pass `true` as the second argument,
since that argument indicates whether the bound is open, and defaults to `false`.
Also, `IDBKeyRange.bound` takes 4 arguments, not 3, since we need to
specify both whether the lower bound is open and whether the upper bound
is open.

**R:** @petele

### DIFF
--- a/src/content/en/ilt/pwa/working-with-indexeddb.md
+++ b/src/content/en/ilt/pwa/working-with-indexeddb.md
@@ -494,7 +494,7 @@ The `bound` method is used to specify both an upper and lower limit, and takes t
 IDBKeyRange.bound(lowerIndexKey, upperIndexKey);
 ```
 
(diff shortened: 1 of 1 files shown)


petele:
@DavidScales & @nasearle - this is in the PWA ILT content, can you take a peek and port into your Google doc versions.

petele:
@kyarik - the 'source of truth' for any of our PWA ILT content is actually a Google doc. I've looped in the right folks who can fix there. Once it's fixed there and we export, I'll close this PR since the fixes will go into the export/update PR. Thanks for calling it out & fixing!

DavidScales:
@kyarik & @petele - I have incorporated these changes into the source doc. Thanks again!

petele:

PR #5602, opened 2018-01-04
Getting started guide was incorrect

gauntface:
Fixes https://github.com/GoogleChrome/workbox/issues/1153

The getting started guide was using the old way of setting up expiration that is no longer supported.


petele:
@gauntface - I wouldn't be opposed to you adding an exclude for `wf_blink_components` in workbox docs. Should be relatively easy, we already do it for a couple of file types.

PR #5643, opened 2018-01-11
Fix indentation in list of examples

tamalet:
The extra line causes the text to move a new paragraph.

**CC:** @petele


tamalet:
I signed it!

Issue #5541, opened 2017-12-19
auto-updated on


gauntface:
If I write the function to manage this, can I just automatically update the wf_updated_on in husky?

This would be based on the git diff logic we already have in gulp tests.

petele:
Sounds awesome!

gauntface:
Logic is here: https://github.com/google/WebFundamentals/pull/5681/commits/3310d84d6e755645ea18e826e861197fd71649a9#diff-c5f9b992fa42296ad4a79bcc0c58861d

PR #5692, opened 2018-01-22
Auto update wf_updated_on field

petele:
**Replaced by #5699**

- Pre-commit hook that will update `wf_updated_on` if it wasn't already updated.

**Fixes:** #5541 

- [ ] This has been reviewed and approved by (@gauntface)

**CC:** @petele


gauntface:
Firebase JS SDK commits the changes from Prettier: https://github.com/firebase/firebase-js-sdk/blob/master/tools/gitHooks/prettier.js

We should try and copy that instead of throwing an error.

PR #5697, opened 2018-01-24
Add known bugs to play promise

beaufortfrancois:
R: @mounirlamouri
BUG: https://bugs.chromium.org/p/chromium/issues/detail?id=718647

### DIFF
--- a/src/content/en/updates/2016/03/play-returns-promise.md
+++ b/src/content/en/updates/2016/03/play-returns-promise.md
@@ -71,7 +71,7 @@ of this new functionality. View it in a browser such as Chrome 50 that supports 
 interface. Be forewarned: the page will automatically play music when you visit it. (Unless, of
 course, it doesn’t!)
 
-## Relevant bugs {: #bugs }
+## Danger zone {: #danger-zone }
 
 ### `<source>` within `<video>` makes `play()` promise never rejects
 
--- a/src/content/en/updates/2016/03/play-returns-promise.md
+++ b/src/content/en/updates/2016/03/play-returns-promise.md
@@ -2,7 +2,7 @@ project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: Say goodbye to automatic playback uncertainty! play() now returns a Promise.
 
-{# wf_updated_on: 2016-03-11 #}
+{# wf_updated_on: 2018-01-24 #}
 {# wf_published_on: 2016-03-11 #}
 {# wf_tags: promises,play,video,audio,chrome50 #}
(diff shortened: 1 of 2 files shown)


src/content/en/updates/2016/03/play-returns-promise.md
```diff
@@ -71,5 +71,15 @@ of this new functionality. View it in a browser such as Chrome 50 that supports
 interface. Be forewarned: the page will automatically play music when you visit it. (Unless, of
 course, it doesn’t!)
 
+## Relevant bugs {: #bugs }
```
mounirlamouri: I wouldn't call this a bug as it's actually part of the spec and there isn't much we can do. The bug will most likely be closed after this article is updated.

src/content/en/updates/2016/03/play-returns-promise.md
```diff
@@ -71,5 +71,15 @@ of this new functionality. View it in a browser such as Chrome 50 that supports
...
+
+### `<source>` within `<video>` makes `play()` promise never rejects
+
+For `<video src="not-existing-video.mp4"\>`, the `play()` promise rejects as
+expected as the video doesn't exist. For `<video><source
+src="not-existing-video.mp4" type='video/mp4'></video>`, the `play()` promise
+never rejects. It only happens if there are no valid sources.
```
mounirlamouri: If we don't want to surface this as a bug, we might want to point to the spec?

src/content/en/updates/2017/06/play-request-was-interrupted.md
```diff
@@ -2,7 +2,7 @@ project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: What is really happening with "DOMException: The play() request was interrupted"?
 
-{# wf_updated_on: 2017-06-16 #}
+{# wf_updated_on: 2018-01-24 #}
 {# wf_published_on: 2017-06-14 #}
```
mounirlamouri: All comments above apply to this file.

beaufortfrancois:
@mounirlamouri I've renamed it to "Danger zone"

mounirlamouri:
Do you think it makes sense to leave the bugs as explicitly? I wonder if readers may still assume that it's a bug and will be fixed.

beaufortfrancois:
I believe that is a way for us to see how many people will stumble upon this and might add their voice so that we can eventually "fix" the spec if needed.

mounirlamouri:
I'm not sure we can just "fix" the spec but that's really your editorial call so LGTM :)

beaufortfrancois:
@petele Can you merge and push live?

petele:
will be pushed live tomorrow

Issue #5780, opened 2018-02-10
RSS feed is broken


nazar-pc:
**Page Affected:** https://developers.google.com/web/updates/rss.xml

**What needs to be done?**
RSS should be valid XML document, otherwise Mozilla Thunderbird doesn't want to read it: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fdevelopers.google.com%2Fweb%2Fupdates%2Frss.xml

This happened a while ago when you've silently changed URL and type of the feed from Atom to RSS.

nazar-pc:
Looks like Atom feed is also back, but it is also broken in the same way

petele:
There was an errant `FF` character in one of the files that broke the feed. I've checked the feed and removing that fixes most stuff. There are still some warnings, but the critical errors have been fixed.

If you still want the ATOM version, you can get it at https://developers.google.com/web/updates/atom.xml

nazar-pc:
I still see `This feed does not validate.` for both feeds and Thunderbird still can't use it.

nazar-pc:
I see that PR is not yet merged, will wait for it and check again

petele:
Merged, and pushing live now, PR will get a comment when it's live on DevSite

nazar-pc:
Works great, thanks!

PR #5782, opened 2018-02-12
Remove errant character that breaks RSS feed output

petele:
- Remove errant `FF` in file that broke RSS feed
- item 2

**Fixes:** #5780 

**Target Live Date:** ASAP


petele:
thanks @ebidel

PR #5785, opened 2018-02-12
Fix incorrect title on Japanese Keyword Hack doc

petele:
- Fixed title

**Target Live Date:** ASAP


petele:
thanks @kaycebasques

PR #5788, opened 2018-02-13
Refactored tests in separate files

petele:
- Moved most tests to individual files, except for `testMarkdown`, that one will likely need to happen in a separate PR as it's pretty big and each time I've tried I've run into all kinds of fun errors. 

**CC:** @gauntface


petele:
> Would make sense to just have a top level tests directory rather than under gulp-tasks/tests/

Open to that, is that common practice or is any specific reason for doing so?

PR #5810, opened 2018-02-21
[lightouse] update template & add "font sizes" reference

kaycebasques:
Hey @rviscomi and @kdzwinel, have a look at the reference for the "font size" audit, plz!

- update template for LH reference docs
- add reference for "font sizes" audit

**Fixes:** N/A

**Target Live Date:** 2018-02-28

- [x] This has been reviewed and approved by @kdzwinel & @rviscomi 

**CC:** @petele

### DIFF
--- a/src/content/en/tools/lighthouse/_toc.yaml
+++ b/src/content/en/tools/lighthouse/_toc.yaml
@@ -48,6 +48,8 @@ toc:
   path: /web/tools/lighthouse/audits/aspect-ratio
 - title: Document Does Not Have A Meta Description
   path: /web/tools/lighthouse/audits/description
+- title: Document Doesn't Use Legible Font Sizes
+  path: /web/tools/lighthouse/audits/font-sizes
 - title: Element aria-* Attributes Are Allowed For This Role
   path: /web/tools/lighthouse/audits/aria-allowed-attributes
 - title: Element aria-* Attributes Are Valid And Not Misspelled Or Non-Existent
--- a/src/content/en/tools/lighthouse/audits/font-sizes.md
+++ b/src/content/en/tools/lighthouse/audits/font-sizes.md
@@ -1,0 +1,47 @@
+project_path: /web/tools/_project.yaml
+book_path: /web/tools/_book.yaml
+description: Reference documentation for the "Document Doesn't Use Legible Font Sizes" Lighthouse audit.
+
+{# wf_updated_on: 2018-02-21 #}
+{# wf_published_on: 2018-02-21 #}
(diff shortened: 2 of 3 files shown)


src/content/en/tools/lighthouse/audits/font-sizes.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Text is illegible because of a missing viewport config {: #viewport }
+
+If Lighthouse reports `Text is illegible because of a missing viewport config`, add a
+`<meta name="viewport" content="width=device-width, initial-scale=1">` tag to the `<head>` of
+your document. See [Has A Viewport Meta Tag With Width Or Intial-Scale][viewport].
+
+[viewport]: /web/tools/lighthouse/audits/has-viewport-meta-tag
+
+### Characters per line {: #characters }
```
rviscomi: This guidance isn't directly relevant to the audit so it's ok to drop it. The current "Learn more" link only points here because the PSI docs on font sizes disappeared and redirected here.

src/content/en/tools/lighthouse/_toc.yaml
```diff
@@ -48,6 +48,8 @@ toc:
   path: /web/tools/lighthouse/audits/aspect-ratio
 - title: Document Does Not Have A Meta Description
   path: /web/tools/lighthouse/audits/description
+- title: Document Doesn't Use Legible Font Sizes
```
rviscomi: Should we update the audit config to use "Does Not" for consistency with other audits?

kaycebasques: One way or the other, I was thinking about putting a PR in to make it consistent across reports

kaycebasques: Contracting it makes it feel a little more impersonal, which seems more in line with LH's voice

src/content/en/tools/lighthouse/audits/font-sizes.md
kaycebasques: Yeah, I get what you're saying. What I was thinking is that it's nice to use this reference as an entry point to related topics that the developer might be interested in, but the tradeoff is that it clutters up the reference with less-relevant content. Something to consider in general.

cc @paulirish & @vinamratasingal

paulirish: yeah line length is technically a readability concern and separate from the legibility of small glyphs. 

but i think we should let prescriptive docs speak for themselves. and just link em without further elaboration.

PR #5841, opened 2018-02-28
Adding FAQ with scope info

gauntface:
Fixes https://github.com/GoogleChrome/workbox/issues/1211

First attempt to cover scope issues with SW


src/content/en/tools/workbox/guides/faq.md
```diff
@@ -0,0 +1,104 @@
...
+navigator.serviceWorker.register(‘/blog/sw.js’, {scope: ‘/blog/article/’});
+
+// Not OK because the scope is shorter than the path /blog/
+navigator.serviceWorker.register(‘/blog/sw.js’, {scope: ‘/’});
+
+// Not OK because the scope doesn't start with /blog/
+navigator.serviceWorker.register(‘/blog/sw.js’, {scope: ‘/about/’});
+```
+
+There is however a way around this limitation, if the web server that
```
jeffposnick: The `,` between "limitation" and "if" should be something else, like a `.` or `:`.

src/content/en/tools/workbox/guides/faq.md
```diff
@@ -0,0 +1,104 @@
...
+serves the service worker file includes a ‘Service-Worker-Allowed’
```
jeffposnick: Can you make it explicit that the header needs to be added to the response for *the service worker JS file*?

Otherwise, developers might thing that the header needs to be added to the response for the current HTML page that they're on in order to "enable" this mode.

src/content/en/tools/workbox/guides/faq.md
```diff
@@ -0,0 +1,104 @@
...
+## Why isn't my service worker being used? {#sw-scope}
+
+This is a fairly common and easy problem to hit when working with service
+workers.
+
+You'll register the service worker, everything seems to work and files are
+cached, but refreshing the page or going offline doesn't show your files
+(i.e. the browser isn't using the files in the cache).
+
+The reason this occurs is that the "scope" of your service worker doesn't
```
jeffposnick: You'd see similar symptoms for the first navigation when you are not using `clientsClaim` (and sometimes when you aren't using `skipWaiting`, if you had a previous service worker open in other tabs).

I like the material that already exists about this at https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#scope_and_control, so I think it could just be linked to here rather than trying to replicate it. But it would be a good idea to acknowledge that there are other possible scenarios to investigate when you're not seeing the right SW in control.

src/content/en/tools/workbox/guides/faq.md
```diff
@@ -0,0 +1,104 @@
+project_path: /web/tools/workbox/_project.yaml
+book_path: /web/tools/workbox/_book.yaml
+description: Frequently Asked Questions when developing with Workbox
+
+{# wf_updated_on: 2018-02-28 #}
+{# wf_published_on: 2018-02-28 #}
+{# wf_blink_components: N/A #}
+
+# FAQ {: .page-title }
```
jeffposnick: In the interest of fighting "doc creep", where we end up with many separate guides/docs all sitting side by side in the ToC: what's the criteria for adding future information into the "FAQ" guide? We have an existing "Troubleshoot and Debug" guide, for instance—when do we add things to that doc vs. adding it to this FAQ?

src/content/en/tools/workbox/guides/faq.md
gauntface: I rejigged things and reworded so it highlights *it's one reason this occurs* and I moved how to identify *if* this is the problem early on.

src/content/en/tools/workbox/guides/faq.md
gauntface: Done.

src/content/en/tools/workbox/guides/faq.md
gauntface: done.

gauntface:
@jeffposnick I just moved it into troubleshooting and removed a tonne of stuff (i.e. just move to root and linked to the url you posted).

src/content/en/tools/workbox/guides/_toc.yaml
```diff
@@ -44,3 +44,5 @@ toc:
   path: /web/tools/workbox/guides/service-worker-checklist
 - title: "Advanced Recipes"
   path: /web/tools/workbox/guides/advanced-recipes
+- title: "FAQ"
```
jeffposnick: You can drop this TOC entry.

src/content/en/tools/workbox/guides/troubleshoot-and-debug.md
```diff
@@ -59,10 +59,45 @@ working on your service worker files.
...
+worker like so:
+
+```javascript
+// OK when ‘Service-Worker-Allowed’ header is set to ‘/’
+navigator.serviceWorker.register(‘/blog/sw.js’, {scope: ‘/’});
+```
+
+You can [learn more about service worker scope here]('/web/fundamentals/primers/service-workers/lifecycle#scope_and_control').
+
+**_Q:_** Why are changes to my service worker not shows?.
```
jeffposnick: "shows" should be "shown"

jeffposnick:
Thanks, Matt.

PR #5837, opened 2018-02-28
Loading Third-party JavaScript

addyosmani:
What's changed, or what was fixed?
- This PR introduces a new article on efficiently loading third-party JS. 
**Fixes:** #5669

**Target Live Date:** 2018-03-02

- [X] This has been reviewed and approved by Speed DevRel, Kenji - the PM for Loading and Pat Meenan.
- [x] I have run `gulp test` locally and all tests pass.
- [X] I have added the appropriate `type-something` label.
- [X] I've staged the site and manually verified that my content displays correctly.

**CC:** @petele

This requires a little more work on linting and adding in alt tags for all images. I'll continue working through these bits shortly.

PR #5844, opened 2018-03-01
Fix for loading third-party JS in TOC

addyosmani:

As part of #5837, I neglected to include the step group and correct TOC entry for the "Loading third-party JavaScript" article. This change adds the article to the TOC correctly.

**Target Live Date:** 2018-03-01

- [ ] This has been reviewed and approved by (addyo, minor change)
- [X] I have added the appropriate `type-something` label. (I think this is content..might be infra)

**CC:** @petele

### DIFF
--- a/src/content/en/fundamentals/performance/_toc.yaml
+++ b/src/content/en/fundamentals/performance/_toc.yaml
@@ -51,6 +51,10 @@ toc:
           step_group: perf-oce
         - title: JavaScript Start-up Optimization
           path: /web/fundamentals/performance/optimizing-content-efficiency/javascript-startup-optimization
+          step_group: perf-oce
+        - title: Loading Third-Party JavaScript
+          path: /web/fundamentals/performance/optimizing-content-efficiency/loading-third-party-javascript
+          step_group: perf-oce
         - title: Web Font Optimization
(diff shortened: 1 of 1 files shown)


robdodson:

robdodson:
Is this change supposed to make it so the side-nav stays open on the left, and the current article is highlighted? It didn't seem to do that when I checked.

Compare this post by Ilya in the same section:

image

To these other posts:

image

image

Entirely possible I'm doing something silly 🤠

addyosmani:
@robdodson It's not you at all. I'm seeing the same behavior locally (and yes it should be keeping the navigation open). I'm using the same format for TOC entries in the optimizing content efficiency section as Ilya was afaik:

```
        - title: Optimizing Encoding and Transfer Size of Text-based Assets
          path: /web/fundamentals/performance/optimizing-content-efficiency/optimize-encoding-and-transfer
          step_group: perf-oce
```

Might require some further investigation 🙃

petele:
Short version: you put the URL into the TOC without the last forward slash, so it thought it was a file, not a directory.

Longer version: our staging env, the left nav is opened/highlighted in JavaScript, it matches the current address bar URL with the one specified in the link. Since the paths didn't match (dir vs file), it didn't expand anything. I'll fix with a direct commit to master.

PR #5857, opened 2018-03-02
 [lighthouse] add reference for indexing audit

kaycebasques:
- added reference for indexing audit
- moved a devtools image to the shared location

**Fixes:** N/A

**Target Live Date:** 2018-03-05

- [ ] This has been reviewed and approved by @rviscomi 

**CC:** @petele


src/content/en/tools/lighthouse/audits/indexing.md
```diff
@@ -0,0 +1,59 @@
...
+{# wf_published_on: 2018-03-02 #}
+{# wf_blink_components: Platform>DevTools #}
+
+# Page Is Blocked From Indexing  {: .page-title }
+
+## Overview {: #overview }
+
+Search engines crawl your page in order to understand your content. If you don't
+give them permission to crawl your page, they can't know what your page
+contains, and therefore are less likely to list your pages in search results.
```
rviscomi: Why "less likely" and not something stronger like "unable"? AFAIK a page that's not indexable will never appear in search.

src/content/en/tools/lighthouse/audits/indexing.md
```diff
@@ -0,0 +1,59 @@
...
+overview of the most common issues.
+
+[spec]: https://developers.google.com/search/reference/robots_meta_tag
+
+The most common problematic meta tag is:
+
+    <meta name="robots" content="noindex"/>
+
+This tag prevents all search engine crawlers from accessing your page. You might
+have meta tags that block specific crawlers, such as:
```
rviscomi: It may be worth mentioning that LH ignores these bot-specific directives and only evaluates the common "robots" directives.

src/content/en/tools/lighthouse/audits/indexing.md
kaycebasques: This passage confused me. I thought it was still possible, just more difficult. But the bottom of the passage confirms what you're saying.

> Using the robots.txt protocol on your site can stop Google from crawling your pages, but it may not always prevent them from being indexed. For example, Google may index your page if we discover it by following a link from someone else's site. To display it in search results, Google will need to display a title of some kind and because we won't have access to any of your page content, we will rely on off-page content such as anchor text from other sites. (To truly block a URL from being indexed, you can use meta tags.)

https://support.google.com/webmasters/answer/35624?hl=en

PR #5858, opened 2018-03-04
Adding "Why Performance Matters" article

malchata:
This went through feedback and I've addressed all points (but am willing to go through further rounds, if need be). This seemed like the right time to get this in the PR process. This article can be viewed locally at http://localhost:8080/web/fundamentals/performance/why-performance-matters/

Things that are still TODO:
- Need to add the article to the ToC. Haven't done this yet because...
- There's some talk of relocating the [RAIL article](https://developers.google.com/web/fundamentals/performance/rail) and having this new article as the overview piece. May need guidance on that.
- I'm sure there's other stuff that will come up.

cc: @addyosmani @petele


addyosmani:
Some notes from our call today:

* We'll reference RAIL in this doc
* TBD whether this will be the new overview page. I'll make a call on this today.
* Jeremy will review if there's any further text we can trim/tweak in the current draft. We discussed that at some point post-launch we might break this down into two docs.

addyosmani:
Locally, it appears that a portion of the start of the article is getting cut off:

image

I wonder if this is the description breaking?

addyosmani:
After re-reading the performance is about retaining users section, I felt that the bullet points here with stats could have been less verbose. For example:

> Pinterest rebuilt pages for performance, focusing on improving a custom metric called User Perceived Wait Time, which measured the time between the start of a navigation event and approximately when the DOM finished loading. A 40% reduction in this metric increased search engine traffic by 15%, as well as a 15% increase in signups.

Perhaps... "**Pinterest rebuilt their pages for performance realizing a 40% reduction in perceived wait times, increasing search engine traffic by 15% and a 15% increase in signups."**. As we're linking off to the original article, the details about their custom metric are available should someone wish to learn those details. 

In the screenshot for Lighthouse, I think we should probably link up to Lighthouse on /web.

I wondered if it made sense to generalize "Performance is about the money" to "Performance is about improving conversions". I don't have a strong opinion on this one but just leaving the thought.

> Image optimization is an incredibly broad and complex topic, but Addy Osmani's Essential Image Optimization makes sense of it all

I think (for consistency) we should reword this to "..but Automating Image Optimization makes sense of it all" as the name for "Essential Image Optimization" on /web is that. I had to rename as we already had some existing image compression docs on the site :)

addyosmani:
For the next round of commits, could we experiment with using this page as the overall overview with "Measuring performance with RAIL" shifting to a dedicated TOC item (the first one down the list).

src/content/en/fundamentals/performance/why-performance-matters/index.md
```diff
@@ -0,0 +1,390 @@
+project_path: /web/fundamentals/_project.yaml
+book_path: /web/fundamentals/_book.yaml
+description: Thanks to mobile device and network proliferation, more people are
```
petele: description should not break - it's only line that's allowed to exceed 80 chars.

malchata:
Got a new commit in:
- Addressed feedback in PR comments.
- Referred to RAIL (and PRPL) in the concluding paragraphs.
- Made this article the new overview page.
- Moved RAIL to a dedicated ToC entry.
- Changed the overview redirect to point to this article.

addyosmani:
Over to @petele for his sign-off.

malchata:
Before we land it, I'll want to do one more quick copy edit.

malchata:
Copy tweaks in!

robdodson:

PR #5859, opened 2018-03-04
Remove protocol from recommended code examples

rootulp:
- These links should be an example of protocol relative intrasite URLs so they should not have the preceding `http:`

**Fixes:** N/A

**Target Live Date:** ASAP

**CC:** @petele


petele:
Thanks for the PR, I've updated our content based on your feedback and replaced it with PR #5862

PR #5870, opened 2018-03-06
Link getImageData() correctly

dalmaer:
Vas is das markdown boog?

some extra `...` got in the mix to mess up the HTML goodness.

**CC:** @petele


petele:
Looks like this was fixed in #5867

PR #5435, opened 2017-12-05
[Translation zh-tw] Fundamentals > Events

kennith:
[Check out a review request at GitLocalize](https://gitlocalize.com/repo/107/zh-TW/review/433)

PR #5887, opened 2018-03-08
[Translation zh-tw] Fundamentals > Events 

kennith:
Recreate PR #5435 because the source repo was deleted. 

**CC:** @petele


petele:
Closing since the Events section is now gone.

PR #5895, opened 2018-03-09
Fix the link to the v3 migration guide

jeffposnick:
R: @petele


jeffposnick:
FYI—This change had previously been deployed, and this commit was just to get GitHub in-sync with what's live on the website.

PR #5935, opened 2018-03-17
Update security.md

LarryLACa:
Delete h3 'TL;DR' after first paragraph.
Already hidden from TOC.  No clear intention, reads fine without it.
Alternatively: change 'TL;DR' text to 'For example:'

**Fixes:** #issue - trivial typo, no issue created

- [x ] I previewed the change on my Win DevTools.
- [x ] I have added the appropriate `type-something` label - here and on patch

**CC:** @petele

### DIFF
--- a/src/content/en/tools/chrome-devtools/security.md
+++ b/src/content/en/tools/chrome-devtools/security.md
@@ -17,7 +17,6 @@ to debug security issues and ensure that you have properly implemented
 HTTPS on your websites.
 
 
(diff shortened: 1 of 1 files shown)


petele:
Congrats on your first commit, and thanks for your help. The `TL;DR` heading was something we previously used in our docs, but haven't kept up with. It's meant as a quick - what's important here. I'd prefer to leave it in until we can properly update this doc.

PR #5941, opened 2018-03-19
More Workbox docs fixes

jeffposnick:
This fixes a few typos/issues with the Workbox focs.

I also expanded a bit in the migration guide for the webpack plugin to address a common issue that seems to be cropping up, related to the use of `globPatterns` in v3.

**Fixes:**
https://github.com/GoogleChrome/workbox/issues/1378
https://github.com/GoogleChrome/workbox/issues/1379

**Target Live Date:** 2018-03-21

- [ ] This has been reviewed and approved by (@gauntface)

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/migrations/migrate-from-v2.md
+++ b/src/content/en/tools/workbox/guides/migrations/migrate-from-v2.md
@@ -447,7 +447,7 @@ between modes explicit, vs. the v2 behavior where behavior changed based on the 
 - By default, assets in the webpack compilation pipeline will be precached, and it is no longer
 necessary to configure `globPatterns`. The only reason to continue using `globPatterns` is if you
 need to precache assets that are
-[not included](https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#cache_additional_non-webpack_assets)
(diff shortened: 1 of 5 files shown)


src/content/en/tools/workbox/guides/migrations/migrate-from-v2.md
```diff
@@ -444,6 +444,13 @@ updated API surface.
 - The API now exposes two classes, `GenerateSW` and `InjectManifest`. This makes the toggling
 between modes explicit, vs. the v2 behavior where behavior changed based on the presence of `swSrc`.
 
+- By default, assets in the webpack compilation pipeline will be precached, and it is no longer
```
ZhangYiJiang: This note should probably also be added to the `glob*` option documentation on https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin - right now the `globPattern` docs seem to suggest only JS, CSS and HTML files are precached since it says that the default value is `['**/*.{js,css,html}']`

jeffposnick: Good point. Thanks.

Issue #5831, opened 2018-02-27
RSS feeds for each section, per year


ebidel:
We have feeds for the last 20 items for each main content section:

https://developers.google.com/web/fundamentals/rss.xml
https://developers.google.com/web/tools/rss.xml
https://developers.google.com/web/show/rss.xml

Let's create feeds per year with everything in it for each year...and remove `<content>` so these are not too big:

https://developers.google.com/web/fundamentals/2018/rss.xml
https://developers.google.com/web/fundamentals/2017/rss.xml
...

ebidel:
This is done.

PR #5974, opened 2018-03-25
Feeds for tools/showcase/fundamentals per year

ebidel:
Does not include per year feeds for shows.

**Fixes:** #5831

**CC:** @petele


ebidel:
How's this look?

petele:
Doing it this morning, want to do it as a separate CL though, so it had to wait to get the official one out.

PR #5977, opened 2018-03-26
Rename google.com/pay to Google Pay from Pay with Google

niallkennedy:
The https://google.com/pay payment method is known as Google Pay since Chrome M65. Change documentation referencing the payment method and its label to match current Chrome UI.

**CC:** @petele @agektmr


niallkennedy:
Screenshot needs updating. Text in background reads "Android Pay" (should be Google Pay) with old Chrome Payment Request dialog in foreground.
`/images/deep-dive/pr-demo-pwg-and-cards-short-blackout.png`

PR #5979, opened 2018-03-27
Workbox latest

gauntface:
Adds workbox 3.0.1 docs and switches from prerelease to latest for reference docs.


gauntface:
@jeffposnick this is an automated set of changes - there is nothing I can do to change what is currently here.

src/content/en/tools/workbox/reference-docs/v3.0.1/index-all.html
```diff
@@ -0,0 +1,131 @@
...
+      <div id="jsdoc-content-container">
+        <div id="jsdoc-banner" role="banner">
+        </div>
+        <div id="jsdoc-main" role="main">
+          <div class="symbol-detail-labels">
+            <span class="label label-kind"><small>Version</small></span> &nbsp;&nbsp;&nbsp;
+            <span class="label label-version"><small>v3.0.1</small></span>
+          </div>
+          <ul>
+            <li><a href="module-workbox-build.html#.ManifestEntry">module:workbox-build.ManifestEntry</a></li>
```
jeffposnick: Here's an example of a URL in the updated docset that refers to the now-non-existent `module-workbox-build.html` file.

This particular file is auto-generated, right? It's really weird that JS Doc would spit out HTML that included a URL referencing an HTML file, but not include the HTML file itself.

Hopefully https://github.com/GoogleChrome/workbox/pull/1405 will help?

Issue #5917, opened 2018-03-15
Wider audio element


jakearchibald:
https://developers.google.com/web/shows/http203/podcast/cooties-symbols-and-shaders

Could we make the audio element `width: 100%`? Some folks have mentioned it's difficult to skip around the podcast.

I know it's a trivial change, but I couldn't figure out the right place to set it.

petele:
Yah, adding `style="width: 100%"` to the audio element is likely the easiest, there are other ways, but this is only applicable on a few pages...

beaufortfrancois:
I've opened https://github.com/google/WebFundamentals/pull/6004 to fix this.

PR #6004, opened 2018-04-05
Wider audio podcast element

beaufortfrancois:
The audio podcast element width is now 100% as some folks have mentioned it was difficult to skip around the podcast.

**Fixes:** #5917

**CC:** @petele


jakearchibald:
LGTM! Thank you.

PR #6024, opened 2018-04-09
Small typo fix for the Workbox landing page

jeffposnick:
R: @petele @gauntface 

As reported at https://github.com/GoogleChrome/workbox/issues/1417


jpmedley:
FWIW, 'stowage' is actually an English word. Isn't it used in that new pirate API.

`pirate.plunder.then(stow(booty) => { ... })`

PR #6083, opened 2018-04-22
Highlight code

dereklimbus:
For unknown reasons, the four spaces doesn't work for the code under Feature Detection, so three backticks is applied instead.

- The code under Feature Detection was not highlighted.

**Fixes:** #issue

**CC:** @petele

### DIFF
--- a/src/content/en/fundamentals/push-notifications/subscribing-a-user.md
+++ b/src/content/en/fundamentals/push-notifications/subscribing-a-user.md
@@ -23,17 +23,17 @@ push is supported with two simple checks.
 
 1. Check for *serviceWorker* on *navigator*.
 1. Check for *PushManager* on *window*.
+```
+if (!('serviceWorker' in navigator)) {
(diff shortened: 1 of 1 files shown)


dereklimbus:
Thanks and sorry for the trouble. It was my first time and I did it online, but will definitely do that in the future.

In terms of why the four spaces didn't work, I did a little experiment just now. It is not because it needs one more line, because no matter how many lines I added before that, it still didn't work. I observed there was a list right above it, and when I removed the list, it worked fine. So I think there might be a conflict between list formatting and code block formatting. But I don't know how it works, so it's just my guess.

I also took a brief look at [Style Guide](https://petele-scratch.appspot.com/web/resources/style-guide#code_blocks), and it says:

> Warning: Wrapping code blocks in ``` is not supported on DevSite. DevSite will not automatically style these blocks and our integration tests will fail.

Not sure if this matters. At least it works fine on the website.

PR #6090, opened 2018-04-24
Add "Was this page helpful?" widget

kaycebasques:
- adds a "was this page helpful?" inline feedback widget
- adds the widget to a few random pages

The new Helpful widget is an abstraction on top of the previous inline feedback work I've done. The main benefit is that we can now drop it in with a single line:

    {% include "web/_shared/helpful.html" %}

When users click the "Yes" or "No" button, it gets logged as a Google Analytics event, as before. The category is now "Helpful", and the label is the path to the page (thanks to `window.parent.location.pathname`, which Dan or Dave pointed me to). This will enable us to compare the helpfulness of every page, side-by-side, in the Google Analytics dashboard. The standardized structure also lends itself well to pulling the data into other dashboards.

We'll be able to replicate this pattern for other interesting general feedback questions, such as "was the page too long?" 

Note that it's not rendering correctly on the local dev server, but I tried it out on staging, and it seems to work just fine!

**Fixes:** N/A

**Target Live Date:** 2018-04-27

- [ ] This has been reviewed and approved by @petele

**CC:** @Meggin

### DIFF
--- a/src/content/en/_shared/helpful.html
+++ b/src/content/en/_shared/helpful.html
@@ -1,0 +1,36 @@
+{% framebox width="auto" height="auto" enable_widgets="true" %}
+<script>
+var label = window.parent.location.pathname,
+  title = 'Feedback%20on%20' + label,
+  url = 'https://github.com/google/webfundamentals/issues/new?title=' + title,
+  link = '<a href="' + url + '">open an issue</a>',
+  response = 'Thanks for the feedback. Please ' + link + ' and tell us how we can improve.';
+var feedback = {
+  category: "Helpful",
+  question: "Was this page helpful?",
+  choices: [
+    {
+      button: {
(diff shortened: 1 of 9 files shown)


src/content/en/_shared/helpful.html
```diff
@@ -0,0 +1,36 @@
+{% framebox width="auto" height="auto" enable_widgets="true" %}
+<script>
+var label = window.parent.location.pathname,
```
petele: Per the Google JS Style guide, you should only have [one variable per declaration](https://google.github.io/styleguide/jsguide.html#features-one-variable-per-declaration) - will change and merge.

kaycebasques:
Just checked prod, not working there because `window.parent.location.pathname` throws a cross-origin error

PR #6098, opened 2018-05-01
New dvd podcast

mustafa-x:
Adding new podcast for the new series :)


petele:
Typos are fine, though according to our style guide, it should be sign-in. The error is being reported because of the missing CLA. @mustafa-x - are you signing in/running from a different account than normal?

PR #6115, opened 2018-05-05
Remove new keyword from workbox.strategies inside registerRoute method

Nooshu:
- When registering a route with a loading strategy in 3.2.0 using the `new` keyword as the documentation on this page mention causes the following error: `Uncaught TypeError: workbox.strategies.staleWhileRevalidate is not a constructor`. Removing the new keyword works as expected so I assume this is an API change and the document hasn't been updated?

**Fixes:** n/a

**Target Live Date:** n/a

**CC:** @petele


jeffposnick:
Thanks!

PR #6120, opened 2018-05-08
[Autoplay policy] Add MediaStream section

beaufortfrancois:
This PR adds a MediaStream section to the Autoplay Policy article to mention that autoplay policy does NOT apply to audio/video elements provided with a MediaStream for now.


jpmedley:
Sorry about that.

PR #6165, opened 2018-05-21
Add CSS Paint Polyfill

developit:
**CC:** @petele


petele:
@developit - please remember to update the `wf_updated_on` field in the future.

PR #6167, opened 2018-05-21
New dvd podcast

mustafa-x:
Ready to push, there is an issues with the Line must be at most 100 characters on line 10 but this can't be broken to a separate line because it breaks the markdown in there. 

Also there is a typo issue, its picking up the word "designing" and  thinking it should be "sign-in".


src/content/en/shows/designer-vs-developer/_toc.yaml
```diff
@@ -3,6 +3,8 @@ toc:
     section:
     - title: 'Podcast'
       section:
+      - title: 'Designing for standards and using Styleable components'
+        path: /web/shows/designer-vs-developer/podcast/DVDS3E02-designing-for-standards.md
```
petele: Don't include `.md` in file paths here

PR #6210, opened 2018-06-01
Fix minor typos

mathiasbynens:
**CC:** @petele


jpmedley:
Pro tip: Making a plural or possessive out of a code term is ugly because it requires mixing type faces. I try to avoid constructions that require that. So your sentence would be like:

And a `BigInt` makes dealing with big integers way easier. I won't hold up this PR for that. Just an FYI.

PR #6214, opened 2018-06-01
Fix Workbox docs typo: s/weary/wary

dtinth:
- Fixed a typo in the docs. “weary” means tired, but I think the docs meant to say “wary” which means cautious.

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/route-requests.md
+++ b/src/content/en/tools/workbox/guides/route-requests.md
@@ -38,7 +38,7 @@ workbox.routing.registerRoute(
 );
 ```
 
-The only thing to be weary of is that this would only match for requests
+The only thing to be wary of is that this would only match for requests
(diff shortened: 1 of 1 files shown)


src/content/en/tools/workbox/guides/route-requests.md
```diff
@@ -38,7 +38,7 @@ workbox.routing.registerRoute(
 );
 ```
 
-The only thing to be weary of is that this would only match for requests
+The only thing to be wary of is that this would only match for requests
 on your origin. If there was a separate site what had the URL
```
dtinth: Should “what” also be changed to “that” or “which”?

jpmedley: It should be 'that'.

dtinth: Thanks, done :D

PR #6225, opened 2018-06-05
Makes building RSS/ATOM feeds optional

petele:
Makes building of RSS/ATOM feeds optional. Takes general build time from about 10 seconds to about 3 seconds.

OK to review, but DO NOT MERGE until I've landed to corresponding CL

### DIFF
--- a/gulp-tasks/build.js
+++ b/gulp-tasks/build.js
@@ -11,6 +11,7 @@ const gulp = require('gulp');
 const path = require('path');
 const glob = require('globule');
 const jsYaml = require('js-yaml');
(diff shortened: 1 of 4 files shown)


gulp-tasks/build.js
```diff
@@ -169,6 +170,7 @@ gulp.task('build:showcase', function() {
   });
 
   // Generate the RSS & ATOM feeds
+  options.title = 'Show Cases';
```
ebidel: one word?

petele: All other instances in here are two words.

PR #6230, opened 2018-06-06
Update to the caching headers section of the SW lifecycle doc

jeffposnick:
This updates @jakearchibald's SW lifecycle docs to reflect the new default behavior.

**Target Live Date:** 2018-06-13

**CC:** @petele


jeffposnick:
The change looks like:

screen shot 2018-06-06 at 1 29 39 pm

jakearchibald:
I think I'd rather the bullet point was changed to reflect reality. As in, state that the fetch for the service worker script ignores `max-age`, but `importScript`s do not. Then use a note to state that this is a new behaviour introduced in 68, and there's an option to control it (linking to your article).

jeffposnick:
Thanks, @jakearchibald. PTAL.

PR #6239, opened 2018-06-08
Generic Sensors Fix

jpmedley:
Pete,  

See the email thread titled, "Generic Sensors Not Actually Out".

**CC:** @petele


jpmedley:
This turns out to be incorrect.

PR #6241, opened 2018-06-08
Generic sensors update

petele:
Whoops!

cc: @jpmedley


jpmedley:
I closed the other PR intentionally. There is apparently a way to shut off a flag before the plumbing is removed. The plumbing that I found is not active. Therefore, Generic Sensors are enabled.

PR #6260, opened 2018-06-14
New dvd podcast

mustafa-x:
Still picking up issues when I pull from master, even after reseting the branch.


petele:
Replaced with #6261 

I'll follow up with you on email about ways to avoid this in the future.

PR #6267, opened 2018-06-15
Add ARCore to requirements.

jpmedley:
This also changes a publish date so that the AR articles are in the intended order on /Updates.


jpmedley:
The line that's warning is one long link followed by a period. Don't really want the period on the next line.

Issue #6277, opened 2018-06-19
Invalid 'ignoreUrlParametersMatching' example in workbox-precaching.md


dejan9393:
**Page Affected:** https://developers.google.com/web/tools/workbox/modules/workbox-precaching#ignore_url_parameters

**What needs to be done?**

The 'ignoreUrlParametersMatching' parameter requires an array, the example on the above mentioned page shows a singular Regex pattern.

jeffposnick:
This was fixed by https://github.com/google/WebFundamentals/pull/6278

PR #6278, opened 2018-06-19
Update ignoreUrlParametersMatching example

dejan9393:
Update the ignoreUrlParametersMatching example on the workbox-precaching.md page to correctly identify the parameter as an array of Regex patterns

- Correctly depict ignoreUrlParametersMatching as an array of Regex patterns

**Fixes:** #6277 

**CC:** @petele


petele:
In future, please remember to update the `{# wf_updated_on: yyyy-mm-dd #}` tag at the top of the doc, it's pretty important and used for localization.

PR #6338, opened 2018-07-11
Add section + diagram to explain how queuing affects available input response time

developit:
- Adds a section with a diagram explaining how idle tasks cause queueing and how that reduces the available time for input response from 100ms to 50ms.

**CC:** @petele @tdresser


src/content/en/fundamentals/performance/rail.md
```diff
@@ -106,6 +106,28 @@ of their time waiting for sites to respond to their input, not waiting for the s
   careful not to block the user. If possible, do work in the background.
 * For actions that take longer than 50ms to complete, always provide feedback.
 
+**50ms or 100ms?:**
```
tdresser: One line 94, I think it might make sense to clarify what we mean.

" ## Response: respond in under 50ms {: #response }"

"Respond in under 50ms" makes it sound like the total time taken should be < 50ms, but it just needs to be under 100ms.

I think that section in general could be a bit clearer. Perhaps instead of "respond in under 50ms" it should be "process events in under 50ms"?

addyosmani: 👍 on Tim's suggestion for tightening up the wording in the respond in <50ms line. Perhaps if we switch to "process events", it's worth including a very brief example so it's clear to folks what we mean too?

developit: Changed the heading and also modified the first guideline:

> Process user input events within 50ms to ensure a visible response within 100ms, otherwise the connection between action and reaction is broken. This applies to most inputs, such as clicking buttons, toggling form controls, or starting animations. This does not apply to touch drags or scrolls.

src/content/en/fundamentals/performance/rail.md
```diff
@@ -91,21 +91,44 @@ fast Wi-Fi connection, so users have grown accustomed to a 1000ms loading experi
...
   input immediately. You can use this 100ms window to do other expensive work. But be
   careful not to block the user. If possible, do work in the background.
 * For actions that take longer than 50ms to complete, always provide feedback.
 
+**50ms or 100ms?:**
+
+The goal is respond to input in under 100ms, so why is our budget only 50ms?
+This is because there is generally other work being done in addition to input
+handling, and that work takes up part of the time available for acceptible input
+repsonse. If an application is performing work in the recommended 50ms chunks
```
tdresser: s/repsonse/response

addyosmani:
Feel free to add your self to the author/contributors list if you like.

Once the above typo is fixed this should be all good to land :)

developit:
Fixed!

PR #6352, opened 2018-07-12
workbox-navigation-preload module doc

jeffposnick:
Docs for the new `workbox-navigation-preload` module, which shipped in Workbox v3.4.1.

This PR depends on https://github.com/google/WebFundamentals/pull/6350 being merged first.

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/modules/_index.yaml
+++ b/src/content/en/tools/workbox/modules/_index.yaml
@@ -153,6 +153,22 @@ landing_page:
         - label: Demo
           path: https://workbox-demos.firebaseapp.com/demo/workbox-streams/
 
+    - items:
+      - heading: workbox.navigationPreload
+        description: >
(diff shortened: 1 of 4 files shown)


src/content/en/tools/workbox/modules/_index.yaml
```diff
@@ -153,6 +153,22 @@ landing_page:
...
           path: https://workbox-demos.firebaseapp.com/demo/workbox-streams/
 
+    - items:
+      - heading: workbox.navigationPreload
+        description: >
+          Enable navigation preload, to get a network response for navigation
+          requests faster.
+        buttons:
+        - label: Learn more
+          path: ./workbox-navigation-preload
```
petele: In future, please try to use full paths, that way if URLs change, it's easier to do find/replace

Issue #2354, opened 2015-12-13
[EN] Feedback for: /web/fundamentals/design-and-ui/input/forms/?hl=en


greenido:
It might be time to replace the link to rAq (which is deprecated) with the article I've wrote on Autofill.

Thoughts?

PaulKinlan:
Sure, assigned to you as you are looking after monetization etc.  Please ensure that these remain up to date.

PaulKinlan:
What is the status of this?

greenido:
@petele how to do wish to do it?
I suggest we just remove this article on rAq and put a link to: https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill
Thoughts?

petele:
Given the flow content from updates should migrate into fundamentals when ready. So, I'd like to see the content from the updates post properly migrated into the fundamentals content. 

Is it officially deprecated or just scheduled to be? If it's already deprecated, we should add a header to the rAc content now that says this is and recommend something else, we can then remove the content later and redirect.

PEte

greenido:
Closing this one as we have this: https://github.com/google/WebFundamentals/issues/2584

PR #6371, opened 2018-07-14
Update Lighthouse doc concerning "Optimize Images"

xiatianyu:
Updated to reflect the following changes made to the "Optimize Images" Lighthouse audit:
* This audit now tests both JPEG and WebP images.
* The highest permissible JPEG quality has been bumped from 85 to 92 since [#2354](/GoogleChrome/lighthouse/commit/e84530e7329f0f025670bcb82ca6e5022f0b8551#diff-fd190dc3127d17941b6683b1a8c4ece6).

**CC:** @petele @patrickhulce


xiatianyu:
I signed it!

patrickhulce:
Thanks for the patch @xiatianyu! I think there might be a bit of confusion between the gatherer and the audit in Lighthouse though :)

The optimized-images gatherer collects both JPEG and WebP data as you identify, but there are two different audits consuming its output. One that just looks at JPEG data (the one that the audit document is for), and one that just looks at WebP data.

petele:
Any updates on this?

patrickhulce:
I think we can close this. The current documentation is correct given how we structure the audits in Lighthouse and while it could use an update to shy away from specifying "compression level 85", moving it to 92 doesn't really help clarify what we're talking about.

kaycebasques:
Whoops, just saw Patrick's comment. Closing.

PR #6389, opened 2018-07-18
Fix the cacheName option in the Workbox guides.

jeffposnick:
This fixes an incorrect option included in some of the Workbox guides.

**Target Live Date:** 2018-07-25

**CC:** @petele


src/content/en/tools/workbox/guides/generate-service-worker/cli.md
```diff
@@ -74,16 +74,16 @@ module.exports = {
 
   // Define runtime caching rules.
```
petele: Looks like this might be shared across several files - if that's the case, you might want to use a shared component similar to https://github.com/google/WebFundamentals/blob/master/src/content/en/tools/workbox/_shared/workbox-sw-cdn-url.html

PR #6399, opened 2018-07-20
 Fixed some typos in PWACompat article

kyarik:

Besides simple changes, "non-standard tags" was substituted
with "standard and non-standard tags", since link rel="icon"
is standard.

**CC:** @petele


PaulKinlan:
Thanks for this. The intent wasn't to say icon was not standard but other tags that are will be generated, but this pr makes it a lot clearer.

PR #6400, opened 2018-07-20
 Fixed single quotes in App Shell article

kyarik:

Using `'` instead of `’`.

**CC:** @petele


PaulKinlan:
Good catch.

PR #6404, opened 2018-07-23
Add Page Lifecycle API article

philipwalton:
**Target Live Date:** 2018-07-24

- [ ] This has been reviewed and approved by (@jpmedley)
- [x] I have run `npm test` locally and all tests pass.
- [x] I have added the appropriate `section-updates` label.
- [x] I've staged the site and manually verified that my content displays correctly.

**CC:** @petele

PR #6410, opened 2018-07-23
New in Chrome 68

petele:


petele:
@philipwalton - your post (#6404) is linked in and ready to go.

src/content/en/updates/2018/07/nic68.md
```diff
@@ -0,0 +1,206 @@
...
+When a user has a large number of tabs running, critical resources such as
+memory, CPU, battery and the network can be oversubscribed, leading to a
+bad user experience.
+
+If your site is running in the background, the system may suspend the it to
+conserve resources. With the new Page Lifecycle API, you can now listen for,
+and respond to these events.
+
+For example, if a tab needs to be discarded to conserve memory, the browser
+will fire the `frozen` event, where you can store any necessary state. Then,
```
philipwalton: This should be the `freeze` event.

src/content/en/updates/2018/07/nic68.md
```diff
@@ -0,0 +1,206 @@
...
+when the user refocuses the tab, the `resume` event is fired, making it
```
philipwalton: The `resume` event is only fired if a page is frozen but *not* discarded.

Since this example kinda conflates the two concepts, I'd simplify the whole paragraph to this:

> For example, if a user's had a tab in the background for a while, the browser may choose to suspend script execution on that page to conserve resources. Before doing so, it will fire the `freeze` event, allowing you to close open IndexedDB or network connections or save any unsaved view state. Then, when the user refocuses the tab, the `resume` event is fired, where you can reinitialize anything that was torn down.

philipwalton:
The diagram linked here is probably fine, but it's a bit old (we should change it in the explainer).

My post links to a newer (albeit more complex) diagram. Do you think it makes sense to link to that instead? https://github.com/google/WebFundamentals/pull/6404/files#diff-db626bcd50800f959d5473bacbc818c8

PR #6429, opened 2018-07-26
Fix array mutation on RSS generation

petele:
Last 8 articles were being dropped from the RSS feed generation because the array of files was being mutated as it generated other files. This prevents the array from being mutated.

**CC:** @ebidel


petele:
@ebidel - I'm going to merge this to get it fixed ASAP, but a review is certainly welcome.

PR #6445, opened 2018-07-31
Fix of typo in code example

pryazhnikov:
"Social Discovery" page has a typo in code examples.
Prefix attribute value at root HTML tag has namespace "g" instead of "og"
which is used at the rest part of the code.

- "g:" in prefix is changed to "og:"

**Fixes:** #issue

**CC:** @petele


PaulKinlan:
thank you

PR #6485, opened 2018-08-10
Warnings about async Workbox importScripts()

jeffposnick:
R: @philipwalton 

This is primarily to add in some info about avoiding asynchronous calls to `importScripts()` when using Workbox, in advance of a [change in Chrome](https://www.chromestatus.com/feature/5748516353736704) to adhere to a restriction that other browsers already employ.

See https://github.com/GoogleChrome/workbox/issues/1504 for more context.

It also fixes a small typo pointed out in https://github.com/GoogleChrome/workbox/issues/1602 and provides some context as to why you might *not* want to call `workbox.skipWaiting()`.

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/get-started.md
+++ b/src/content/en/tools/workbox/guides/get-started.md
@@ -3,7 +3,7 @@ book_path: /web/tools/workbox/_book.yaml
 description:Get Started with Workbox.
 
 {# wf_blink_components: N/A #}
-{# wf_updated_on: 2018-03-13 #}
+{# wf_updated_on: 2018-08-10 #}
 {# wf_published_on: 2017-11-15 #}
 
 # Get Started {: .page-title }
@@ -71,6 +71,15 @@ officially loaded in our service worker.
 ![DevTools screenshot of Workbox loading in a service worker.](../images/guides/get-started/yay-loaded.png)
 
 Now we can start using Workbox.
+
+Warning: Importing `workbox-sw.js` will create a
+[`workbox` object](/web/tools/workbox/modules/workbox-sw) inside of your service worker, and that
+instance is responsible for importing other helper libraries, based on the features you use. Due to
+restrictions in the
+[service worker specification](https://www.chromestatus.com/feature/5748516353736704),
+these imports need to happen either inside of an `install` event handler, or synchronously in the
+top-level code for your service worker. More details, along with workarounds, can be found in the
+[`workbox-sw` documentation](/web/tools/workbox/modules/workbox-sw#avoid_async_imports).
 
 ## Using Workbox
 
@@ -155,7 +164,7 @@ workbox.routing.registerRoute(
 
 Routing and caching strategies are performed by the `routing` and
 `strategies` modules, but there are plenty of other modules, each offering
-specific behaviours that you can use in your service worker.
+specific behaviors that you can use in your service worker.
 
 You'll find a number of guides that cover other features of Workbox as well
 as more information on configuring Workbox. Find a full list on the left, but
--- a/src/content/en/tools/workbox/modules/workbox-strategies.md
+++ b/src/content/en/tools/workbox/modules/workbox-strategies.md
@@ -3,7 +3,7 @@ book_path: /web/tools/workbox/_book.yaml
 description: The module guide for workbox-routing.
 
 {# wf_blink_components: N/A #}
-{# wf_updated_on: 2018-04-03 #}
+{# wf_updated_on: 2018-08-10 #}
 {# wf_published_on: 2017-11-27 #}
 
 # Workbox Strategies {: .page-title }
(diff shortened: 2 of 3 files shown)


src/content/en/tools/workbox/guides/get-started.md
```diff
@@ -155,7 +164,7 @@ workbox.routing.registerRoute(
 
 Routing and caching strategies are performed by the `routing` and
 `strategies` modules, but there are plenty of other modules, each offering
-specific behaviours that you can use in your service worker.
+specific behaviors that you can use in your service worker.
```
philipwalton: lols

src/content/en/tools/workbox/modules/workbox-strategies.md
```diff
@@ -136,7 +136,7 @@ workbox.routing.registerRoute(
   new RegExp('/images/'),
   workbox.strategies.cacheFirst({
     cacheName: 'image-cache'
-  }
+  })
```
philipwalton: Can we put commas here (and below) while we're adding stuff :)

src/content/en/tools/workbox/modules/workbox-sw.md
```diff
@@ -66,7 +66,72 @@ workbox.setConfig({
...
-With this, you’ll use only the local workbox files.
+With this, you’ll use only the local Workbox files.
+
+## Avoid Async Imports
+
+Under the hood, loading new modules for the first time involves calling
+[`importScripts()`](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts)
+with the path to the corresponding JavaScript file (either hosted on the CDN, or via a local URL).
+In either case, an important restriction applies: the implicit calls to `importScripts()` can only
+happen inside of a service worker's `install` handler, *or* during the synchronous,
```
philipwalton: I don't think you need a comma after "handler".

src/content/en/tools/workbox/modules/workbox-sw.md
```diff
@@ -66,7 +66,72 @@ workbox.setConfig({
...
+If you need to write code that would otherwise run afoul of this restriction, you can explicitly
+trigger the `importScripts()` call outside of the event handler, using the
+[`workbox.loadModule()`](/web/tools/workbox/reference-docs/latest/workbox#.loadModule) method:
+
+```js
+importScripts('{% include "web/tools/workbox/_shared/workbox-sw-cdn-url.html" %}');
+
+// This will trigger the importScripts() needed for workbox.core and workbox.strategies
+workbox.loadModule('workbox-core');
+workbox.loadModule('workbox-strategies');
```
philipwalton: Hmmm, this doesn't seem like the best workaround to me.

The problem with the `loadModule()` method is it doesn't load dependencies, which means the developer has to know what they are, but I don't think we can/should expect them to know that.

I think a much better option is to recommend that they just reference them in the top-level scope, e.g.:

```js
importScripts('https://storage.googleapis.com/workbox-cdn/releases/3.4.1/workbox-sw.js');

const {strategies, routing, googleAnalytics} = workbox;
```

Alternatively we could make `loadModule()` dependency-aware, but I still think it's cleaner to reference the modules in the top-level scope.

jeffposnick: Here's an example of a SW that just does `workbox.loadModule('workbox-google-analytics');`, and that pulls in a bunch of the related other libraries. I don't think that's any different than what's loaded with your approach.

https://rawgit.com/jeffposnick/4c62a8815b679fd22fb24d57607b6f89/raw/f8c2bb303e50895c33bbd1fe1eb0fa1be9d086c2/index.html

jeffposnick: (I'm going to remove the call to `workbox.loadModule('workbox-core');` from the example because I realize that's not needed.)

src/content/en/tools/workbox/modules/workbox-strategies.md
jeffposnick: As it turns out, looking through this again made me spot a bunch of references to `workbox.routing()` which were not ever the correct syntax. I've updated those in the latest commit.

src/content/en/tools/workbox/modules/workbox-sw.md
philipwalton: Ahhh, yeah the reference to `workbox-core` is what made me think that, and then I looked at the implementation of `loadModule()` and didn't see any dependency references, so I jumped to that conclusion.

But I guess since the build version of `workbox-google-analytics` references `workbox.core`, `workbox.backgroundSync`, etc. it works.

philipwalton: What if we document both approaches? That might also help explain with examples exactly what the `workbox-sw` loader is doing.

jeffposnick: Okay, I put in another example showing what you describe.

PR #6487, opened 2018-08-13
Updated media intent handler image.

beaufortfrancois:
This PR removes clankium application icon from the media intent handler image.


Meggin:
I'm not able to checkout this branch for some reason. Can you provide the link to the file that references this image? That way I can make sure the link is OK.

PaulKinlan:
Clicked the link and pressed view image and could see the difference via history. LGTM.

PR #6523, opened 2018-08-20
Update Google Fonts caching advice

philipwalton:
Updated the Google Fonts caching recommendations.

**Target Live Date:** ASAP

- [ ] This has been reviewed and approved by (@jeffposnick)

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/common-recipes.md
+++ b/src/content/en/tools/workbox/guides/common-recipes.md
@@ -32,7 +32,7 @@ up too much storage on the user's device).
 workbox.routing.registerRoute(
   /^https:\/\/fonts\.googleapis\.com/,
   workbox.strategies.staleWhileRevalidate({
-    cacheName: 'google-fonts',
+    cacheName: 'google-fonts-stylesheets',
   }),
 );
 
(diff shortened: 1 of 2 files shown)


src/content/en/tools/workbox/guides/common-recipes.md
```diff
@@ -12,22 +12,43 @@ This page contains a set of example caching strategies you can use with Workbox.
...
+Here we've limited the age of the cached font files to one year (matching the
+HTTP `Cache-Control` header) and the max entries to 30 (to ensure we don't use
+up too much storage on the user's device).
 
 ```javascript
+// Cache the Google Fonts stylesheets with a stale while revalidate strategy.
 workbox.routing.registerRoute(
-  new RegExp('https://fonts.(?:googleapis|gstatic).com/(.*)'),
+  /^https:\/\/fonts\.googleapis\.com/,
+  workbox.strategies.staleWhileRevalidate({
```
jeffposnick: I'm not 100% sure what the behavior would be if you use the same `google-fonts` cache name for two different strategies, and one of those strategies uses cache expiration.

I *think* what would end up happening is that only the entries that were created by the strategy that uses the expiration plugin would end up being expired, but it would be good to either confirm that, or to use different cache names for the CSS and Fonts.

philipwalton: K, I can switch to use a different cache name for this example, but it'd probably be good to verify the behavior and, if it's as you fear, add a warning to the lib and/or the docs.

philipwalton: Updated.

PR #6510, opened 2018-08-17
PWA-ILT: 2018 content refresh updates

DavidScales:
This commit updates the content for the PWA ILT materials at https://developers.google.com/web/ilt/pwa/ with our "version 2" updates. 

- [x] This has been reviewed and approved by (David Scales / dscales)
- [x] I have run `npm test` locally and all tests pass.
- [x] I have added the appropriate `type-something` label. (I believe)
- [x] I've staged the site and manually verified that my content displays correctly.

**CC:** @petele

PR #6557, opened 2018-08-30
Backup  pwa ilt style corrections

DavidScales:
Style edits to PWA-ILT

* style folders, files, and url's as code instead of bold
* minor content fixes, mostly spacing/punctuation 

If this passes review, it is ready to merge (no breaking changes). 

- [x] This has been reviewed and approved by (dscales)

**CC:** @petele


DavidScales:
FYI @jpmedley many of the changes you requested from me in #6510 were merged in on #6547, but I've finished some of your other requests in this PR. 

Thanks for your review! 👍 

Some remaining notes on your feedback (cc: @petele too):

> Remove the extra space between 'some' and '[push'.

Looks like there are multiple instances where an extra space is inserted before a link and I can't see anything wrong with the source docs so I must suspect it could be an issue with the claat or cleanup? Doesn't appear to be a regression though.

> Why not rename the directory in the zip file?

My understanding is that GitHub names all zip downloads with the pattern `repo-branch.zip`. So I don't think I can change it myself. That said, I don't think I actually need the "Rename the folder to pwa-training-labs so that the name is consistent with the lab instructions." line here since this is the only doc that actually references the top-level directory. I've reworded this section, and it was merged in with #6547 but I can revise.

> This is a note not a key point. 

We had been using the [Codelab style](https://docs.google.com/document/d/1VjyoX4DqeTMB6gr63ss_hAByRpzELkXuIaIrioMisBo/edit#heading=h.33hkuldw88lv), where notes are inside a single cell table with a light 
green background. From experimentation it looks like the green background causes the "key point" conversion by claat, and that the single cell tables cause issues if there is code inside. We've adjusted the note formatting in the source docs and they were correct as of #6547.

Thanks again!

jpmedley:
Many of the changes I asked for, particularly the formatting ones, were a pretty big ask. Thank you for doing all that. I know how tedious it was.

I had issues with the way file names were used as headings. What happened with that?

DavidScales:
Heh, it was a bit of work but it looks much better and more consistent now. 

Yes! Thanks for catching that. I had actually forgotten. I've pushed another commit that removes them as requested.

PR #6573, opened 2018-09-01
[Translate ru] Accessibility ->index

DmitrySkripunov:
[Check out a review request at GitLocalize](https://gitlocalize.com/repo/107/ru/review/6826)


ilyaspiridonov:
@agektmr A few mistakes here and there but in general good to go.

PR #6657, opened 2018-09-21
Add note re: code caching

mathiasbynens:
Quick clarification on the state of code caching for modules in Chrome.

https://bugs.chromium.org/p/chromium/issues/detail?id=841466

**CC:** @petele


mathiasbynens:
Those are all false positives, @WebFundBot.

PR #6719, opened 2018-10-12
Fixes widget escaping

petele:
**CC:** @petele


kaycebasques:
Excellent. Thank you.

PR #6738, opened 2018-10-17
Small typo fix/rewording of a Workbox guide

jeffposnick:
R: @philipwalton @petele

Just resolves a few issues that someone on Twitter spotted.

### DIFF
--- a/src/content/en/tools/workbox/guides/handle-third-party-requests.md
+++ b/src/content/en/tools/workbox/guides/handle-third-party-requests.md
@@ -20,7 +20,7 @@ different and what you can do in Workbox to support these requests.
 ## Cross-Origin Requests and Opaque Responses
 
 One of the security mechanisms in browsers is that when a piece of JavaScript
-requests a URL on a different origin, it’s prevented from being able to
+requests a URL from a different origin, it’s prevented from being able to
 access the body and many other details of the response.
 
 When you get a response like this, it’s known as an "opaque response". Some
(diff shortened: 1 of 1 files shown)


src/content/en/tools/workbox/guides/handle-third-party-requests.md
```diff
@@ -129,6 +128,6 @@ workbox.routing.registerRoute(
 ```
 
 <aside markdown="1" class="warning">
-<strong>Warning:</strong> This will cache a response that could be an error and it will not
-get updated!
+<strong>Warning:</strong> This will cache a response that could be an error, would then never end
```
philipwalton: I think you mean "never end up" or "never get".

src/content/en/tools/workbox/guides/handle-third-party-requests.md
```diff
@@ -20,9 +20,8 @@ different and what you can do in Workbox to support these requests.
 ## Cross-Origin Requests and Opaque Responses
 
 One of the security mechanisms in browsers is that when a piece of JavaScript
-requests a URL on a different origin, it’s not prevented from being able to
-view the response. The reason for this is so websites can’t try and scan for
-URLs on a user’s network.
+requests a URL on a different origin, it’s prevented from being able to
```
philipwalton: Nit: should this be "from a different origin"?

jeffposnick:
Feedback addressed.

PR #6769, opened 2018-10-24
Update to Node.js v10

mathiasbynens:
Now that our internal infrastructure was updated accordingly, it's time to update this repository as well. This patch updates the Node.js version used for testing to v10+.

The patch also moves the Node.js version number to an `.nvmrc` file, so that people using `nvm` can just enter `nvm use` to install and switch to the expected Node.js version for this project. Travis CI uses the `.nvmrc` file as well, so this is the single source of truth.

Once this change is in, we can consider merging https://github.com/google/WebFundamentals/pull/6095.

**CC:** @petele


petele:
Can you also update package.json to require node 10

PR #6864, opened 2018-11-20
Switch localStorage to IDB in SW recommendations

jeffposnick:
R: @malchata @petele 

https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/client-hints/#client_hints_in_service_workers suggests using `localStorage` inside of a service worker, but it's actually not available there. `IndexedDB` should be used instead.


malchata:
Thanks for catching this, @jeffposnick, it was clear that I was mistaken.

Can you add yourself at the bottom in the credits in alphabetical order?

jeffposnick:
Sure thing!

PR #6865, opened 2018-11-20
Update anchor id for "quota usage"

jeffposnick:
R: @petele @kaycebasques 

@kaycebasques changed the `id` to use for the anchor in https://github.com/google/WebFundamentals/pull/6856, but unfortunately, we already have that https://developers.google.com/web/tools/chrome-devtools/progressive-web-apps#opaque-responses URL hardcoded into Chrome's DevTools interface.

I'm cool with the change to the section heading that Kayce made, but we need to use the specific `id` in this PR so that the original URL works.


kaycebasques:
Thanks for the fix. Sounds like I need to create some sort of test that throws an error if someone changes an anchor that DevTools depends on.

PR #6896, opened 2018-11-27
Rename “package name maps” to “import maps”

mathiasbynens:
Update phrasing per upstream name change.

**CC:** @petele


mathiasbynens:
Shut up @WebFundBot

PR #7015, opened 2018-12-20
feature-policy: update moved spec.

ebidel:
**CC:** @petele


dandv:
https://github.com/WICG/webappsec-feature-policy/issues/265

clelland:
Yeah, we're at an unfortunate midway point in the transition. The eventual destination should be w3c/webappsec-feature-policy, but we have to finish that transfer first.

PR #7099, opened 2019-01-11
User Activation v2 article.

jpmedley:


jpmedley:
@petele How do I find out what's blocking this? 'Unable to stage' doesn't provide any additional information.

beaufortfrancois:
Explanations are great! I feel like it lacks concrete examples about what we can do now.

petele:
I agree with @beaufortfrancois - but let's get this out now, and I've filed #7112

PR #7241, opened 2019-02-10
Add hot path with WebAssembly article

surma:
New article

**Target Live Date:** Wednesday, 2019-02-13

- [x] This has been reviewed and approved by (@jakearchibald @mathiasbynens @kripken @MaxGraey @ashleygwilliams et al)

**CC:** 

- @jpmedley: Can you please give this a review?
- @ashleygwilliams: If you have time to spare, can you please review this? I am talking about Rust, `wasm-pack` and Twiggy amongst other things :)
- @MaxGraey: Do you mind giving this a review? I am talking about AssemblyScript :)	

//edit: [Live preview has been generated](https://pr-7241-dot-web-central.appspot.com/web/updates/2019/02/hotpath-with-wasm).


MaxGraey:
@surma Thanks for sharing your experience with AssemblyScript! All perfectly! It might be a good addition:
```
The AssemblyScript team has been very responsive and is actively
working on improving their toolchain. We will keep an eye on
AssemblyScript in the future.
```
Or something similar to your discretion.

surma:
Thanks for reviewing, @MaxGraey! Regarding your note, that is [exactly how the article ends](https://github.com/google/WebFundamentals/pull/7241/files#diff-96406654e0197f030bce05f152e99edaR483), or am I misunderstanding you?

MaxGraey:
@surma ouch, my bad.  :joy:

src/content/en/updates/2019/02/hotpath-with-wasm.md
```diff
@@ -0,0 +1,492 @@
...
+Rust is a new, modern systems programming language. Amongst other things, it
+aims to be useful in the same places that C has been used traditionally, but
+brings modern language concepts and tooling. Rust also supports WebAssembly as a
+first-class citizen and the Rust team has contributed a lot of excellent tooling
+to the WebAssembly ecosystem.
+
+One of these tools is [`wasm-pack`](https://rustwasm.github.io/wasm-pack/),
+which is mostly written by [Ashley Williams](https://twitter.com/ag_dubs) and
+[Nick Fitzgerald](https://twitter.com/fitzgen?lang=en). `wasm-pack` takes your
+code and turns it into a web-friendly npm package that works out-of-the-box with
```
ashleygwilliams: you *can* make an npm package- but it's not required! something like a "web-friendly package (that you *could publish to npm*)" might work better

src/content/en/updates/2019/02/hotpath-with-wasm.md
```diff
@@ -0,0 +1,492 @@
...
+bundlers like webpack. `wasm-pack` is an extremely convenient experience, but
+currently only works for Rust. The plan is to add support for other
+WebAssembly-targeting languages in the future.
```
ashleygwilliams: This plan is... less than clear- the way this sentence states it is certainly ideal/ambitious. it's true but we may want to dial it back a bit, not sure what wording i'd pick tho

src/content/en/updates/2019/02/hotpath-with-wasm.md
```diff
@@ -0,0 +1,492 @@
...
+
+In Rust, slices are what arrays are in C. And just like in C, we need to create
+slices that use our start addresses. This goes against the memory safety model
+that Rust enforces, so to get our way we have to use the `unsafe` keyword,
+allowing us to write code that doesn't comply with that model.
+
+Note: This is not a best practice. In our experience it is usually worth it to
```
ashleygwilliams: is there a best practice example we could give? it's tough to only give one example *and* state that it's not best practice. usually people will ignore that not best practice bit...

src/content/en/updates/2019/02/hotpath-with-wasm.md
```diff
@@ -0,0 +1,492 @@
...
+#### Rust
+
+Rust is a new, modern systems programming language. Amongst other things, it
+aims to be useful in the same places that C has been used traditionally, but
+brings modern language concepts and tooling. Rust also supports WebAssembly as a
+first-class citizen and the Rust team has contributed a lot of excellent tooling
+to the WebAssembly ecosystem.
+
+One of these tools is [`wasm-pack`](https://rustwasm.github.io/wasm-pack/),
+which is mostly written by [Ashley Williams](https://twitter.com/ag_dubs) and
```
ashleygwilliams: as much as i love the shoutout it's probably best to link the rustwasm working group here instead of nick and i :)

src/content/en/updates/2019/02/hotpath-with-wasm.md
```diff
@@ -0,0 +1,492 @@
...
+the glue code and instantiate the WebAssembly modules with the vanilla APIs.
+This is often possible with Emscripten as long as you are not using anything
+from the C standard library.
+
+Note: We are working with the Emscripten team to make the glue code smaller or
+even non-existent at times.
+
+#### Rust
+
+Rust is a new, modern systems programming language. Amongst other things, it
```
ashleygwilliams: you may consider trying to use some of the marketing language that the new rust website uses- particularly the bit around empowering folks who may not have originally gravitated to or felt successful in other "systems" languages

src/content/en/updates/2019/02/hotpath-with-wasm.md
surma: Will dial it back :)

src/content/en/updates/2019/02/hotpath-with-wasm.md
surma: Fair point! I’ll add a best-practice example to the gist. Sound good?

src/content/en/updates/2019/02/hotpath-with-wasm.md
surma: I just looked at the contributor graph 😜 Will fix

src/content/en/updates/2019/02/hotpath-with-wasm.md
surma: understood.

MaxGraey:
@surma One correction.
Should be:
```ts
store<u32>(offset + i * 4, load<u32>(in_idx * 4));
```
instead:
```ts
store<u32>(offset + i * 4 + 4, load<u32>(in_idx * 4 + 4));
```

surma:
@MaxGraey Yes and no :D While you are technically correct, I have this note in the article: 
screen shot 2019-02-11 at 16 31 32

So instead of going back and worth between using and not using address 0, I just consistently avoid address 0 (even though AssemblyScript would do just fine).

MaxGraey:
Yes, and we also using `8` bytes for `HEAP_BASE` when using memory manager, but in case when we direct access to linear memory this could ignore. Also this require properly handle from host like
```ts
new Uint8ClampedArray(IMAGE_SIZE * IMAGE_SIZE * 4 + 4, 4)
```
instead
```ts
new Uint8ClampedArray(IMAGE_SIZE * IMAGE_SIZE * 4)
```
So if host setup with properly this doesn't matter, but if not we got `out of memory` exception)

dcodeIO:
There are two reasons for keeping the first 8 bytes of memory free. First is, as in other languages, that a pointer at address `0` is equivalent to `null`, so there shouldn't be actual data stored at that address to not confuse a check like `if (myObject == null) { /* considered not set, but would be actually set */ }`. The other is that Binaryen's asm.js output repurposes the first 8 bytes for type casting, i.e. converting i64<->f64, which might not be well documented. However, in any scenario that otherwise guarantes to not store a managed object at offset 0, which is the case here, the first reason doesn't apply and the second doesn't as well if no asm.js is generated. TLDR: Should be fine here. Hope this helps :)

src/content/en/updates/2019/02/hotpath-with-wasm.md
```diff
@@ -0,0 +1,490 @@
...
+no browser APIs involved that could affect performance, so this is just about
+raw execution speed. 8 seconds is much less acceptable. Enter WebAssembly.
+
+## WebAssembly for predictable performance
+
+In general, JavaScript and WebAssembly can achieve the same peak performance.
+However, in JavaScript it's often tricky to stay on the "fast path". One key
+benefit that WebAssembly offers is _predictable_ performance, even across
+browsers. The strict typing and low-level architecture allows for stronger
+assumptions and more optimizations during compilation. The function above is a
```
mathiasbynens: "more optimizations during compilation" seems to contradict what is said earlier about JS and wasm having the same peak performance. 

How about replacing this sentence:

> The strict typing and low-level architecture allows for stronger
> assumptions and more optimizations during compilation.

...with something like:

> JavaScript compilers generate high-performance machine code based on _speculative assumptions_, but have to deoptimize when those assumptions are violated (aka. "falling off the fast path"). In WebAssembly, there is no deoptimization step, thanks to the stronger guarantees offered by its strict typing and low-level architecture. WebAssembly can be optimized without speculating at all.

cc @gahaas

surma: I don’t think it contradicts the initial statement: AFAIU, difference is that wasm allows more optimizations _during compilation_, while JavaScript will  observed the code _running_ for a bit to make assumptions about typings and _then_ it will apply stronger optimizations.

I’d like to avoid going to deep into compiler theory here in the article, so I am trying to keep this section simple. WDYT?

gahaas: I agree with @mathiasbynens that it's easy to misunderstand the text as it is, but also with @surma  that it would be good to keep this section simple. I cannot find a good formulation yet though.

gahaas: > In general, JavaScript and WebAssembly can achieve the same peak performance. However, for JavaScript this performance can only be achieved on the "fast path", and it's often tricky to stay on that "fast path". One key benefit that WebAssembly offers is predictable performance, even across browsers. This means that in WebAssembly there only exists one "fast path", and it cannot be left. The strict typing and low-level architecture allow stronger assumption so that WebAssembly code only has to be optimized once.

Maybe like this?

surma: I like that approach. I’ll work it into the doc with some light rephrasing. Thank you very much

mathiasbynens: > I don’t think it contradicts the initial statement: AFAIU, difference is that wasm allows more optimizations _during compilation_, while JavaScript will observed the code _running_ for a bit to make assumptions about typings and _then_ it will apply stronger optimizations.

The "during compilation" is the confusing part, since the current text is conflating two different meanings of "compilation".

For JS, one meaning is "compiling bytecode" i.e. turning the AST into bytecode that can then start to run (in the interpreter). The second meaning is "compiling machine code" (i.e. what an optimizing compiler does for hot code).

For Wasm, you get a bytecode format from the start; the only meaning of "compiling" then is "compiling machine code" (i.e. the second meaning above).

It only makes sense to compare those two; anything else is apples vs. oranges. And there is not much to compare _during_ compilation, since JS and Wasm use the same optimizing compiler.

I agree we should avoid diving into compiler theory :) but I think we can clarify this section without digging too deep. I like @gahaas's phrasing (or something among those lines). 

Alternatively, perhaps some inspiration could be taken from [the Liftoff blog post](https://v8.dev/blog/liftoff#tiering-up-from-liftoff-to-turbofan), e.g.

> WebAssembly is statically typed, so the engine can generate optimized code right away.

mathiasbynens: Looks like this wasn't resolved yet. Also, the same statement appears in the intro of the article --- let's fix it there as well?

PR #7250, opened 2019-02-12
PiP shipped in Chrome OS 72

beaufortfrancois:
The API has shipped in Chrome 70 on Mac, Linux, and Windows and Chrome 72 on Chrome OS. This simply updates article that links to https://github.com/WICG/picture-in-picture/blob/master/implementation-status.md


beaufortfrancois:
You can find following line at the very end of the article `[implementation status page]: https://github.com/WICG/picture-in-picture/blob/master/implementation-status.md`so I'll squash and merge.

PR #7258, opened 2019-02-13
Use srcObject instead of createObjectURL for audio recording example

ElijahLynn:
Was going through https://developers.google.com/web/fundamentals/media/recording-audio/ and got error 

> (index):9 Uncaught (in promise) TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.

Used suggestion here and works nicely in Chrome stable 72. 

https://stackoverflow.com/a/53821674/292408
https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL

- item 1
- item 2

**Fixes:** #issue

**CC:** @petele

### DIFF
--- a/src/content/en/fundamentals/media/recording-audio/index.md
+++ b/src/content/en/fundamentals/media/recording-audio/index.md
@@ -44,7 +44,7 @@ the `files` property of the event object.
   recorder.addEventListener('change', function(e) {
     var file = e.target.files[0];
     // Do something with the audio file.
-    player.src =  URL.createObjectURL(file);
+    player.srcObject = file;
   });
 &lt;/script>
 </pre>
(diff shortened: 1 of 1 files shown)


petele:
@beaufortfrancois - can you take a look and approve if you're OK with the change?

beaufortfrancois:
Change is OK. Thank you @ElijahLynn for this. Do you mind updating all other occurrences of `createObjectURL` in the doc? If not, I'll do in a follow-up change.

ElijahLynn:
Sure, done. However, I discovered that due to an unescaped `<script> tag, the example of "Save the data from the microphone" was not rendering (nor below it) which also contained a createObjectURL(). Should I put that into this PR?

image

beaufortfrancois:
I'd merge this first and create another PR for the `script` tag issue.

ElijahLynn:
Thanks, will work on the other one soon, I couldn't get the example working on first try, will have to set aside more time for it.

PR #7274, opened 2019-02-15
[lighthouse] tap targets reference

kaycebasques:

- adds reference doc the the "tap targets" audit

**Fixes:** N/A

**Target Live Date:** 2019-02-22

- [ ] This has been reviewed and approved by @kdzwinel

**CC:** @petele


kaycebasques:
@kdzwinel does this reference doc for the tap targets audit look OK to you? Am I using "bounding rectangle" correctly in this context?

kaycebasques:
cc @rviscomi

src/content/en/tools/lighthouse/audits/tap-targets.md
```diff
@@ -0,0 +1,68 @@
...
+started boosting the ranking of mobile-friendly pages on mobile search results back in 2015.
+See [Rolling out the mobile-friendly update][webmasters].
+
+A tap target is inappropriately-sized when its bounding rectangle is less than 48 pixels. 48
+pixels is about the size of a human finger. Any target smaller than a finger requires a
+lot of precision to tap. The bounding rectangle is the target's `width` (or `height`) +
+`padding` + `border` + `margin`. Both of the icons in **Figure 2** are appropriately-sized.
+
+<figure>
+  <img src="/web/fundamentals/accessibility/imgs/touch-target.jpg"
```
rviscomi: Is this image in the PR?

src/content/en/tools/lighthouse/audits/tap-targets.md
```diff
@@ -0,0 +1,68 @@
...
+[webmasters]: https://webmasters.googleblog.com/2015/04/rolling-out-mobile-friendly-update.html
+[MDN]: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect#Value
+
+Tap targets are interactive elements, like buttons or links, that users frequently tap.
+Appropriately-sized tap targets make pages more mobile-friendly and accessible. Google Search
+started boosting the ranking of mobile-friendly pages on mobile search results back in 2015.
+See [Rolling out the mobile-friendly update][webmasters].
+
+A tap target is inappropriately-sized when its bounding rectangle is less than 48 pixels. 48
+pixels is about the size of a human finger. Any target smaller than a finger requires a
```
rviscomi: "a human finger" sounds like it's been cut off. Instead maybe "a person's finger"?

kaycebasques:
@mattzeunert does this doc look technically accurate to you? @rviscomi told me that you're working on SEO audits now. LMK if I'm mistaken and you're not the right person to review this.

src/content/en/tools/lighthouse/audits/tap-targets.md
```diff
@@ -0,0 +1,68 @@
...
+
+[webmasters]: https://webmasters.googleblog.com/2015/04/rolling-out-mobile-friendly-update.html
+[MDN]: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect#Value
+
+Tap targets are interactive elements, like buttons or links, that users frequently tap.
+Appropriately-sized tap targets make pages more mobile-friendly and accessible. Google Search
+started boosting the ranking of mobile-friendly pages on mobile search results back in 2015.
+See [Rolling out the mobile-friendly update][webmasters].
+
+A tap target is inappropriately-sized when its bounding rectangle is less than 48 pixels. 48
```
mattzeunert: I'm not sure how much of this info is worth putting into the docs, but fyi here's a more detailed description of when we fail a tap target:

- targets that are at least 48x48px never fail
- we have some tolerance as well, so in practice tap targets greater than maybe 40x40px never fail
- links in a block of text never fail
- if a tap target is small AND there's another tap target nearby, then we fail the tap target

We never directly look at margins, instead we simulate a 48x48px finger and see if it touches anything it shouldn't touch. (If there is a margin that makes it less likely that the finger touches something it shouldn't.)

mattzeunert: I think we should point to remedies like adding margin or making the tap target bigger, and outline the success conditions we already talk about elsewhere in the docs (e.g. 48x48px min size). Rather than trying to explain what causes an LH failure.

src/content/en/tools/lighthouse/audits/tap-targets.md
mattzeunert: Make sure the examples actually fail in LH, since sometimes we try to be lenient 😂

mattzeunert:
> does this reference doc for the tap targets audit look OK to you? Am I using "bounding rectangle" correctly in this context?

We don't really determine pass/fail based on a bounding rect, so I think talking about size and distance of tap targets is better.

src/content/en/tools/lighthouse/audits/tap-targets.md
kaycebasques: The image already exists in the repo

src/content/en/tools/lighthouse/audits/tap-targets.md
kaycebasques: Done

PR #7320, opened 2019-02-27
Auto-generated docs for WB v4

jeffposnick:
R: @philipwalton 

A couple of manual changes were required, but this is mostly the output of `gulp workbox`.

This also removes most of our old auto-generated docs for previous releases, which are not linked to directly.

[WF_IGNORE:TYPOS,HELPFUL_WIDGET,MAX_LEN]


jeffposnick:
(Checking on a couple of 404s pointed out in the Warnings above...)

PR #7336, opened 2019-03-01
Add notes on Safari quirks for hidden lifecycle transitions.

ericandrewlewis:
- Add notes on Safari quirks for hidden lifecycle transitions.

**Fixes:** I didn't open an issue but can?

**Target Live Date:** ?

- [x] This has been reviewed and approved by (@philipwalton )

**CC:** @petele / @philipwalton

### DIFF
--- a/src/content/en/updates/2018/07/page-lifecycle-api.md
+++ b/src/content/en/updates/2018/07/page-lifecycle-api.md
@@ -2,7 +2,7 @@ project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: The Page Lifecycle API brings app lifecycle features common on mobile operating systems to the web. Browsers are now able to safely freeze and discard background pages to conserve resources, and developers can safely handle these interventions without affecting the user experience.
 
-{# wf_updated_on: 2018-09-04 #}
+{# wf_updated_on: 2019-03-01 #}
 {# wf_published_on: 2018-07-24 #}
 {# wf_tags: performance #}
 {# wf_blink_components: Blink #}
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2018/07/page-lifecycle-api.md
```diff
@@ -447,12 +447,19 @@ that pertain to lifecycle and lists what states they may transition to and from.
       point.</p>
       <aside class="warning">
         <strong>Warning:</strong>
-        The <code>beforeunload</code> event should only be used to alert the
-        user of unsaved changes. Once those changes are saved, the event should
-        be removed. It should never be added unconditionally to the
-        page, as doing so can hurt performance in some cases. See the
-        <a href="#legacy-lifecycle-apis-to-avoid">legacy APIs section</a>
-        for details.
+        The <code>beforeunload</code> event should only be used in two cases:
```
philipwalton: I think I'd prefer to only mention this in the "Managing cross-browsers differences" section. There are lots of subtle compat issues with the advice here, and I think it would be distracting to list it all inline rather than consolidated in a single section.

src/content/en/updates/2018/07/page-lifecycle-api.md
```diff
@@ -603,6 +610,26 @@ window.addEventListener('pagehide', (event) => {
...
   }
 }, {capture: true});
+
+// Safari does not reliably fire the `pagehide` or `visibilitychange`
+// events when closing a tab, so we have to use `beforeunload` with a
+// timeout to check whether the default action was prevented.
+// - https://bugs.webkit.org/show_bug.cgi?id=151610
+// - https://bugs.webkit.org/show_bug.cgi?id=151234
+// NOTE: we only add this to Safari because adding it to Firefox would
+// prevent the page from being eligible for bfcache.
```
philipwalton: Again, I think it would be best to just list this in the "Managing cross-browsers differences" section. If I included all the compat fixes I had in this code sample, it would just end up being the entire contents of the [page-lifecycle](https://github.com/GoogleChromeLabs/page-lifecycle/blob/0.1.1/src/Lifecycle.mjs) library.

src/content/en/updates/2018/07/page-lifecycle-api.md
philipwalton: Maybe adapt the comment text and say something like:

"Safari does not reliably fire the `pagehide` or `visibilitychange` events when closing a tab (webkit bugs: [151610](https://bugs.webkit.org/show_bug.cgi?id=151610) and [151234](https://bugs.webkit.org/show_bug.cgi?id=151234)), so in Safari you may need to *also* listen to the `beforeunload` event in order to detect a change to the hidden state. But since the `beforeunload` event can be cancelled, you need [to wait until after the event has finished propagating](https://github.com/GoogleChromeLabs/page-lifecycle/blob/0.1.1/src/Lifecycle.mjs#L156-L172) to know if the state has changed to hidden. **Important:** using the `beforeunload` event this way should *only* be done in Safari, as using this event in other browsers can hurt performance. See the [legacy APIs section](#legacy-lifecycle-apis-to-avoid) for details."

ericandrewlewis: Done.

src/content/en/updates/2018/07/page-lifecycle-api.md
ericandrewlewis: Done.

src/content/en/updates/2018/07/page-lifecycle-api.md
```diff
@@ -616,7 +616,7 @@ The above code does three things:
 <aside class="warning">
   <strong>Warning!</strong>
   This code yields different results in different browsers, as the order
-  (and reliability) of events has not been consistently implemented. To learn
+  and reliability of events has not been consistently implemented. To learn
```
ericandrewlewis: This small change makes it a bit more clear that reliability of the events across browsers is problematic.

PR #7362, opened 2019-03-07
Add note for mistaken notice.

jpmedley:
@petele Kind of a rush on this. I just found because the feature owner didn't bother to update the ticket until this week.


jpmedley:
Thanks a bunch.

PR #7366, opened 2019-03-07
Updated to the skipWaiting docs and related recipe

jeffposnick:
R: @philipwalton 

This accounts for changes in Workbox 4.1.0.

[WF_IGNORE:TEMPLATE_TAGS]

### DIFF
--- a/src/content/en/tools/workbox/guides/advanced-recipes.md
+++ b/src/content/en/tools/workbox/guides/advanced-recipes.md
@@ -2,7 +2,7 @@ project_path: /web/tools/workbox/_project.yaml
 book_path: /web/tools/workbox/_book.yaml
 description: Advanced recipes to use with Workbox.
 
-{# wf_updated_on: 2019-02-27 #}
+{# wf_updated_on: 2019-03-07 #}
 {# wf_published_on: 2017-12-17 #}
 {# wf_blink_components: N/A #}
 
@@ -15,32 +15,42 @@ worker has updated and waiting to install.
 
 To do this you'll need to add some code to your page and to your service worker.
 
-**Add to your page**
+**Code in your page**
 
 ```html
 <script type="module">
-import {Workbox} from 'https://storage.googleapis.com/workbox-cdn/releases/4.0.0/workbox-window.prod.mjs';
+import {Workbox} from 'https://storage.googleapis.com/workbox-cdn/releases/4.1.0/workbox-window.prod.mjs';
 
 if ('serviceWorker' in navigator) {
   const wb = new Workbox('/sw.js');
 
   // Add an event listener to detect when the registered
   // service worker has installed but is waiting to activate.
-  wb.addEventListener('waiting', () => {
+  wb.addEventListener('waiting', (event) => {
+    // `event.wasWaitingBeforeRegister` will be true if this is
+    // the first time the updated service worker is waiting.
+    // When `event.wasWaitingBeforeRegister` is false, a previously
+    // updated same service worker is still waiting.
+    // You may want to customize the UI prompt accordingly.
+
     // Assumes your app has some sort of prompt UI element
     // that a user can either accept or reject.
     const prompt = createUIPrompt({
       onAccept: async () => {
-        // Send a message to the service worker telling it to
-        // skip waiting. If/when the service worker replies,
-        // you know that `skipWaiting()` was called.
+        // Assuming the user accepted the update, set up a listener
+        // that will reload the page as soon as the previously waiting
+        // service worker has taken control.
+        wb.addEventListener('controlling', (event) => {
+          window.location.reload();
+        });
+
+        // Send a message telling the service worker to skip waiting.
+        // This will trigger the `controlling` event handler above.
         // Note: for this to work, you have to add a message
         // listener in your service worker. See below.
-        await wb.messageSW({type: 'SKIP_WAITING'});
-
-        // Reload to be controlled by the new service worker.
-        window.location.reload();
+        wb.messageSW({type: 'SKIP_WAITING'});
       },
+
       onReject: () => {
         prompt.dismiss();
       }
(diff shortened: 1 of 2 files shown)


jeffposnick:
Here's what the updated recipe looks like, since it's a little hard to make things out in the diff:

```
<script type="module">
import {Workbox} from 'https://storage.googleapis.com/workbox-cdn/releases/4.1.0/workbox-window.prod.mjs';

if ('serviceWorker' in navigator) {
  const wb = new Workbox('/sw.js');

  // Add an event listener to detect when the registered
  // service worker has installed but is waiting to activate.
  wb.addEventListener('waiting', (event) => {
    // `event.wasWaitingBeforeRegister` will be true if this is
    // the first time the updated service worker is waiting.
    // When `event.wasWaitingBeforeRegister` is false, a previously
    // updated same service worker is still waiting.
    // You may want to customize the UI prompt accordingly.

    // Assumes your app has some sort of prompt UI element
    // that a user can either accept or reject.
    const prompt = createUIPrompt({
      onAccept: async () => {
        // Assuming the user accepted the update, set up a listener
        // that will reload the page as soon as the previously waiting
        // service worker has taken control.
        wb.addEventListener('controlling', (event) => {
          window.location.reload();
        });

        // Send a message telling the service worker to skip waiting.
        // This will trigger the `controlling` event handler above.
        // Note: for this to work, you have to add a message
        // listener in your service worker. See below.
        wb.messageSW({type: 'SKIP_WAITING'});
      },

      onReject: () => {
        prompt.dismiss();
      }
    })
  });

  wb.register();
}
</script>
```

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -15,32 +15,42 @@ worker has updated and waiting to install.
...
 
 if ('serviceWorker' in navigator) {
   const wb = new Workbox('/sw.js');
 
   // Add an event listener to detect when the registered
   // service worker has installed but is waiting to activate.
-  wb.addEventListener('waiting', () => {
+  wb.addEventListener('waiting', (event) => {
+    // `event.wasWaitingBeforeRegister` will be true if this is
+    // the first time the updated service worker is waiting.
```
philipwalton: Actually the reverse is true. `event.wasWaitingBeforeRegister` is true after a reload and it's *still* waiting.

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -15,32 +15,42 @@ worker has updated and waiting to install.
 
 To do this you'll need to add some code to your page and to your service worker.
 
-**Add to your page**
+**Code in your page**
 
 ```html
 <script type="module">
-import {Workbox} from 'https://storage.googleapis.com/workbox-cdn/releases/4.0.0/workbox-window.prod.mjs';
+import {Workbox} from 'https://storage.googleapis.com/workbox-cdn/releases/4.1.0/workbox-window.prod.mjs';
```
philipwalton: Can we use the import that automatically adds the latest version?

BTW, I think in order to use it you have to use `<pre class="prettyprint">`...

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -15,32 +15,42 @@ worker has updated and waiting to install.
...
     const prompt = createUIPrompt({
       onAccept: async () => {
-        // Send a message to the service worker telling it to
-        // skip waiting. If/when the service worker replies,
-        // you know that `skipWaiting()` was called.
+        // Assuming the user accepted the update, set up a listener
+        // that will reload the page as soon as the previously waiting
+        // service worker has taken control.
+        wb.addEventListener('controlling', (event) => {
+          window.location.reload();
```
philipwalton: I don't think we need to wait for controlling here, do we?

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -53,32 +63,40 @@ if ('serviceWorker' in navigator) {
...
-package to register a service worker and respond if it gets stuck in the waiting
+package to register a service worker and react if it gets stuck in the waiting
 phase.
 
 When a waiting service worker is found we inform the user that an updated
 version of the site is available and prompt them to reload. If they accept the
 prompt, we `postMessage()` the new service worker telling it to run
-`skipWaiting()` meaning it'll start to activate.
+`skipWaiting()`, meaning it'll start to activate. Once the new service worker
+has activated and taken control, we reload the current page, causing the latest
```
philipwalton: If activation is blocked for some reason, the new SW will not take control. IIRC that's why I didn't wait for controlling in the previous version fo the recipe.

src/content/en/tools/workbox/guides/advanced-recipes.md
```diff
@@ -53,32 +63,40 @@ if ('serviceWorker' in navigator) {
...
+Workbox build tool in `InjectManifest` mode, then you need to add this to your
+service worker file yourself:
 
 ```javascript
 addEventListener('message', (event) => {
   if (event.data && event.data.type === 'SKIP_WAITING') {
     skipWaiting();
-
-    // Reply to the window so it knows `skipWaiting()` was called.
-    event.ports[0].postMessage(true);
```
philipwalton: Why did you remove this?

src/content/en/tools/workbox/guides/advanced-recipes.md
jeffposnick: I think it's safest to do that, since it's the only way to be sure that the new SW will control the page after the reload is if the controller has actually changed.

src/content/en/tools/workbox/guides/advanced-recipes.md
jeffposnick: I'll check that out.

src/content/en/tools/workbox/guides/advanced-recipes.md
jeffposnick: Gah. I knew that.

src/content/en/tools/workbox/guides/advanced-recipes.md
jeffposnick: Then what's the point of reloading if activation is blocked?

Are there scenarios in which reloading will clear up whatever is blocking activation?

src/content/en/tools/workbox/guides/advanced-recipes.md
jeffposnick: The way the code is written now, there's nothing that executes after the call to `wb.messageSW({type: 'SKIP_WAITING'})`, so there's no reason to wait for a response.

philipwalton: Previously I was using `await wb.messageSW()`, which would await a reply and ensure skipWaiting had been called before executing the next line of code.

src/content/en/tools/workbox/guides/advanced-recipes.md
jeffposnick: So https://github.com/google/WebFundamentals/blob/master/src/content/en/tools/workbox/_shared/workbox-sw-cdn-url.html explicitly includes the full URL to `workbox-sw.js`

We can update that file to just have the URL prefix, and then update the other docs that use the include file to combine the prefix + `workbox-sw.js`, and then update this usage to be the prefix + `workbox-window.mjs`.

How about doing that in another PR, though?

src/content/en/tools/workbox/guides/advanced-recipes.md
jeffposnick: The other issue is that the `message` handler in the service worker you get when using `GenerateSW` with `skipWaiting: false` (courtesy of https://github.com/GoogleChrome/workbox/pull/1929) doesn't call `event.ports[0].postMessage(true)`.

philipwalton: Hmmm, do you think we should change that. We could maybe even do a 4.1.1 release?

I don't mind code that waits for controlling, but my preference would be to always reply to the message event if we can. Reason being that in the future we might have to implement timeouts for messages and I'd rather not have to wait for a timeout in any cases other than version mis-matches.

jeffposnick: "reply to the message event **if we can**" is relevant here, since our `message` event listeners can't assume that they'll always be passed a `MessageChannel`. I know that `workbox-window` will always pass in one, but there is nothing stopping folks from calling `postMessage()` without using `workbox-window`.

So we'd need to conditionally check to see if `event.ports[0]` existed.

I am cool with that, but rather than rushing this out in a v4.1.1, why not implement that sort of standardization in `workbox-messages` and then we can build on top of that in a future release?

philipwalton: We can do that, however, if we're not `await`-ing the SW response, then we probably do need to wait until the controller change happens.

jeffposnick: Yup.

In other words, we can keep that `'controlling'` event listener logic that's already in this PR?

philipwalton: SGTM

src/content/en/tools/workbox/guides/advanced-recipes.md
philipwalton: Oh, right, this is a reference to `workbox-window` ... yeah, let's do that in another PR.

PR #7368, opened 2019-03-08
workbox: use correct filename in register-sw snippet

paulirish:
- @marcushellberg reported this: https://twitter.com/marcushellberg/status/1103827852552736768
- i changed the filename (to `service-worker.js`) in the register-sw shared snippet

@philipwalton i have no idea if this is correct so please tell me if this is incorrect.

**Target Live Date:** whenever

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/_shared/register-sw.html
+++ b/src/content/en/tools/workbox/guides/_shared/register-sw.html
@@ -6,7 +6,7 @@
 if ('serviceWorker' in navigator) {
   // Use the window load event to keep the page load performant
   window.addEventListener('load', () => {
-    navigator.serviceWorker.register('/sw.js');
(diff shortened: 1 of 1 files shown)


philipwalton:
Thanks, Paul. I saw that and filed https://github.com/google/WebFundamentals/issues/7369 because I wanted to confirm with @jeffposnick first that making this change wouldn't then be inconsistent with another default elsewhere.

jeffposnick:
Seems fine to me; that is the current default: https://github.com/GoogleChrome/workbox/blob/e15877f21f0290ad90ccb482baae20b606f9d3b6/packages/workbox-webpack-plugin/src/generate-sw.js#L46

petele:
Spoke with @jeffy - all good.

PR #7387, opened 2019-03-13
Clarify how built-in modules are standardized

philipwalton:
Also, emphasize that they're currently in an experimental phase, and
soften the language around deploying them today.

**Target Live Date:** ASAP

- [x] This has been reviewed and approved by (@domenic, @yoavweiss PTAL)

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2019/03/kv-storage.md
+++ b/src/content/en/updates/2019/03/kv-storage.md
@@ -9,7 +9,7 @@ description: An introduction to the new KV Storage API, built-in modules, and im
 {# wf_featured_snippet: An introduction to the new KV Storage API, built-in modules, and import maps. #}
 {# wf_blink_components: Blink>Storage #}
 
-# KV Storage, the Web's First Built-in Module {: .page-title }
+# KV Storage: the Web's First Built-in Module {: .page-title }
 
 {% include "web/_shared/contributors/philipwalton.html" %}
 
@@ -38,10 +38,12 @@ But what if it were possible to get the performance of an asynchronous storage
 API with the simplicity of the `localStorage` API, without having to pay the
 file size cost?
 
-Well, now there is. Chrome is experimenting with a new feature called [built-in
+Well, soon there may be. Chrome is
+[experimenting](https://groups.google.com/a/chromium.org/d/msg/blink-dev/sEwWEF80T4s/Nss9VxM3BAAJ)
+with a new feature known as [built-in
 modules](https://github.com/tc39/proposal-javascript-standard-library), and the
 first one we're planning to ship is an asynchronous key/value storage module
-called [KV Storage](https://github.com/WICG/kv-storage)
+called [KV Storage](https://github.com/WICG/kv-storage).
 
 But before I get into the details of the KV Storage module, let me explain
 what I mean by _built-in modules_.
@@ -54,8 +56,15 @@ are just like regular JavaScript
 except that they don't have to be downloaded because they ship with the browser.
 
 Like traditional web APIs, built-in modules must go through a standardization
-process and have well-defined specifications, but unlike traditional web APIs,
-they're not exposed on the global scope&mdash;they're only available via
+process &mdash; each will have its own specification that requires a [design
+review](https://github.com/w3ctag/design-reviews) and positive signs of
+support from both web developers and other browser vendors before it can ship.
+(In Chrome, built-in modules will follow the same [launch
+process](https://www.chromium.org/blink/launching-features) we use to implement
+and ship all new APIs.)
+
+Unlike traditional web APIs, built-in modules are not exposed on the global
+scope &mdash; they're only available via
 [imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import).
 
 Not exposing built-in modules globally has a lot of advantages: they won't add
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2019/03/kv-storage.md
```diff
@@ -144,12 +153,13 @@ If you're familiar with using native JavaScript modules in browsers, you
 probably know that (at least up until now) importing anything other than a URL
 will generate an error. And `std:kv-storage` is not a valid URL.
 
-So that raises the question: _do we have to wait until all browsers support
-built-in module before we can use it in our code?_
+So that raises the question: _do we have to wait until all browsers support a
+built-in module before we can use it in our code?_ Thankfully, the answer is no!
```
yoavweiss: So, that's already kinda true today for other features, with the help of polyfills.
But the great thing about built-ins is that they are:
a) polyfillable by definition
b) will only load the polyfill when needed (in browsers that support built-ins/import-maps), without requiring any complex server side shenanigans.

You may want to emphasize those points here.

philipwalton: The difference here though is that in the example shown, built-in modules are imported using syntax that will break in most browsers. And normally there's no way to polyfill unsupported syntax, so I do think it's different in this case.

As for the other points, I do I cover them in the section that follows, but I'd be happy to expand on them in a separate PR if you feel it's necessary (I want to get these changes in soon if possible).

src/content/en/updates/2019/03/kv-storage.md
```diff
@@ -54,8 +56,15 @@ are just like regular JavaScript
...
 
 Like traditional web APIs, built-in modules must go through a standardization
-process and have well-defined specifications, but unlike traditional web APIs,
-they're not exposed on the global scope&mdash;they're only available via
+process &mdash; each will have its own specification that requires a [design
+review](https://github.com/w3ctag/design-reviews) and positive signs of
+support from both web developers and other browser vendors before it can ship.
+(In Chrome, built-in modules will follow the same [launch
+process](https://www.chromium.org/blink/launching-features) we use to implement
+and ship all new APIs.)
```
tomayac: Suggestion: add a quick explanation about where people can see how we track "positive signs". Currently, for example on the [KV Storage ChromeStatus page](https://www.chromestatus.com/feature/6428344899862528), we have this: 

| Browser | Support |
| :---| :--- |
| Firefox: | No public signals |
| Edge: | [Public support](https://www.w3.org/2018/10/22-WebPlat-minutes.html#item05) |
| Safari: | No public signals |
| Web Developers: | Strongly positive `// 🤔 Link?` |

tomayac: @philipwalton FYI

src/content/en/updates/2019/03/kv-storage.md
```diff
@@ -144,12 +153,13 @@ If you're familiar with using native JavaScript modules in browsers, you
...
-So that raises the question: _do we have to wait until all browsers support
-built-in module before we can use it in our code?_
+So that raises the question: _do we have to wait until all browsers support a
+built-in module before we can use it in our code?_ Thankfully, the answer is no!
 
-Thankfully, the answer is **no!** You can actually use built-in modules in your
-code today, with the help of another new feature called
-[import maps](https://github.com/WICG/import-maps).
+You can actually use built-in modules as soon as even one browser supports them
+thanks to the help of another feature we're
```
domenic: I find the older wording more accurate; you don't need to wait for one browser.

src/content/en/updates/2019/03/kv-storage.md
```diff
@@ -275,10 +285,10 @@ code or even put breakpoints in it!):
 
 ## Please give us feedback
 
-This introduction should have given you a taste of what's possible with built-in
-modules. And hopefully you're excited! We'd really love for developers to try
-out the KV Storage module (as well as all the new features discussed here) and
-give us feedback.
+This introduction should have given you a taste of what may be possible with
```
domenic: This change seems unnecessarily vague.

src/content/en/updates/2019/03/kv-storage.md
```diff
@@ -289,9 +299,11 @@ features mentioned in this article:
 - [Import Maps](https://github.com/WICG/import-maps)
 
 If your site currently uses `localStorage`, you should try switching to the KV
-Storage API, and if you sign up for the [KV Storage origin
+Storage API to see if it meets all your needs. And if you sign up for the [KV
+Storage origin
 trial](https://developers.chrome.com/origintrials/#/trials/active), you can
-actually deploy your changes today! All your users should benefit from better
-performance, and Chrome 74+ users won't have to pay any extra download cost.
+actually deploy these features today. All your users should benefit from better
```
domenic: Not sure I understand this change; just reduces the enthusiasm.

src/content/en/updates/2019/03/kv-storage.md
domenic: I don't think every new feature's article should need to explain the Blink launch process and ChromeStatus to developers. Linking, as is done here, makes sense to me as an intermediate ground.

tomayac: People were specifically critical about this piece (and may read this popular post without much context), so IMHO here it makes sense to quickly summarize the process.

src/content/en/updates/2019/03/kv-storage.md
philipwalton: That's true. I can revert it.

philipwalton: Originally I was thinking you'd never want to implement this before a single browser supports it. But that may not actually be true in the case of client work where you deliver a site that don't expect to change in the future. Using import maps preemptively may be a way to get future perf gains without having to update the site. (Assuming the name of the module doesn't change).

src/content/en/updates/2019/03/kv-storage.md
philipwalton: The main change is the addition of "to see if it meets all your needs". IIRC I was thinking we should add that because we know there may be use cases the KV Storage API can't handle (e.g. when the page is unloading).

Is there another wording you'd prefer?

domenic: The chunk above makes sense, it's just the "your changes" -> "these features" and period -> exclamation point that I was confused about. But, this is minor; not worth doing a revet commit.

PR #7433, opened 2019-03-19
[Workbox] New third-party plugin section

jeffposnick:
R: @philipwalton
CC: @tpiros from Cloudinary as an FYI.

This adds a new "Third-party Plugins" section to the Workbox docs, with some general info about publicizing your plugins, and specific info about the plugin that Cloudinary created.


tpiros:
This is awesome @jeffposnick . I will also investigate the idea on writing an article about the experience of creating the plugin.

PR #7478, opened 2019-04-05
Add audio playlist sample

beaufortfrancois:
This is only about adding the audio playlist sample in the Picture-in-Picture article.


beaufortfrancois:
FYI @mounirlamouri

beaufortfrancois:
@PaulKinlan @petele Can you review/merge?

ahtee:
For picture in picture, it seems like the player should be opened to a relative position of the `window`, currently it has opened in different parts of the `screen` and the browser is not full-width of the page. However the functionality of the player appearing on top is consistent and shows above all other windows i have open. 

Would this be a consideration to add? Make Picture-in-picture be relative to the `window` object that calls PiP? Would the best position then be bottom-right?

beaufortfrancois:
@ahtee For security concerns, Chrome has chosen to always show the Picture-in-Picture window in the bottom right corner so that users know what to expect.

Note that there is feature request currently to persist window size/position across sessions: https://bugs.chromium.org/p/chromium/issues/detail?id=949570#

ahtee:
@beaufortfrancois thanks for that context. I have no additional feedback on my end other than adding the widget for the page that the PR has a warning on listed above..

> WARNINGS
> src/content/en/updates/2018/10/watch-video-using-picture-in-picture.md#L340 - Consider adding > a "was this page helpful?" widget to your page: 
> https://developers.google.com/web/resources/widgets#helpful

PR #7493, opened 2019-04-10
Adding `webp` to the image file extensions

nhoizey:
I've seen developers copy/paste this recipe and wonder why WebP image were not cached. 😅

I think using `destination===image` would be better, but I guess the purpose of this recipe is to show a regex.


PaulKinlan:
Good catch.

nhoizey:
👍

PR #7498, opened 2019-04-11
Web share target updates for new manifest values

jeffposnick:
R: @petele @PaulKinlan 

While testing out web share target, I noticed that DevTools will warn you about missing manifest attribute values:

Screen Shot 2019-04-11 at 10 35 40 AM

Those are, I guess, recently added to the spec, but they're not mentioned at all in https://developers.google.com/web/updates/2018/12/web-share-target

This PR adds a description of the `method` and `enctype` manifest attributes, along with some brief description of their default values, and why you might need to customize them.


PaulKinlan:
Are you testing on stable? iirc, this hasn't landed there yet.

jeffposnick:
That screenshot of DevTools telling me to set `method` and `enctype` in the manifest is from OS X 73.0.3683.103, the current stable release.

PR #7514, opened 2019-04-17
Introduce Site Kit section

felixarntz:
This pull-request adds a new `/web/site-kit/` section, which for now only holds a single page to aid users of the Site Kit WordPress plugin to set up an app to use for their plugin. Due to it being a single page, there is no `_book.yaml` necessary.

There are currently two test failures:
1. `ERROR: src/content/en/site-kit/_index.yaml#17 can not read a block mapping entry; a multiline key may not be an implicit key` (not sure what's wrong there with the YAML multiline string, it compiles on devsite)
2. `ERROR: src/content/en/site-kit/_project.yaml instance.google_analytics_ids[1] does not conform to the "wfUAString" format` (if it's really not possible to have another GA tracking ID, then I'll remove it of course)

Since this is a new section, maybe we should add a label `section-sitekit`, to account for it in the future.

**Target Live Date:** 2019-04-22 (if possible)

**CC:** @petele


felixarntz:
This is good to go from my end. Let me know if any further changes are required.

petele:
OK - when do you want this to go live?

felixarntz:
It can go live as soon as possible.

PR #7526, opened 2019-04-20
Workbox doc build changes

philipwalton:
@jeffposnick, this PR implements the changes we talked about in our last Workbox weekly meeting. It updates the `gulp workbox` task so it now does the following:

- Checks for the latest tag in each major version (ignoring v1) and build the docs for that tag in a directory corresponding to the major version (e.g. `v2`, `v3`, and `v4` and the moment).
- Checks to see if the latest prerelease version is greater than the latest stable version, if it is then it builds docs for that tag in the `prerelease` directory
- Builds the docs for the latest stable version in the `latest` directory.

Note, for the first and second bullets in the list above, for any tag lower than the latest stable version, if a directory already exists it is assumed to be current and skipped. Deleting the directory and re-running the task would allow us to rebuild it if necessary.

I've also added new redirects so (at the moment) `v4` and `prerelease` both redirect to `latest`, but in theory these redirects aren't really needed because each directory should always have the latest version (corresponding to that directory) in it.

Let me know if you have any concerns with this approach.

**Target Live Date:** ASAP

**CC:** @petele


philipwalton:
My thinking is there would always be a `prerelease` folder, and it wouldn't be touched unless there's a newer prerelease version, but in this case I had to generate the initial folder for the v4 prerelease docs since there wasn't already one.

I expect that directory to not be touched by this task's logic again until we have v5 prereleases.

Since we have redirects now for `prerelease` => `latest`, it shouldn't affect anything. We could remove those redirects now if we wanted people to have access to prerelease docs (and obviously we'll need to remove them when we have v5 prerelease docs).

philipwalton:
Rebased the changes against the latest master and force pushed to resolve the merge conflicts...

PR #7548, opened 2019-04-25
Explicitly mention private methods and accessors support

mathiasbynens:


mathiasbynens:
Thanks to @domenic and @benlesh for pointing this out!

PR #7554, opened 2019-04-28
Add pseudo elements to styles forcing no motion

meduzen:
Because pseudo elements also fire `animationend` and `transitionend` events. See this [Twitter conversation](https://twitter.com/meduzen/status/1122087878895570944) for details.


meduzen:
I signed it!

Issue #7665, opened 2019-05-27
Inadequate Instructions


mudhutnm:
At URL https://developers.google.com/web/tools/chrome-devtools/beginners/html, the first Instruction says:

1. Open the _source code._ This tab will be called the editor tab throughout this tutorial.

However, the tutorial does not show where the button is to do this, and it is not evident in Chrome how to do this. Stumped on step 1.

The screen shot shows a URL at Glitch.com. Are we supposed to go to Glitch, or work in Chrome?

jpmedley:
You're supposed to follow the link. It opens Glitch in a new tab, which will be called the editor tab. Sorry about the confusion.

mudhutnm:
Hello,

 

So in the future, when we want to create a Web page on our own, we should go directly to Glitch? Is Glitch a Google product?

 

Thank you,

 

Dan Sandhaus

703-622-6822

PR #7671, opened 2019-05-28
Improve instructions.

jpmedley:
**Fixes:** #7665 
I think the problem here is a general one. In may cases, a link such as this one is tangential to the page it contains, creating the expectation the link does not need to be followed to understand the page's content. 

In this case, the link is central to what the page is trying to accomplish. This change make it clear that the link *must* be followed to start the task described on the page.

**CC:** @petele


mizansurovi:
1 banner
Our Journey Since 24.12.2012

PR #7214, opened 2019-02-06
added link to asyncCSS (advanced CSS loader)

optimalisatie:
What's changed, or what was fixed?
added a link upon the request in the article.

Link: https://github.com/style-tools/async-css/wiki/Defer-Unused-CSS-for-Google-Lighthouse-Optimization

AsyncCSS provides unique solutions to resolve the Google Lighthouse penalty while also truly achieving the best stylesheet performance for any website (no concessions for a good score).

More info: https://docs.style.tools/async-css

PR #7688, opened 2019-05-30
Update unused-css.md

optimalisatie:
PR https://github.com/google/WebFundamentals/pull/7214 was published yesterday but the link was changed.

URL: https://developers.google.com/web/tools/lighthouse/audits/unused-css
New link: https://github.com/style-tools/async

The repository has been changed in the mean time. The async loader was further improved and it combines both the stylesheet and script loader in `$async()`. By using an GCC based IIFE generator it would be possible to select only the modules that are needed and to exclude the javascript loader.

The key USP of the loader may be the ability to use a HTML `data-c` attribute with 100% JSON based control over the loader. It will enable to use the loader with complex page-based config and the fastest startup time in combination with a strict `Content-Security-Policy`.

https://github.com/style-tools/async

What's also unique is that you can capture async scripts and rewrite or remove them, for example rewrite `analytics.js` to bypass the header issues in the Google Lighthouse audit.


mizansurovi:
1 banner
Our Journey Since 24.12.2012

PR #7691, opened 2019-05-31
Payment App: Restrict "default_applications"

agektmr:
Add a note that `default_applications` URL must match the origin of where payment manifest file is hosted.

**CC:** @petele @rsolomakhin


mizansurovi:
1 banner
Our Journey Since 24.12.2012

PR #7692, opened 2019-05-31
Fixes and improvements to Subscribing a User

kyarik:

In the "Subscribing a User" article,

* Improved some sentences
* Changed the relative path passed to `navigator.serviceWorker.register()` to be absolute (i.e., changed `'service-worker.js'` to `'/service-worker.js'`) because this is indeed how it is mentioned in a paragraph within the article and also because passing a relative path is problematic since what it resolves to will depend on the page on which the registration is performed

**CC:** @petele


mizansurovi:
1 banner
Our Journey Since 24.12.2012

PR #7734, opened 2019-06-06
📖 Add more information about registration and scope

nuragic:
Add info about how change the SW scope e.g. URL without trailing slashes.

Cc @jeffposnick

### DIFF
--- a/src/content/en/ilt/pwa/introduction-to-service-worker.md
+++ b/src/content/en/ilt/pwa/introduction-to-service-worker.md
@@ -113,7 +113,7 @@ if ('serviceWorker' in navigator) {
 }
 ```
 
-This code starts by checking for browser support by examining `navigator.serviceWorker`. The service worker is then registered with `navigator.serviceWorker.register`, which returns a promise that resolves when the service worker has been successfully registered. The `scope` of the service worker is then logged with `registration.scope`.
+This code starts by checking for browser support by examining `navigator.serviceWorker`. The service worker is then registered with `navigator.serviceWorker.register`, which returns a promise that resolves when the service worker has been successfully registered. The `scope` of the service worker is then logged with `registration.scope`. If the service worker is already installed, `navigator.serviceWorker.register` returns the registration object of the currently active service worker.
 
 The `scope` of the service worker determines which files the service worker controls, in other words, from which path the service worker will intercept requests. The default scope is the location of the service worker file, and extends to all directories below. So if __service-worker.js__ is located in the root directory, the service worker will control requests from all files at this domain.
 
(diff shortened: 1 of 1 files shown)


nuragic:
This PR has been "triggered" by https://github.com/GoogleChrome/workbox/issues/2067#issuecomment-498809335

😄

src/content/en/ilt/pwa/introduction-to-service-worker.md
```diff
@@ -127,9 +127,27 @@ navigator.serviceWorker.register('/service-worker.js', {
...
 
-In this case we are setting the scope of the service worker to `/app/`, which means the service worker will control requests from pages like `/app/`, `/app/lower/` and `/app/lower/lower`, but not from pages like `/app` or `/`, which are higher.
+In this case we are setting the scope of the service worker to `/app/`, which means the service worker will control requests from pages like `/app/`, `/app/lower/` and `/app/lower/lower`, **but not from pages like `/app` or `/`**, which are higher.
+
+If you want the service worker to control higher pages e.g. `/app` (without the trailing slash) you can indeed change the scope option, but you'll also need to set the [Service-Worker-Allowed HTTP Header](https://w3c.github.io/ServiceWorker/#service-worker-script-response) in your server config for the request serving the service worker script, otherwise you'll get an error. Configuration example:
+
+#### main.js
+
+```
+navigator.serviceWorker.register('/service-worker.js', {
```
jeffposnick: If the location of the service worker script is `/service-worker.js`, then I you'd be able to set the scope to `/app` without a problem, since the service worker script is hosted at the root level. (I tested this locally in Chrome to confirm.)

In order to illustrate a scenario in which setting the scope to `/app` would fail, you'd need to use a service worker script URL hosted at a deeper level, e.g., `/app/service-worker.js`.

(The example in the service worker specification is misleading; I've filed https://github.com/w3c/ServiceWorker/pull/1422)

src/content/en/ilt/pwa/introduction-to-service-worker.md
```diff
@@ -127,9 +127,27 @@ navigator.serviceWorker.register('/service-worker.js', {
...
+  scope: '/app'
+});
+```
+
+#### Your server config (e.g. `nginx.conf`)
```
jeffposnick: Instead of (e.g. `nginx.conf`) in the heading, could you include something along the lines of this in the paragraph, before the example:

> This is an example of adding the `Service-Worker-Allowed` header to the service worker script response using the popular [nginx web server](https://www.nginx.com/resources/wiki/start/topics/examples/full/). If you're using a different web server, please refer to its documentation on setting response headers to make the equivalent change.

src/content/en/ilt/pwa/introduction-to-service-worker.md
```diff
@@ -127,9 +127,27 @@ navigator.serviceWorker.register('/service-worker.js', {
 });
 ```
 
-In this case we are setting the scope of the service worker to `/app/`, which means the service worker will control requests from pages like `/app/`, `/app/lower/` and `/app/lower/lower`, but not from pages like `/app` or `/`, which are higher.
+In this case we are setting the scope of the service worker to `/app/`, which means the service worker will control requests from pages like `/app/`, `/app/lower/` and `/app/lower/lower`, **but not from pages like `/app` or `/`**, which are higher.
+
+If you want the service worker to control higher pages e.g. `/app` (without the trailing slash) you can indeed change the scope option, but you'll also need to set the [Service-Worker-Allowed HTTP Header](https://w3c.github.io/ServiceWorker/#service-worker-script-response) in your server config for the request serving the service worker script, otherwise you'll get an error. Configuration example:
```
jeffposnick: Could you remove the following words?

> , otherwise you'll get an error. Configuration example:

I don't think they're necessarily.

nuragic:
Thanks for the review! I'll add a new commit  with these fixes tomorrow.

samdutton:
I'm really not a Service Worker expert, but LGTM.

mizansurovi:
Our Journey Since 24.12.2012

mizansurovi:
1 banner
Our Journey Since 24.12.2012

PR #7735, opened 2019-06-06
Modified triggerPushMsg() in Web Push article

kyarik:

In the "Sending Messages with Web Push Libraries" article, I changed the `triggerPushMsg` function from this

```javascript
const triggerPushMsg = function(subscription, dataToSend) {
  return webpush.sendNotification(subscription, dataToSend)
  .catch((err) => {
    if (err.statusCode === 410) {
      return deleteSubscriptionFromDatabase(subscription._id);
    } else {
      console.log('Subscription is no longer valid: ', err);
    }
  });
};
```

To this

```javascript
const triggerPushMsg = function(subscription, dataToSend) {
  return webpush.sendNotification(subscription, dataToSend)
  .catch((err) => {
    if (err.statusCode === 404 || err.statusCode === 410) {
      console.log('Subscription has expired or is no longer valid: ', err);
      return deleteSubscriptionFromDatabase(subscription._id);
    } else {
      throw err;
    }
  });
};
```

This makes the `if` statement consistent with the following paragraph from the article:
> In this example, it checks for status codes `404` and `410`,
> which are the HTTP status codes for 'Not Found' and 'Gone'.
> If we receive one of these, it means the subscription has
> expired or is no longer valid. In these scenarios, we need
> to remove the subscriptions from our database.

Also, in the `else` statement we are now throwing `err` rather than letting `undefined` be returned causing the promise returned by `triggerPushMsg()` to resolve.

We do this because in case of other errors, such as `400` (Invalid request) or `429` (Too many requests),
the promise returned by `triggerPushMsg()` should reject.

Finally, I added a paragraph which explains that in case of an error other than 404 and 410, we `throw err` to let the promise reject.

**CC:** @petele

### DIFF
--- a/src/content/en/fundamentals/push-notifications/sending-messages-with-web-push-libraries.md
+++ b/src/content/en/fundamentals/push-notifications/sending-messages-with-web-push-libraries.md
@@ -232,7 +232,7 @@ provided subscription.
       return webpush.sendNotification(subscription, dataToSend)
       .catch((err) => {
         if (err.statusCode === 404 || err.statusCode === 410) {
-          console.error('Subscription has expired or is no longer valid: ', err);
+          console.log('Subscription has expired or is no longer valid: ', err);
           return deleteSubscriptionFromDatabase(subscription._id);
         } else {
           throw err;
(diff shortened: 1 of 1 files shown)


mizansurovi:
1 banner
Our Journey Since 24.12.2012

PR #7759, opened 2019-06-10
Update Permission UX article JA translation

agektmr:
Update Permission UX article JA translation

[Check out a review request at GitLocalize](https://gitlocalize.com/repo/107/ja/review/8913)


agektmr:
This translation should be good enough to merge
cc: @petele

PR #7776, opened 2019-06-12
Add 'out of order' sign to Chacmool article.

jpmedley:


jpmedley:
@samdrazin I had to. I don't want to live a world where I never have a reason to say 'Chacmool'.

PR #7786, opened 2019-06-14
Fixes to the Common Notification Patterns article

kyarik:

In the "Common Notification Patterns" article,

- Improved some sentences
- Replaced some words with more appropriate ones
- Added missing determiners and missing punctuation
- Removed an extra "as"
- Fixed some typos

Besides all that, I also added backticks where they were missing and in some cases replaced quotes with backticks. It's worth mentioning that in the case of `'http://<Site Origin>/'`, replacing quotes with backticks was necessary because otherwise that would be interpreted as `'http://`, followed by the opening of a `<Site>` tag that has an `Origin` attribute, followed by `/'`. That indeed resulted in `'http://<Site Origin>/'` being shown as `'http:///'`.

**CC:** @petele


kyarik:
Actually, the origin already includes the scheme, so saying `http://<Site Origin>/` is not correct. We would instead need to say  `http://<Host>/`. However, I think it's better to just say `https://example.com/` since it's simpler to understand.

PR #7820, opened 2019-06-25
Fixes to the Get Started Workbox guide

kyarik:

In the *Get Started* Workbox guide,

- Changed `new RegExp('.*\.js')` to `/\.js$/` when matching JavaScript files so that the regular expression is consistent with the one written in a previous code snippet. Also, `new RegExp('.*\.js')` would be problematic because it does not escape the backslash (effectively saying `/.*.js/`) and even if it did escape the backslash, it would match not only `.js` files, but also`.json` files.
- Added missing punctuation
- Added missing determiner
- Fixed some typos

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/get-started.md
+++ b/src/content/en/tools/workbox/guides/get-started.md
@@ -3,7 +3,7 @@ book_path: /web/tools/workbox/_book.yaml
 description:Get Started with Workbox.
 
 {# wf_blink_components: N/A #}
-{# wf_updated_on: 2019-02-01 #}
+{# wf_updated_on: 2019-06-25 #}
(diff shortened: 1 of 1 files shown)


petele:
@jeffposnick - can you make sure the updated regex is OK, if it's good, feel free to approve & merge.

kyarik:
Just to put this in context, the first snippet in the guide is:

```javascript
workbox.routing.registerRoute(
  /\.js$/,
  …
);
```

and the snippet that follows it is:

```javascript
workbox.routing.registerRoute(
  new RegExp('.*\.js'),
  new workbox.strategies.NetworkFirst()
);
```

PR #7899, opened 2019-07-15
Fix issue where backslash doesn't render

kmcnellis:
https://developers.google.com/web/updates/2015/01/ES6-Template-Strings#multiline_strings

`Current solutions for them require that strings either exist on a single line or be split into multiline strings using a \ (blackslash) before each newline` 
instead of 
`Current solutions for them require that strings either exist on a single line or be split into multiline strings using a `` (blackslash) before each newline.`

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2015/01/ES6-Template-Strings.md
+++ b/src/content/en/updates/2015/01/ES6-Template-Strings.md
@@ -2,7 +2,7 @@ project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: Template String literals with embedded expressions. We've been waiting for this for ages. Literally
 
-{# wf_updated_on: 2019-03-16 #}
+{# wf_updated_on: 2019-07-16 #}
 {# wf_published_on: 2015-01-19 #}
 {# wf_tags: news,es6,javascript,ecmascript #}
 {# wf_blink_components: N/A #}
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2015/01/ES6-Template-Strings.md
```diff
@@ -80,7 +80,7 @@ If you require backticks inside of your string, it can be escaped using the back
 
 ## Multiline Strings
 
-Multiline strings in JavaScript have required hacky workarounds for some time. Current solutions for them require that strings either exist on a single line or be split into multiline strings using a `\` (blackslash) before each newline. For example:
+Multiline strings in JavaScript have required hacky workarounds for some time. Current solutions for them require that strings either exist on a single line or be split into multiline strings using a `\ ` (blackslash) before each newline. For example:
```
kyarik: There was a typo in the original sentence: blackslash should be backslash.

Also, you should fix this backslash rendering problem on line `77` as well.

Finally, are you sure this shouldn't be fixed with `` `\\` `` instead?

kmcnellis: The internal devsite markdown guide has a bit more detail.  It says:

In an inline code span, spaces immediately following the opening backtick as
well as those immediately preceding the closing backtick are not rendered. So if
an inline code span ends with a backslash (`\ `), add a space before the closing
backtick to render a literal backslash.

kmcnellis: Updated with the spelling fix, a date update, and the second backslash.

PR #7945, opened 2019-07-31
Fixes to the Payment Method Basics article

kyarik:

In the [*Payment Method Basics*](https://developers.google.com/web/fundamentals/payments/basics/payment-method-basics) article,

- Fixed a sentence that states that, in an example, two payment methods are declared, where in fact three are declared (Google Pay is missing)
- Replaced periods with colons when introducing a snippet or an image
- Added missing punctuation

**CC:** @petele


petele:
@agektmr - can you review and approve please?

PR #7964, opened 2019-08-06
Allow autoplay with sound if PWA installed on desktop

beaufortfrancois:
This PR updates the autoplay policy article by mentioning that autoplay with sound is allowed if PWA installed on desktop.


beaufortfrancois:
FYI @mounirlamouri

PR #7999, opened 2019-08-19
Clarify no install checks are needed before a badging attempt and no counter capping should be applied

tomayac:
This from a discussion with @mgiuca around https://github.com/GoogleChromeLabs/airhorn/pull/50/commits/8dd22754be0d12e8086d9d3c542ffb92688fa6bf.

**Fixes:** #issue

**Target Live Date:** any

- [X] This has been reviewed and approved by (@mgiuca)
- [X] I have run `npm test` locally and all tests pass.

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2018/12/badging-api.md
+++ b/src/content/en/updates/2018/12/badging-api.md
@@ -82,10 +82,13 @@ Examples of sites that may use this API include:
 3. Click the **Set** or **Clear** button to set or clear the badge from the app
    icon. You can also provide a number for the *Badge value*.
    
-Note: While the Badging API requires an installed app with an icon that
-actually can be badged, we excplititly advise against making calls to
-the Badging API dependent on the install state. Instead, just call it
-when it exists. If it works, it works. If not, it simply doesn't.
+Note: While the Badging API *in Chrome* requires an installed app
+with an icon that actually can be badged, we advise against
+making calls to the Badging API dependent on the install state.
+The Badging API can apply to *anywhere* a browser might want to show a badge,
+so developers shouldn't make any assumptions about what situations
+the browser will make badges work. Just call the API when it exists.
+If it works, it works. If not, it simply doesn't.
 
 ## How to use the badging API {: #use }
 
--- a/src/content/en/updates/2018/12/badging-api.md
+++ b/src/content/en/updates/2018/12/badging-api.md
@@ -3,7 +3,7 @@ book_path: /web/updates/_book.yaml
 description: The Badging API is a new web platform API that allows installed web apps to set an application-wide badge, shown in an operating-system-specific place associated with the application, such as the shelf or home screen.
 
 {# wf_published_on: 2018-12-11 #}
-{# wf_updated_on: 2019-03-07 #}
+{# wf_updated_on: 2019-08-19 #}
 {# wf_featured_image: /web/updates/images/generic/notifications.png #}
 {# wf_tags: capabilities,badging,install,progressive-web-apps,serviceworker,notifications,origintrials #}
 {# wf_featured_snippet: The Badging API is a new web platform API that allows installed web apps to set an application-wide badge, shown in an operating-system-specific place associated with the application, such as the shelf or home screen. Badging makes it easy to subtly notify the user that there is some new activity that might require their attention, or it can be used to indicate a small amount of information, such as an unread count. #}
@@ -81,6 +81,11 @@ Examples of sites that may use this API include:
    running as an installed PWA (in your task bar or dock).
 3. Click the **Set** or **Clear** button to set or clear the badge from the app
    icon. You can also provide a number for the *Badge value*.
+   
+Note: While the Badging API requires an installed app with an icon that
+actually can be badged, we excplititly advise against making calls to
+the Badging API dependent on the install state. Instead, just call it
+when it exists. If it works, it works. If not, it simply doesn't.
 
 ## How to use the badging API {: #use }
 
--- a/src/content/en/updates/2018/12/badging-api.md
+++ b/src/content/en/updates/2018/12/badging-api.md
@@ -3,7 +3,7 @@ book_path: /web/updates/_book.yaml
 description: The Badging API is a new web platform API that allows installed web apps to set an application-wide badge, shown in an operating-system-specific place associated with the application, such as the shelf or home screen.
 
 {# wf_published_on: 2018-12-11 #}
-{# wf_updated_on: 2019-08-19 #}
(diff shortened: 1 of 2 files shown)


src/content/en/updates/2018/12/badging-api.md
```diff
@@ -81,6 +81,11 @@ Examples of sites that may use this API include:
    running as an installed PWA (in your task bar or dock).
 3. Click the **Set** or **Clear** button to set or clear the badge from the app
    icon. You can also provide a number for the *Badge value*.
+   
+Note: While the Badging API requires an installed app with an icon that
+actually can be badged, we excplititly advise against making calls to
```
mgiuca: "explicitly" is misspelled.

(I would drop this word anyway. The fact that you are explicitly stating this makes it explicit :) You don't need the word "explicit" in the document.)

src/content/en/updates/2018/12/badging-api.md
```diff
@@ -81,6 +81,11 @@ Examples of sites that may use this API include:
    running as an installed PWA (in your task bar or dock).
 3. Click the **Set** or **Clear** button to set or clear the badge from the app
    icon. You can also provide a number for the *Badge value*.
+   
+Note: While the Badging API requires an installed app with an icon that
```
mgiuca: I think even this is maybe giving the wrong impression: the badging API doesn't necessarily only apply to the installed app. That's simply currently how it is in Chrome. The badging API can apply to anywhere the browser might want to show a badge. So developers shouldn't make any assumptions about what situations the browser will make badges work. Just call the API if it exists.

src/content/en/updates/2018/12/badging-api.md
tomayac: Dropped it like it's hot.

src/content/en/updates/2018/12/badging-api.md
```diff
@@ -81,6 +81,14 @@ Examples of sites that may use this API include:
    running as an installed PWA (in your task bar or dock).
 3. Click the **Set** or **Clear** button to set or clear the badge from the app
    icon. You can also provide a number for the *Badge value*.
+   
+Note: While the Badging API *in Chrome* requires an installed app
+with an icon that actually can be badged, we advise against
+making calls to the Badging API dependent on the install state.
```
kyarik: If it is advised against making calls to the Badging API based on the install state, based on what should the API be called?
If I understand this correctly, the intent here is to say that having an installed app does not guarantee that the Badging API will work, even when it's supported. Is this correct?

src/content/en/updates/2018/12/badging-api.md
```diff
@@ -81,6 +81,14 @@ Examples of sites that may use this API include:
...
+The Badging API can apply to *anywhere* a browser might want to show a badge,
+so developers shouldn't make any assumptions about what situations
+the browser will make badges work. Just call the API when it exists.
+If it works, it works. If not, it simply doesn't.
```
kyarik: I think this is a little bit too intricate. If I understand this correctly, the intent is to just say that although the Badging API is available, when it is used, it's up to the browser to decide whether it's appropriate to show the badge or not?

src/content/en/updates/2018/12/badging-api.md
kyarik: Wait, after reading the title of this pull request again, I realized that this is not what you're trying to say. Instead, you're saying that the Badging API, when available, can be called even if the app is not installed. Is this correct?

src/content/en/updates/2018/12/badging-api.md
kyarik: Again, after reading the pull request title, I realized that all you're trying to say is that the browser may show some form of badge even when the app is not installed. So, developers can call the Badging API (when available) even though their app is not installed...

src/content/en/updates/2018/12/badging-api.md
mgiuca: Yes. The Badge API should be called if the Badge API exists and you want to set a badge. Don't make assumptions about what circumstances the browser will show it. It is a "best effort" API.

src/content/en/updates/2018/12/badging-api.md
mgiuca: Yes. The badge API (ostensibly) has nothing to do with whether the app is installed or running in a window. It's just a way of saying "please associate this number with my website until further notice". The browser chooses when and how to show it; right now, we choose to show it on the app icon if it's installed, and we don't show it if it isn't installed.

src/content/en/updates/2018/12/badging-api.md
tomayac: Thanks! I hope this feedback is dealt with in https://github.com/google/WebFundamentals/pull/7999/commits/9ec4212a74fbc2a285e9d162e1acada2852ed6e5.

src/content/en/updates/2018/12/badging-api.md
tomayac: Thanks! I hope this feedback is dealt with in https://github.com/google/WebFundamentals/pull/7999/commits/9ec4212a74fbc2a285e9d162e1acada2852ed6e5.

src/content/en/updates/2018/12/badging-api.md
tomayac: Thanks! I hope this feedback is dealt with in https://github.com/google/WebFundamentals/pull/7999/commits/9ec4212a74fbc2a285e9d162e1acada2852ed6e5.

src/content/en/updates/2018/12/badging-api.md
```diff
@@ -148,14 +156,22 @@ in this case, the browser will attempt to provide the best representation for
...
 ## Feedback {: #feedback }
 
 We need your help to ensure that the Badging API works in a way that meets your
 needs and that we’re not missing any key scenarios.
 
 <aside class="key-point">
-  <b>We need your help!</b> - Will the current design (allowing an integer
-  from 1-99, 99+, or a white dot/flag) meet your needs? If it won’t, please
+  <b>We need your help!</b> - Will the current design (allowing any integer
+  that *in Chrome* gets displayed as 1–99 or 99+,
```
mgiuca: I don't think we should call out Chrome's behaviour. We don't want developers writing for Chrome, we want them writing for the API we're presenting. The decision about how the integer is displayed isn't up to Chrome, it's up to the underlying OS, so it will be different on Windows, macOS, Chrome OS, etc. That should be irrelevant to developers (as you wrote above). Just set a number and _something_ appropriate will be displayed.

All we should be asking here is "will the current design (allowing either an integer or a flag value) meet your needs".

PR #8009, opened 2019-08-22
Remove Chrome behavior mention from Feedback section

tomayac:
Acting upon @mgiuca's feedback in https://github.com/google/WebFundamentals/pull/7999#pullrequestreview-278165684.

**Target Live Date:** any

- [X] This has been reviewed and approved by (@mgiuca)
- [X] I have run `npm test` locally and all tests pass.

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2018/12/badging-api.md
+++ b/src/content/en/updates/2018/12/badging-api.md
@@ -168,12 +168,11 @@ We need your help to ensure that the Badging API works in a way that meets your
 needs and that we’re not missing any key scenarios.
 
 <aside class="key-point">
-  <b>We need your help!</b> - Will the current design (allowing any integer
-  that *in Chrome* gets displayed as 1–99 or 99+,
-  or no value that *in Chrome* gets displayed as a white dot/flag)
-  meet your needs? If it won’t, please
-  file an issue in the <a href="https://github.com/WICG/badging/issues">
-  WICG/badging repo</a> and provide as much detail as you can. In addition,
+  <b>We need your help!</b> - Will the current design
+  (allowing either an integer or a flag value) meet your needs?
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2018/12/badging-api.md
```diff
@@ -168,12 +168,11 @@ We need your help to ensure that the Badging API works in a way that meets your
...
 
 <aside class="key-point">
-  <b>We need your help!</b> - Will the current design (allowing any integer
-  that *in Chrome* gets displayed as 1–99 or 99+,
-  or no value that *in Chrome* gets displayed as a white dot/flag)
-  meet your needs? If it won’t, please
-  file an issue in the <a href="https://github.com/WICG/badging/issues">
-  WICG/badging repo</a> and provide as much detail as you can. In addition,
+  <b>We need your help!</b> - Will the current design
+  (allowing either an integer or a flag value) meet your needs?
```
kyarik: What do you mean by "flag value" here? Shouldn't this say instead "no value (which gets displayed as, e.g., a dot)"?

Anyway, I think this should be rephrased to something like:
"Will the current design, which allows you to pass either an integer value or no value at all, meet your needs?"
or
"Will the current design, which allows you to pass an optional integer value, meet your needs?"

What do you think?

CC @mgiuca @tomayac

mgiuca: We use the term "flag" to mean the Boolean presence indicator (as opposed to an integer). I see that it's used in this doc a few times as well ("another app showing a flag type badge"). Feel free to phrase this however you want; you don't need to use the word "flag" here.

I wouldn't say "no value" because "no value" means the badge isn't shown at all.

Issue #8015, opened 2019-08-23
Update KV storage to use default export


johnsonjo4531:
**Page Affected:** https://developers.google.com/web/updates/2019/03/kv-storage

**What needs to be done?**

The following page https://developers.google.com/web/updates/2019/03/kv-storage shows `storage` as a named export, but it has been moved to be a default export. All named exports for `storage` need to be changed to a default export. The `storageArea` named export is still the same so should remain unchanged.

**Question**

Do I have to sign the CLA for a markdown edit?

PR #8016, opened 2019-08-23
Update KV Storage syntax to use default export

johnsonjo4531:
As pointed out in this [pull request in the spec's GitHub repo for KV storage](https://github.com/WICG/kv-storage/pull/74). The KV storage api now uses a default export instead of a named export. This can be tested in Chrome Canary with experimental web features turned on and is also present in the current spec.

- Updated a markdown file to reflect the current KV Storage API.

**Fixes:** https://github.com/google/WebFundamentals/issues/8015


petele:
@philipwalton - can you take a look and review & merge? Thanks!

cc: @domenic

PR #8071, opened 2019-09-03
Update Surma's link for houdini intro

connerturner:
- Changed domain name for surma's houdini intro 

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2016/02/smooth-scrolling-in-chrome-49.md
+++ b/src/content/en/updates/2016/02/smooth-scrolling-in-chrome-49.md
@@ -2,7 +2,7 @@ project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: Tired of jittery scrolling? Great, because Chrome 49 is shipping with a new smooth scroll right out of the box!
 
-{# wf_updated_on: 2016-02-01 #}
+{# wf_updated_on: 2019-09-03 #}
 {# wf_published_on: 2016-02-01 #}
 {# wf_tags: scroll,chrome49 #}
 {# wf_featured_image: /web/updates/images/2016/02/smooth-scrolling-in-chrome-49/smooth-scroll.png #}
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2016/02/smooth-scrolling-in-chrome-49.md
```diff
@@ -37,7 +38,7 @@ There are other scroll-related goodies in the works that are also worth mentioni
...
+Compositor Worklets are part of [Houdini](https://wiki.css-houdini.org/), and are yet to be fully spec’d out and implemented. That said, [as the patches land](http://crbug.com/436952), they will allow you to write JavaScript that’s run as part of the compositor’s pipeline, which in general means that scroll-coupled effects like parallaxing will be kept perfectly in sync with the current scroll position. Given the way that scrolling is handled today, where scroll events are only periodically sent to the main thread (and can be blocked by other main thread work), this would represent a huge leap forward. If you’re interested in Compositor Worklets, or any of the other exciting new features that Houdini brings, look over the [Intro to Houdini post by Surma](https://dassur.ma/things/houdini-intro/), the [Houdini specs](https://drafts.css-houdini.org/), and contribute your thoughts to the [Houdini mailing list](https://lists.w3.org/Archives/Public/public-houdini/)!
```
petele: Hey @robdodson & @mfriesenhahn - I don't think web.dev enforces line breaks in MD docs. On WebFu we have, and this is why. On longer lines, where the changes are minor, Git & other diff tools don't highlight changes, slowing down the review process. If I knew exactly what was changing, I could do it super quick.

robdodson: yeah I bet we could add this to remark-lint. @mfriesenhahn wdyt?

mfriesenhahn: I do have one reservation. The other problem we run into is that when people break their lines at the 80 character mark, it subdivides semantic units, which means making multiple GH suggestions for a single fix--which is often confusing. That seems to come up more often than this problem.

Maybe we could tell people to break lines by semantic units (phrases, clauses) and allow some flexibility on line length? (E.g., give people that guidance and set remark to whatever the upper limit is for change highlighting.)

mfriesenhahn: Here's an example of what I mean by breaking lines at semantic units:

```
For users on slow connections,
such as 2G, 3G, or slow Wi-Fi,
third-party scripts dynamically injected via `document.write()`
can delay the display of main page content by tens of seconds.
```

It can still lead to multiple suggestions for a multi-phrase fix, but at least multiple edits to fix a single phrase will will be avoided.

robdodson: I like that idea (I use it for code comments) but I don't think there would be a way for us to automatically enforce it... I guess I'll defer to the writer/reviewers on this one. If they want us to change the linting we can add the 80 characters thing. Or, as Michael mentioned, if it's nice because you can keep the suggestion in a single comment then we can stick with what we have now. I don't have a strong feeling either way.

PR #8100, opened 2019-09-09
Fixes to the LayoutNG article

kyarik:

In the [*LayoutNG*](https://developers.google.com/web/updates/2019/06/layoutNG) article,

- Added missing periods/commas.
- Fixed a broken word due to a line-break.
- Removed hyphens in "For elements that size-to-content".
- Removed underscores from `_T_`, which were probably added to make the T italic, but this doesn't work because we're inside of a block-level element (`<figcaption>`). However, rather than wrapping the T in `<i>`, it's probably better to leave it roman type so it's consistent with the way it is mentioned in a paragraph above.
- Removed the misplaced space in "Layout NG".

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2019/06/layoutNG.md
+++ b/src/content/en/updates/2019/06/layoutNG.md
@@ -3,7 +3,7 @@ book_path: /web/updates/_book.yaml
 description: A new layout engine for Chromium
 
 {# wf_published_on: 2019-06-06 #}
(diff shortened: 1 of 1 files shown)


ESCHESLOW:
Edward Cheslow
escheslow2@gmail.com
DONE. FINISHED THANKS

PR #8177, opened 2019-09-27
Add notes on Google Play Signing to TWA Quickstart.

PEConn:
Adds a section about how to generate a Digital Asset Links file for users of Google Play Signing.


PEConn:
@petele Could you please have a look over this?

PR #8178, opened 2019-09-27
Updating to the correct way to add audio to an element from a file

PaulKinlan:
cc @beaufortfrancois 

**CC:** @petele


beaufortfrancois:
Shall we add a comment as well to tell web dev that they should call revokeObjectURL as well when they are done?

PR #8190, opened 2019-10-02
Fixes to the Cookie Store API article

kyarik:
In the *Asynchronous Access to HTTP Cookies* article,

- Removed an extra "the".
- Replaced incorrect `in` with `of` when iterating over changed cookies and deleted cookies.
- Appended a missing "s" to make "service worker" plural.

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2018/09/asynchronous-access-to-http-cookies.md
+++ b/src/content/en/updates/2018/09/asynchronous-access-to-http-cookies.md
@@ -2,7 +2,7 @@ project_path: /web/_project.yaml
 book_path: /web/updates/_book.yaml
 description: The Cookie Store API offers asynchronous access to HTTP cookies, and opens up the cookie jar to service workers.
 
-{# wf_updated_on: 2018-11-29 #}
+{# wf_updated_on: 2019-10-02 #}
 {# wf_published_on: 2018-09-06 #}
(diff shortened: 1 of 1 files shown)


src/content/en/updates/2018/09/asynchronous-access-to-http-cookies.md
```diff
@@ -146,11 +146,11 @@ The Cookie Store API brings an alternative method for observing cookie
 changes, which does not require polling.
 
     cookieStore.addEventListener('change', (event) => {
-      for (const cookie in event.changed) {
+      for (const cookie of event.changed) {
```
petele: @surma - can you comment on the `of`/`in` difference?

surma: If `changed` is an array, `in` would iteration over the _indices_, while`of` would iterator over the items in the array. Looking at the body, `of` seems indeed to be intended here.

kyarik: Indeed, `event.changed` and `event.deleted` are arrays. Using `for...in` would iterate over all the enumerable properties of the array, which in this case are just the indexes. While using `for...of` would iterate through all the values of the array, which is what was intended if we look at the body of the `for` loop.

PR #8235, opened 2019-10-17
Improve *What's new in DevTools (Chrome 79)* article

sigurdschneider:
- Clarify which search box got added
- Clarify that coverage only displays if it is available


kaycebasques:
LGTM, thanks for the update

PR #8291, opened 2019-11-18
Replace getQueue() with defaultQueue in Web GPU article

beaufortfrancois:
In an effort to support multi-queue with WebGPU, the shape of the API has changed in the [spec side](https://github.com/gpuweb/gpuweb/pull/490). This PR replaces `getQueue()` with `defaultQueue` in the Web GPU article so that web developers can continue to experiment with GPU Compute in Chrome Canary.

FYI @kangz


beaufortfrancois:
There's no "implicity" but "simplicity" in the article, therefore I think we're good.

PR #8310, opened 2019-11-27
Add What's New In DevTools (Chrome 80)

kaycebasques:
**Target Live Date:** 2019-12-05

- [x] This has been reviewed and approved by @kaycebasques 


src/content/en/updates/2019/12/devtools.md
```diff
@@ -0,0 +1,232 @@
...
+
+{# https://chromium.googlesource.com/devtools/devtools-frontend/+/f05ffb295a3a4c36f8a5154285f6641a87f1f8b5 #}
+
+Chromium issue [#1029031](https://crbug.com/1029031)
+
+## Audits panel updates {: #audits }
+
+### Plugins {: #plugins }
+
+TODO
```
kaycebasques: TODO

src/content/en/updates/2019/12/devtools.md
```diff
@@ -0,0 +1,232 @@
...
+{# wf_featured_snippet: Support for let redeclarations in the Console, improved WebAssembly debugging, and more. #}
+{# wf_blink_components: Platform>DevTools #}
+
+# What's New In DevTools (Chrome 80) {: .page-title }
+
+{% include "web/_shared/contributors/kaycebasques.html" %}
+
+## Support for `let` redeclarations in the Console {: #let }
+
+The Console now supports redeclarations of `let` statements. The inability to redeclare
```
mathiasbynens: `let` and `class`

src/content/en/updates/2019/12/devtools.md
```diff
@@ -0,0 +1,232 @@
...
+`let` statements was a common annoyance for web developers who use the Console to experiment with
+new JavaScript code.
+
+<aside class="warning">
+  Redeclaring a <code>let</code> statement in a script will still cause a
```
mathiasbynens:
```suggestion
  Redeclaring a <code>let</code> statement in a script outside of the Console or within a single Console input still causes a
```

src/content/en/updates/2019/12/devtools.md
```diff
@@ -0,0 +1,232 @@
...
+## Improved WebAssembly debugging {: #webassembly }
+
+DevTools has started to support the [DWARF Debugging Standard](http://dwarfstd.org/),
+which means that you can now step over code, set breakpoints, and resolve stack traces in your
+source languages within DevTools.
+Check out [Improved WebAssembly debugging in Chrome DevTools](/web/updates/2019/12/webassembly)
+for the full story.
+
+<figure>
+  <img src="../../images/2019/12/dwarf.png"
```
mathiasbynens: This screenshot looks different than all the others. Can we get one in the same style? @rreverser, do you have a demo link handy that can be used to reproduce this?

src/content/en/updates/2019/12/devtools.md
```diff
@@ -0,0 +1,232 @@
...
+</figure>
+
+Right-click the **Waterfall** table header and select **Path** or **URL** to show the new
+columns.
+
+{# https://chromium.googlesource.com/chromium/src/+/923cd866fb089619bbbd971ff16cd7d766c1fb67 #}
+
+Chromium issue [#993366](https://crbug.com/993366)
+
+### Updated User-Agent string for Firefox {: #firefox }
```
mathiasbynens: There's some open CLs pending review here to update UA strings for other browsers as well, both in the Network Conditions tab and in the Device Toolbar. I'm hopeful that by the M80 branch point, they'll have landed.

IMHO for this section we could just state something like:

```md
### Updated User-Agent strings

DevTools supports setting a custom User-Agent string through the Network Conditions tab but also through the Device Toolbar (as part of device emulation). The User-Agent string affects the `User-Agent` HTTP header attached to network resources, but also the value of `navigator.userAgent`.

The User-Agent strings in our predefined lists have been updated to reflect modern browser versions.
```

src/content/en/updates/2019/12/devtools.md
```diff
@@ -0,0 +1,232 @@
...
+
+Use the new **Path** and **URL** columns in the **Network** panel to see the absolute path
+or full URL of each network resource.
+
+<figure>
+  <img src="../../images/2019/12/columns.png"
+       alt="A screenshot of the new Path and URL columns in the Network panel."/>
+</figure>
+
+Right-click the **Waterfall** table header and select **Path** or **URL** to show the new
```
ghodir: This should work with any column header, not just the waterfall.

src/content/en/updates/2019/12/devtools.md
RReverser: Note that this is a separate post, and screenshots within it are consistent... but yeah, they're not from macOS like the usual screenshots seem to be.

I don't have a demo link. I feel like I've had to answer this question quite frequently lately 😅 unfortunately, it's not possible to provide one, because DWARF uses local file paths and so debugging would work only on the same machine, at least until we implement remapping UI or something.

But, given that this is in Chromium and if you have Mac, I'm happy to sit down with you and remake the screenshots.

RReverser: Ah, in this context it's actually not a separate post, so yeah, def need to change.

src/content/en/updates/2019/12/webassembly.md
```diff
@@ -0,0 +1,84 @@
...
+</figure>
+
+However, source maps were designed for text formats with clear mappings to JavaScript concepts and values, not for binary formats like WebAssembly with arbitrary source languages, type systems, and a linear memory. This made the integration hacky, limited, and not widely supported outside Emscripten.
+
+## Enter DWARF
+
+On the other hand, many native languages already have a common debugging format,
+[DWARF](http://dwarfstd.org/), that provides all the necessary information for debuggers to resolve locations, variable names, type layouts, and more.
+
+While there are still some WebAssembly-specific features that need to be added for full compatibility, compilers like Clang and Rust already support emitting DWARF information in WebAssembly modules, which enabled the DevTools team to start integrating DWARF information
```
RReverser: Hmm I'm not sure "integrating DWARF information" makes sense as it's just information and is part of user-supplied Wasm modules, not DevTools. I'd leave this either as in original draft

```suggestion
While there are still some WebAssembly-specific features that need to be added for full compatibility, compilers like Clang and Rust already support emitting DWARF information in WebAssembly modules, which enabled the DevTools team to start using it
```

or maybe say

```suggestion
While there are still some WebAssembly-specific features that need to be added for full compatibility, compilers like Clang and Rust already support emitting DWARF information in WebAssembly modules, which enabled the DevTools team to start integrating DWARF support
```

or something similar.

src/content/en/updates/2019/12/webassembly.md
```diff
@@ -0,0 +1,84 @@
...
+directly into DevTools.
+
+As a first step, the Chrome DevTools team implemented native source mapping using this information, so you can start debugging Wasm modules produced by any of these compilers without resorting to the disassembled format or having to use any custom scripts.
```
RReverser: I don't know if anyone cares about these details... but I'm not on the Chrome DevTools team, so I just carefully said "we" (as a wider team) in the draft 😅 Maybe worth keeping that way?

```suggestion
As a first step, we implemented native source mapping using this information, so you can start debugging Wasm modules produced by any of these compilers without resorting to the disassembled format or having to use any custom scripts.
```

src/content/en/updates/2019/12/webassembly.md
```diff
@@ -0,0 +1,84 @@
...
+Instead, you just need to tell your compiler to include debug info like you normally would on other platforms. For example, in Clang this can be done by passing the `-g` flag during compilation:
```
RReverser: I'd add a newline / start a new paragraph here:

```suggestion

Instead, you just need to tell your compiler to include debug info like you normally would on other platforms. For example, in Clang this can be done by passing the `-g` flag during compilation:
```

src/content/en/updates/2019/12/webassembly.md
```diff
@@ -0,0 +1,84 @@
...
+
+    clang -g ...sources… -target wasm32 -o out.wasm
+
+You can use same `-g` flag in Rust:
+
+    rustc -g source.rs --target wasm32-unknown-unknown -o out.wasm
+
+If you’re using Cargo, the debug info will be included by default:
```
RReverser: Cargo is also Rust (just a more general builder), so this sentence is meant to be a continuation of the last one. I probably haven't made that very clear in the draft, so maybe like this it will work better:

```suggestion
...or, if you’re using Cargo, the debug info will be included by default:
```

src/content/en/updates/2019/12/webassembly.md
```diff
@@ -0,0 +1,84 @@
...
+## The future
+
+There is still quite a bit of work to do though.
+For example, on the tooling side, Emscripten (Binaryen) and wasm-pack (wasm-bindgen) don’t support updating DWARF information on transformations they perform yet. For now, they won’t benefit from this integration. 
+
+And on the Chrome DevTools side, we’ll be evolving integration more over time to ensure a seamless debugging experience, including: 
+
+* Resolving variable names
+* Pretty-printing types
+* Evaluating expressions in source languages
```
RReverser:
```suggestion
* Evaluating expressions in source languages
* ...and much more!
```

(as these are just examples)

src/content/en/updates/2019/12/devtools.md
```diff
@@ -0,0 +1,232 @@
...
+</figure>
+
+{# https://chromium.googlesource.com/devtools/devtools-frontend/+/53a8548ce03957e7704ae88b780dfa019bf967e0 #}
+
+Chromium issue [#1004193](https://crbug.com/1004193)
+
+## Improved WebAssembly debugging {: #webassembly }
+
+DevTools has started to support the [DWARF Debugging Standard](http://dwarfstd.org/),
+which means that you can now step over code, set breakpoints, and resolve stack traces in your
```
RReverser: Hmm... note sure if / how to clarify this, but should we make a distinction that you could do all that with source maps before, but in a more limited manner?

src/content/en/updates/2019/12/devtools.md
kaycebasques: @ghodir yes that was the old functionality that I was familiar with. Looks like a regression has been introduced, because it currently only works on the **Waterfall**.

src/content/en/updates/2019/12/devtools.md
kaycebasques: Changed to `increased support`

src/content/en/updates/2019/12/devtools.md
kaycebasques: Thanks for the updated screenshot @mathiasbynens & @RReverser

mathiasbynens:
> :tada: This has been pushed live to https://developers.google.com/web/

Where though? I was expecting <https://developers.google.com/web/updates/2019/12/devtools> but that's a 404.

mathiasbynens:
Never mind, it's live now.

PR #8335, opened 2019-12-10
Update `cache-control-header.png` for correction of an incorrect CC value

Malvoz:
https://developers.google.com/web/tools/lighthouse/audits/cache-policy has an image which displays an incorrect cache-control value `must_revalidate`, it should be `must-revalidate`. I've modified the image (it's also 8KB lighter, wihoo) to correct this. I have also updated the reference to cache-bust the image, such that it is reflected on the live page:

```diff
-<img src="images/cache-control-header.png"
+<img src="images/cache-control-header.png?v=2"
```


kaycebasques:
Thank you for the fix. This guide now lives on web.dev so I have copied the new optimized image over to that repository: https://github.com/GoogleChrome/web.dev/pull/3134

PR #8501, opened 2020-02-20
$ no longer an alias for document.querySelector

KennethKinLum:
It could be in the past, the function is an alias for `document.querySelector`, but if the statement  $: "this function is an alias for the document.querySelector() function" is passed around, it might cause some confusion...

```
// not exactly an alias
document.querySelector === $   // false
```   

The real situation is, when called with one argument, this function (`$`) calls `document.querySelector()`
But when called with `$(sel, node)`, calls `node.querySelector(sel)`, so it would not be accurate to call this an alias. I think in the past, it doesn't accept a second argument, and it was an alias for `document.querySelector()` but not any more.

- item 1
- item 2

**Fixes:** #issue

**CC:** @petele

### DIFF
--- a/src/content/en/tools/chrome-devtools/console/utilities.md
+++ b/src/content/en/tools/chrome-devtools/console/utilities.md
@@ -64,7 +64,7 @@ while `$1` returns the previously selected one:
 
 `$(selector)` returns the reference to the first DOM element
 with the specified CSS selector.
-This function is an alias for the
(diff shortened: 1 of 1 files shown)


petele:
@kaycebasques - can you take a look at this please?

kaycebasques:
I'll also add a paragraph to explain the usage with 2 arguments.

PR #8633, opened 2020-04-14
Added an option to enable in GitLocalize special front matter treatment

svasilenkov:


svasilenkov:
@googlebot I signed it!

PR #8651, opened 2020-04-24
[GitLocalize] Advise contributors that translations are no longer rev…

ilyaspiridonov:
Contributors will now see a pop up saying that translations are no longer reviewed.


ilyaspiridonov:
@agektmr Please check if the wording is OK 
thank you!

PR #8680, opened 2020-05-01
Use matchCallbacks instead of RegExps in Workbox docs

jeffposnick:
Fixes https://github.com/GoogleChrome/workbox/issues/2486 by using `matchCallback` functions with appropriate predicates instead of `RegExp` objects in most places in our docs. (`RegExp` objects are still fully supported, but an explicit `matchCallback` can be cleaner.)

R: @petele @philipwalton 
CC: @tomayac because I think he suggested we do this a while ago 😄


tomayac:
Non-owner LGTM. A well-received simplification… :-)

PR #8686, opened 2020-05-05
Redirect Lighthouse pages to new resources

paulirish:
- Redirect from the old scoring docs to the new one on web.dev
- Simplify the TOC entirely and point to the new web.dev docs

At this point the audit docs on DGC are outdated/deprecated and we now consider all the web.dev locations to be the canonical resources.

**Target Live Date:** whenever

**CC:** @petele


src/content/en/tools/_redirects.yaml
```diff
@@ -24,6 +24,12 @@ redirects:
 - from: /web/tools/chrome-devtools/console/console-reference
   to: /web/tools/chrome-devtools/console/api
 
+- from: /web/tools/lighthouse/v3/scoring
```
petele, line 27: This redirect will not work because the original content is still there. DevSite will only redirect if the path has been removed.

src/content/en/tools/_redirects.yaml
```diff
@@ -24,6 +24,12 @@ redirects:
...
+  to: https://web.dev/performance-scoring/
+  
+- from: /web/tools/lighthouse/scoring
```
petele, line 30: This redirect will not work because the original content is still there. DevSite will only redirect if the path has been removed.

src/content/en/tools/lighthouse/_toc.yaml
```diff
@@ -1,183 +1,7 @@
...
-  - title: Links Do Not Have Descriptive Text
-    path: /web/tools/lighthouse/audits/descriptive-link-text
-  - title: Page has unsuccessful HTTP status code
-    path: /web/tools/lighthouse/audits/successful-http-code
-  - title: Page is blocked from indexing
-    path: /web/tools/lighthouse/audits/indexing
-  - title: robots.txt is not valid
-    path: /web/tools/lighthouse/audits/robots
-  - title: Tap targets are not sized appropriately
-    path: /web/tools/lighthouse/audits/tap-targets
```
petele, line 183: You've removed this from the TOC, but left the content. That means it will still be online and people can get to it via existing links, search, etc. There's no indication on the page that it now lives on web.dev, or that it may be out of date. And, because it's not in the TOC, it won't show up in the left hand nav, and provide no way for a user to navigate away from it.

src/content/en/tools/lighthouse/_toc.yaml
```diff
@@ -1,183 +1,7 @@
...
+- title: Lighthouse Scoring Guide
+  path: https://web.dev/performance-scoring/
```
petele, line 5: please add `status: external` to this so folks know they're about to be redirected to an external site.

src/content/en/tools/lighthouse/_toc.yaml
```diff
@@ -1,183 +1,7 @@
...
+- title: Audit Documentation
+  path: https://web.dev/learn/#lighthouse
```
petele, line 7: please add `status: external` to this so folks know they're about to be redirected to an external site.

src/content/en/tools/lighthouse/_toc.yaml
paulirish: for posterity: kayce fixed this and the following comments in https://github.com/google/WebFundamentals/pull/8691

PR #8707, opened 2020-05-14
Update to the Issues tab snippet

danieltxok:
1. Changed 'was built' for 'aims', I don't want to give the impression that the tab is in its final stage.
2. 'aggregates' was repetitive.

**CC:** @petele @samdutton @kaycebasques


samdutton:
@kaycebasques OK to merge this and get these changes published, along with updates from #8709?

PR #8720, opened 2020-05-21
Fix one zh-cn translatation error

JadeTao:
correct the translation of the word 'implicit' from '显式' to '隐式'

What's changed
- line#296

en:
https://github.com/google/WebFundamentals/blob/master/src/content/en/fundamentals/performance/http2/index.md
"Neither stream A nor B specifies a parent dependency and are said to be dependent on the implicit "root stream";"

wrong zh-cn:
https://github.com/google/WebFundamentals/blob/master/src/content/zh-cn/fundamentals/performance/http2/index.md
"数据流 A 和数据流 B 都没有指定父依赖项，依赖于显式“根数据流”；"

The word 'implicit' was translated to '显式', which means 'explicite' in English.

I just corrected this word.


petele:
Hey @JadeTao - Thanks for the PR. It looks like the whole file was changed, not just the one thing you mentioned. Can you update the PR so that only the actual content that changes was changed please?

JadeTao:
Hi @petele. Finally I get it done.
windows `crlf` and wierd`^M` string caused this problem. now it's gone.

petele:
Awesome, thanks @JadeTao!

PR #8733, opened 2020-05-27
Lighthouse tools: saving results to the cloud

adamhenson:
- The page of the docs this change affects lists different ways of using Lighthouse. It doesn't describe a way to maintain a historical record of Lighthouse audits, so this change illustrates one example of doing this (with Automated Lighthouse Check)[https://www.automated-lighthouse-check.com/].
- This change links to [the list of many Lighthouse integrations](https://github.com/GoogleChrome/lighthouse#lighthouse-integrations) which deserves a spot on this page of the docs.

**CC:** @petele


petele:
Thanks for the PR. We don't list any Lighthouse integrations on this page and I'd prefer to keep it clean and simple. It looks like Automated Lighthouse Check is already listed on the integrations page in the Git repo.

adamhenson:
Fair enough, thanks.

adamhenson:
I expected this, however I should point out PageSpeed Insights is most certainly an integration @petele.

PR #8753, opened 2020-06-03
Redirects to newer WebPerf content on web.dev

addyosmani:
We've discussed redirecting older WebPerf content we wrote here to web.dev. I'd like to propose we do this for at least a few more of the guides in /performance. 

**CC:** @petele @philipwalton 

Pete, feel free to let me know if there are other steps needed here (I think you may have once mentioned needing to remove the content from the corresponding directories, but I may be wrong...)

### DIFF
--- a/src/content/en/_redirects.yaml
+++ b/src/content/en/_redirects.yaml
@@ -73,6 +73,18 @@ redirects:
 - from: /web/fundamentals/security/encrypt-in-transit/why-https
   to: https://web.dev/why-https-matters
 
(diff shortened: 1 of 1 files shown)


petele:
Hey @addyosmani & @philipwalton - DevSite redirects only work if the requested content does not exist. So **in addition to creating the redirect, you must delete the content as well**. Be sure to check that you've deleted all languages as well.

I'll file a PR to fix now, but please keep this in mind for the future.

PR #8758, opened 2020-06-05
Prevent versioned Workbox docs from being indexed

jeffposnick:
R: @philipwalton @petele 

Fixes https://github.com/GoogleChrome/workbox/issues/2513

This adds a new `gulp workbox` step that will ensure that versioned reference docs aren't indexed.

This PR also includes the results of running the command, i.e. the versioned HTML includes the appropriate `<meta>` tag.

(I tested to confirm it can be run multiple times without adding in extra tags.)


jeffposnick:
Hmm, I'm getting

Screen Shot 2020-06-07 at 12 03 22 PM

on the staging. Looking into it more, but please hold off on merging for the time being.

jeffposnick:
Okay, the issue was fixed by @petele in https://github.com/google/WebFundamentals/pull/8765

@philipwalton this can be merged if it looks good to you.

PR #8820, opened 2020-06-22
Remove & redirect `/web/shows/`

petele:
- Remove and redirect shows


petele:
Note, this leaves http203 and designer-vs-developer since they also include podcasts.

PR #9115, opened 2020-11-13
[devtools-blog] Simulate CVD in the Blink Renderer

chybie:
- [x] This has been reviewed and approved by (@chybie)
- [x] I have run `npm test` locally and all tests pass.
- [x] I have added the appropriate `type-something` label.
- [x] I've staged the site and manually verified that my content displays correctly.

PR #9116, opened 2020-11-13
Support webp and avif image formats in WebFu

chybie:
@petele, can we support webp and avif image formats in WebFu? Will that involve any other server side changes that we might want to avoid till the new site in?

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -32,10 +32,10 @@ handlers:
   http_headers:
     x-frame-options: SAMEORIGIN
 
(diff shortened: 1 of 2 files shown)


chybie:
fyi @mathiasbynens

mathiasbynens (approved):

mathiasbynens:
@petele could you PTAL? We could really use this for #9115.

chybie:
@petele , merge this as per the reply.

PR #9135, opened 2020-11-23
Chromium Chronicle Ep 15

petele:
Chromium Chronicle ep 15


JoeNotCharles:
The internal version of this has a footer with some links that could be included here:

https://gn.googlesource.com/gn/+/master/docs/reference.md#var_visibility - All the patterns you can use to restrict visibility in GN.
https://chromium.googlesource.com/chromium/src/+/master/buildtools/checkdeps/README.md - Complete documentation of DEPS files and the checkdeps tool.

PR #9202, opened 2021-01-13
Update guidance to use the pagehide event over unload

rowan-m:
https://developers.google.com/web/updates/2020/05/heavy-ad-interventions suggests using the `unload` and `beforeunload` events to report on interventions. This conflicts with the advice to never use `unload` and use `pagehide` instead in https://web.dev/bfcache/#never-use-the-unload-event

The article demo has been updated to make use of the `pagehide` event and warn against `unload` and `beforeunload`.

**CC:** @petele

### DIFF
--- a/src/content/en/updates/2020/05/heavy-ad-interventions.md
+++ b/src/content/en/updates/2020/05/heavy-ad-interventions.md
@@ -188,8 +188,8 @@ window.addEventListener('pagehide', (event) => {
 });
 ```
 
-Remember the `pagehide` event restricts the amount of work that can happen
-within it to protect the user experience. For example, trying to send a
+Remember that, to protect the user experience, the `pagehide` event restricts 
+the amount of work that can happen within it. For example, trying to send a
 `fetch()` request with the reports will result in that request being canceled.
 You should use `navigator.sendBeacon()` to send that report and even then, this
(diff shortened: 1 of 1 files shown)


rowan-m:
Low priority update, so just good to go out whenever there's a push.

rowan-m:
fyi @mihajlija

src/content/en/updates/2020/05/heavy-ad-interventions.md
```diff
@@ -178,21 +178,25 @@ observer.observe();
...
 });
 ```
 
-Caution: The `unload` and `beforeunload` events both restrict the amount of work
-that can happen within them to protect the user experience. For example, trying
-to send a `fetch()` request with the reports will result in that request being
-canceled. You should use `navigator.sendBeacon()` to send that report and even
-then, this is only best-effort by the browser not a guarantee.
+Remember the `pagehide` event restricts the amount of work that can happen
+within it to protect the user experience. For example, trying to send a
```
mihajlija, line 192:
```suggestion
Remember that, to protect the user experience, the `pagehide` event restricts 
the amount of work that can happen within it. For example, trying to send a
```

mihajlija (approved):

petele (approved):

PR #9221, opened 2021-01-20
[devtools] Add missing image

chybie:


bmeurer (approved):

tomayac:
Arrived here too late to approve, but thanks for reacting on my bug report so quickly! `<3`

PR #9222, opened 2021-01-20
Fix typo in `rel` attribute value

mathiasbynens:


petele:
These docs have been migrated to dcc. I've created a PR there (https://github.com/GoogleChrome/developer.chrome.com/pull/287) to address this.

Since the docs no longer live here, I'm going to close this PR.

PR #9236, opened 2021-01-28
Replace defaultQueue with queue

beaufortfrancois:
This PR is about replacing the soon to be deprecated `defaultQueue` attribute with `queue` in the GPU Compute article.
See https://chromium-review.googlesource.com/c/chromium/src/+/2651406

I'll update samples as well.

@kangz Can you review?

@petele Can you merge once it's approved by @Kangz?


Kangz (approved):
LGTM but maybe we should batch these changes at the same time as we do the PSA for other people using WebGPU?

beaufortfrancois:
When is the next batch happening?

Kangz:
A couple weeks most likely?

beaufortfrancois:
> A couple weeks most likely?

Let's hold off then.

beaufortfrancois:
It seems it's going to the only change in the next batch that will impact this article, so as discussed offline with @kangz, @petele you can merge and push live.

petele (approved):

PR #9269, opened 2021-02-24
Update advanced-recipes.md

darvid7:
Add `await` in the `messageSW` call. `messageSW` returns a promise https://github.com/GoogleChrome/workbox/blob/v6/packages/workbox-window/src/messageSW.ts, I assume that the onAccept handler is async because of this await.

- adds an await

Just changes docs so omittting the rest of the template.

**CC:** @petele

### DIFF
--- a/src/content/en/tools/workbox/guides/advanced-recipes.md
+++ b/src/content/en/tools/workbox/guides/advanced-recipes.md
@@ -48,7 +48,7 @@ if ('serviceWorker' in navigator) {
           // instructing it to activate.  
           // Note: for this to work, you have to add a message
           // listener in your service worker. See below.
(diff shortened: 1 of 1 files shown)


jeffposnick:
Thanks for your contribution!

In this particular case, we don't want to `await` the return value, because the service worker in the example doesn't generate a response. As per the [`messageSW` documentation](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-window.Workbox#messageSW), the promise that `messageSW()` returns in that case would never resolve.

(I don't remember why we originally had an `async` around the `onAccept` function, but you're right in that it's not necessary as written.)

darvid7:
ah I missed that detail in `messageSW` thanks for the context! I'll make another PR to remove the async because I found it a bit confusing when I read reading the example.

PR #9272, opened 2021-02-26
Update advanced-recipes.md

darvid7:
Remove `async` from onAccept.

context is https://github.com/google/WebFundamentals/pull/9269#issuecomment-786367988

**CC:** @petele


jeffposnick (approved):
Thanks!

PR #9284, opened 2021-03-03
Set up redirects to DCC for Android content

andreban:
- Deletes web on Android content.
- Updates footer.
- Sets up redirects to new locations.

**Update:** Content has been deployed to https://developer.chrome.com/docs/android/

~~**Important**: This is a draft PR, and can only be deployed after https://github.com/GoogleChrome/developer.chrome.com/pull/470 is live.~~

**CC:** @petele


src/content/en/_redirects.yaml
```diff
@@ -429,3 +429,56 @@ redirects:
 
 - from: /web/updates/2021/01/nic88
   to: https://developer.chrome.com/blog/new-in-chrome-88/
+
+## Redirects for Web-on-Android content migrated to DCC
+
+- from: /web/android
+  to: https://developer.chrome.com/docs/android/
```
petele, line 436: I'd suggest moving this one to the end of the web on android redirects, and change it to:

```suggestion
- from: /web/android/...
  to: https://developer.chrome.com/docs/android/
```

That way, hard links will redirect to their appropriate place, and all other links will redirect to the dcc/docs/android/

andreban: Good idea! Accepted the suggestion!

petele (approved):

PR #9304, opened 2021-03-12
Make HTML examples valid

mathiasbynens:
Thanks to @zcorpan for spotting this and [pointing it out](https://twitter.com/zcorpan/status/1370336807519461377)!


src/content/en/updates/2020/11/cvd.md
```diff
@@ -293,25 +299,27 @@ Let’s start with part 2: how can we avoid adding the SVG to the DOM? One idea
...
 
-The first change is the XML namespace declaration at the top. The second addition is the so-called “solidus” — the slash that indicates the `<feColorMatrix>` tag both opens and closes the element. The HTML parser couldn’t care less about `/>` vs. `>`, but in XML the difference matters.
+The first change is the XML namespace declaration at the top. The second addition is the so-called “solidus” — the slash that indicates the `<feColorMatrix>` tag both opens and closes the element. This last change is not actually necessary (we could just stick to the explicit `</feColorMatrix>` closing tag instead), but since XML supports this `/>` shorthand (unlike HTML), we might as well make use of it.
```
zcorpan, line 322: This is still incorrect--HTML does support /> on SVG elements, so you can use that syntax both in HTML and XML.

mathiasbynens, line 322: Ooh, thanks! So how about this then:

```suggestion
The first change is the XML namespace declaration at the top. The second addition is the so-called “solidus” — the slash that indicates the `<feColorMatrix>` tag both opens and closes the element. This last change is not actually necessary (we could just stick to the explicit `</feColorMatrix>` closing tag instead), but since both XML and SVG-in-HTML support this `/>` shorthand, we might as well make use of it.
```

mathiasbynens: Ooh, thanks! So how about this then:

```suggestion
The first change is the XML namespace declaration at the top. The second addition is the so-called “solidus” — the slash that indicates the `<feColorMatrix>` tag both opens and closes the element. This last change is not actually necessary (we could just stick to the explicit `</feColorMatrix>` closing tag instead), but since both XML and SVG-in-HTML support this `/>` shorthand, we might as well make use of it.
```

zcorpan: Yeah, now it's correct.

chybie (approved):

PR #9344, opened 2021-04-27
Remove workbox-precaching from recipes guide

jeffposnick:
This removes a reference to `workbox-precaching` from the `workbox-recipes` docs, as it overcomplicates the recipe and is not strictly necessary as of Workbox v6.1.0.

This also adds a call to `setDefaultHandler()` to the offline fallback example, since there needs to be a matching route for the offline fallback to take effect.

Fixes https://github.com/GoogleChrome/workbox/issues/2827, fixes https://github.com/GoogleChrome/workbox/issues/2824

**CC:** @petele @tropicadri @Snugug


src/content/en/tools/workbox/modules/workbox-recipes.md
```diff
@@ -22,23 +22,36 @@ The offline fallback recipe allows your service worker to serve a web page, imag
 
 This recipe, by default, assumes the fallback page is `offline.html` and that there isn't an image or font fallback. See the [offline fallback options](/web/tools/workbox/reference-docs/latest/module-workbox-recipes#~offlineFallback) for a list of all configuration options.
 
+The offline fallback will only be applied if there's a matching [route](/web/tools/workbox/modules/workbox-routing) for a given request. When using offline fallback alongside other recipes, like for the page and image caches, a route will automatically be set up for those requests. When using offline fallback on its own, you will need to explicitly create a route that matches page and image requests. The simplest approach is to use the [`setDefaultHandler()`](/web/tools/workbox/modules/workbox-routing#set_a_default_handler) method to create a route that applies the [`NetworkOnly`](/web/tools/workbox/modules/workbox-strategies#network_only) strategy to all requests, as shown below.
```
Snugug, line 25: Sorry, can't do inline comments for whatever reason here. Just some wordsmithing:

`Other recipes, like the [page cache](#page_cache), [static resource cache](#static_resources_cache), or [image cache](#image_cache), set up routes for their respective caches, allowing the offline fallback to work as expected. If you're using the offline fallback recipe on its own, you'll need to create routes yourself. The simplest way to do is to use…`

Snugug: Is this something we want to include as an option in the offline fallback recipe itself? Is there a way to test for this to see if routes have been created and do it automatically instead of through a config option, if we were to include it?

Snugug (commented):
A wordsmithing comment, take it or leave it, and a question about whether to include this functionality directly in Recipes

src/content/en/tools/workbox/modules/workbox-recipes.md
jeffposnick: Changed the wording.

I can imagine enabling a default `NetworkOnly` handler being a useful option in the offline fallback recipe, but I am not sure about enabling that by default—definitely not in v6 at least, as it would be a breaking change.

src/content/en/tools/workbox/modules/workbox-recipes.md
```diff
@@ -22,23 +22,36 @@ The offline fallback recipe allows your service worker to serve a web page, imag
 
 This recipe, by default, assumes the fallback page is `offline.html` and that there isn't an image or font fallback. See the [offline fallback options](/web/tools/workbox/reference-docs/latest/module-workbox-recipes#~offlineFallback) for a list of all configuration options.
 
+The offline fallback will only be applied if there's a matching [route](/web/tools/workbox/modules/workbox-routing) for a given request. Other recipes, like the [page cache](#page_cache), [static resource cache](#static_resources_cache), or [image cache](#image_cache), set up routes for their respective caches, allowing the offline fallback to work as expected. If you're using the offline fallback recipe on its own, you'll need to create routes yourself. The simplest way to do is to use the [`setDefaultHandler()`](/web/tools/workbox/modules/workbox-routing#set_a_default_handler) method to create a route that applies the [`NetworkOnly`](/web/tools/workbox/modules/workbox-strategies#network_only) strategy to all requests, as shown below.
```
tropicadri, line 25: I am confused by the sentence "allowing the offline fallback to work as expected", I think it means that the other recipes set up the required route that you are describing, but "work as expected" sounds broader. Maybe switching the order of the sentences first to If you are using the offline fallback on its own.... and then something like when offline fallback is used with another recipes like...those recipes do the set up the route required by offline fallback or something like that.

jeffposnick: Okay, I've swapped the order. Let me know if that reads better.

tropicadri: I'm not that invested in this, take it or leave it :), I don't like the "allowing the offline fallback to work as expected." what does work as expected means? if the offline fallback requires a route and there is no route and it doesn't work then it is working as expected because it requires a matching route. I think what the other recipes are doing is setting up that route for you, so you don't have to do that step on your own.

jeffposnick: Gotcha—let me know what you think about the new wording.

tropicadri (approved):

PR #9387, opened 2021-07-07
Add redirect for autoplay article.

jpmedley:


petele (changes requested):
@jpmedley - you also need to delete the source doc in `en` and any other languages.

jpmedley:
PTAL

petele (approved):

PR #9399, opened 2021-07-26
Fix MediaRecorder code sample

beaufortfrancois:
This CL fixes an issue with MediaRecorder code sample where `dataavailable` event was never fired because `stop()` was not called.


beaufortfrancois:
@petele Please review/merge

petele (approved):

PR #9463, opened 2021-11-16
Update the Working with the Fetch API article

umaar:
We'd like to update https://developers.google.com/web/ilt/pwa/working-with-the-fetch-api so it's more up-to-date! This is so we can link to it from some internal training. Have updated some hyperlinks and the code samples, so they use more modern syntax.

Have staged the site and manually verified that my content displays correctly.

**CC:** @petele @housseindjirdeh


housseindjirdeh (approved):
Thanks for spending the time to update this and fixing outdated links! This LGTM 👍

umaar:
Thank you!

PR #9470, opened 2021-11-30
Remove all references to Google Surveys

samtstern:
DevSite is trying to move away from Google Surveys (f.k.a Google Consumer Surveys a.k.a `gcs`) and instead move all surveys onto HaTS.

Right now we don't detect any active surveys on these IDs so this removal is safe. For information on using HaTS surveys see internal docs:
https://developers.devsite.corp.google.com/devsite/reference/metadata/project?hl=en#showing_surveys


samtstern:
@petele @chybie PTAL?

jeffposnick (approved):

PR #9475, opened 2021-12-07
Remove all references to prshim from l10n'd files.

agektmr:
Remove all references to prshim from l10n'd files.


agektmr:
Note these files are already set to be redirected so it's safe to delete.

chybie (approved):

PR #9528, opened 2022-03-14
Redirects batch 7

mattletford:
from: /web/updates/2011/08/Saving-generated-files-on-the-client-side
to: https://web.dev/browser-fs-access/

from: /web/updates/2018/08/chrome-ux-report-dashboard
to: https://web.dev/chrome-ux-report-data-studio-dashboard/

from: /web/fundamentals/primers/service-workers/high-performance-loading
to: https://developer.chrome.com/docs/workbox/caching-strategies-overview/

from: /web/fundamentals/primers/service-workers/lifecycle
to: https://developer.chrome.com/docs/workbox/service-worker-lifecycle/

@rachelandrew

PR #9536, opened 2022-03-15
fix: correct broken redirects

kaczor6418:
Revert "Redirects batch 7 (#9528)"

The original PR broke some redirects for old articles. I have created a bug for GoogleChrome/developer.chrome.com In which I have described what is broken now. Here you can find issue I have raised https://github.com/GoogleChrome/developer.chrome.com/issues/2291

This is why I think it should be reverted


rachelandrew:
as mentioned in my reply to the issue, please don't put back old content. Everything is being moved off this site.

PR #9593, opened 2022-05-23
Adds a caveat to the SW performance article

jeffposnick:
R: @philipwalton 
CC: @petele 

The conclusions in this article from 2016 don't necessarily apply today. I believe it's more useful to keep this alive and add a note to the top of it, but alternatively, we could find someplace else to redirect the URL to.


philipwalton (approved):

malchata (approved):
LGTM.

PR #9624, opened 2022-07-15
Redirect crux docs

rachelandrew:
This is, I believe, the final redirect from this site to the new Chrome User Experience docs.

I think those docs are in google3, rather than in GitHub so no deletions here.

**CC:** @petele

### DIFF
--- a/src/content/en/tools/_redirects.yaml
+++ b/src/content/en/tools/_redirects.yaml
@@ -466,12 +466,12 @@ redirects:
 
 - from: /web/tools/workbox/reference-docs/latest/workbox
   to: https://developer.chrome.com/docs/workbox/modules/workbox-sw/
-  
+
 # Generic redirect. If it wasn't specified above, and there is no matching
 # path within this folder, redirect to the URL specified.
 - from: /web/tools/workbox/reference-docs/...
   to: https://developer.chrome.com/docs/workbox/reference/
-  
+
 # Generic redirect. If it wasn't specified above, and there is no matching
 # path within this folder, redirect to the URL specified.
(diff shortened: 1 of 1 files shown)


src/content/en/tools/_redirects.yaml
```diff
@@ -467,11 +467,36 @@ redirects:
...
+  to: https://developer.chrome.com/docs/crux/release-notes/
+
+- from: /web/tools/chrome-user-experience-report/api
+  to: https://developer.chrome.com/docs/crux/api/
+
+- from: /web/tools/chrome-user-experience-report/api/...
+  to: https://developer.chrome.com/docs/crux/api/
+
+- from: /web/tools/
+  to: https://developer.chrome.com/docs/
```
petele, line 488: not necessary - this is handled by the redirect at the end of the file

src/content/en/tools/_redirects.yaml
```diff
@@ -467,11 +467,36 @@ redirects:
...
+  to: https://developer.chrome.com/docs/crux/
+
+- from: /web/tools/chrome-user-experience-report/bigquery/getting-started
+  to: https://developer.chrome.com/docs/crux/bigquery/
+
+- from: /web/tools/chrome-user-experience-report/bigquery/changelog
+  to: https://developer.chrome.com/docs/crux/release-notes/
+
+- from: /web/tools/chrome-user-experience-report/api
+  to: https://developer.chrome.com/docs/crux/api/
```
petele, line 482: this is covered by the one immediately below it `/api/...`

petele (commented):
I'll make the two quick changes then merge and push

